PR #44, opened 2026-07-05
fix(status): repaint pogo status --live in place to stop flicker

### LINKED ISSUE #43: pogo status --live flickers on every refresh
## Describe the Bug

`pogo status --live` flashes/blanks the whole screen on every refresh instead of updating smoothly in place.

## To Reproduce

1. Start pogod with at least one agent running.
2. Run `pogo status --live` (optionally `--interval 300ms` to make it obvious).
3. Watch the dashboard refresh — the screen goes fully blank, then repaints, on each tick.

## Expected Behavior

The dashboard updates in place like `top`/`watch` — changed values are overwritten without the screen ever going blank.

## Actual Behavior

Each refresh emits `ESC[2J` (erase entire screen) **before** the data fetch, then redraws via ~20 separate writes.

Root cause (`cmd/pogo/main.go`, live loop): the full-screen erase runs at the top of every tick, and the subsequent `renderStatus()` performs an `mg list` subprocess call plus two pogod HTTP calls (~300–500ms) before printing anything. So the terminal sits fully blank for the entire fetch latency on every tick, which reads as flicker. Emitting the frame as many small writes widens the gap.

Confirmed by capturing the emitted byte stream: `ESC[2J` appears once per tick, ahead of the content.

## Expected fix

Fetch + format the whole frame into a buffer first, then repaint in a single write using cursor-home (`ESC[H`) + per-line clear-to-EOL (`ESC[K`) + trailing erase-to-end-of-screen (`ESC[J`); reserve `ESC[2J` for a one-time initial clear. No control bytes should be emitted until the frame is ready.

## Environment

- OS: macOS (Darwin 25.5.0)
- Go version: go1.26.3
- Pogo version: commit 920e8c9 (main)
- Shell: zsh
- Editor/integration: terminal

## Additional Context

Same flicker-free repaint technique used by `top`/`watch`. Fix in progress.

bath-tub:
Fixes #43.

## Problem

`pogo status --live` blanks the whole screen on every refresh. The live loop emitted `ESC[2J` (erase entire screen) at the top of each tick, then `renderStatus()` ran an `mg list` subprocess plus two pogod HTTP calls (~300–500ms) before printing anything. The terminal sat fully blank for the whole fetch latency each tick, which reads as flicker. Emitting the frame as ~20 separate writes widened the gap.

## Fix

- `renderStatus()` now builds the entire frame into a `strings.Builder` and returns it, instead of printing incrementally.
- Live mode repaints in a single write: cursor-home (`ESC[H`) → each overwritten line cleared to EOL (`ESC[K`) → trailing erase-to-end-of-screen (`ESC[J`) to drop leftover lines from a taller previous frame.
- `ESC[2J` full clear is emitted only once at startup. No control bytes are written until the frame is ready — same technique as `top`/`watch`.
- Non-live text output and `--json` are unchanged.

## Verification

Captured the emitted byte stream against a live pogod (10 frames over 3s at `--interval 300ms`):

| | before | after |
|---|---|---|
| full-screen erase `ESC[2J` | every tick | 1 (startup only) |
| per-frame repaint | blank → fetch → draw | `ESC[H` + 17× `ESC[K` + `ESC[J`, single write |

`go build ./...`, `./build.sh` (13/13 tests), `go vet`, and `gofmt -l` all clean.

### DIFF
--- a/cmd/pogo/main.go
+++ b/cmd/pogo/main.go
@@ -255,7 +255,11 @@ Use --json for the raw structured response.`,
 	var statusInterval time.Duration
 	var statusTag string
 
-	renderStatus := func() {
+	// renderStatus fetches the current dashboard state and returns it as a
+	// fully-formatted text frame. In JSON mode it prints directly and returns
+	// "". The whole frame is built into a buffer before returning so live mode
+	// can write it to the terminal in a single flicker-free update.
+	renderStatus := func() string {
 		type statusReport struct {
 			Agents    []agent.AgentInfo       `json:"agents"`
 			WorkItems string                  `json:"work_items,omitempty"`
@@ -293,21 +297,25 @@ Use --json for the raw structured response.`,
 
 		if jsonOutput {
 			cli.PrintJSON(report)
-			return
+			return ""
 		}
 
 		// --- Text output ---
+		// Build the entire frame into a buffer so callers can emit it in one
+		// write. Never print incrementally here: in live mode a partially
+		// written frame is exactly what causes visible flicker.
+		var b strings.Builder
 
 		if statusLive {
-			fmt.Printf("pogo status --live  (every %s, Ctrl-C to quit)\n\n", statusInterval)
+			fmt.Fprintf(&b, "pogo status --live  (every %s, Ctrl-C to quit)\n\n", statusInterval)
 		}
 
 		// Agents section
-		fmt.Println("=== Agents ===")
+		fmt.Fprintln(&b, "=== Agents ===")
 		if agentErr != nil {
-			fmt.Printf("  (unavailable: %s)\n", agentErr)
+			fmt.Fprintf(&b, "  (unavailable: %s)\n", agentErr)
 		} else if len(agents) == 0 {
-			fmt.Println("  No agents running.")
+			fmt.Fprintln(&b, "  No agents running.")
 		} else {
 			crew := 0
 			polecats := 0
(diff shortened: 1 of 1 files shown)


drellem2:
## Review round 1: PASS
Reviewer: mg-261c · triage contract: mg-6405 · blocking: 0 · advisory: 2

Reviewed against the approved triage recommendation (buffered frame, single-write `ESC[H` + per-line `ESC[K` + trailing `ESC[J` repaint, `ESC[2J` startup-only, non-live/JSON untouched). The diff delivers exactly that scope — no creep, no omissions.

### Verification
- `gofmt -l`, `go vet ./...`, `go build ./...` clean on the PR head; CI 10/10 green. `go test ./...` passes except a pre-existing `internal/project` environment failure that reproduces identically on the merge base and is untouched by this diff.
- Byte capture under a pty against a live pogod (7 frames, `--interval 300ms`): exactly 1× `ESC[2J` at startup, then per-frame `ESC[H` + 26× `ESC[K` + `ESC[J`, one write per frame. No blank-screen window.
- Fetch-failure path (pogod port unreachable + `mg` absent from PATH): every section renders `(unavailable: …)` inside the buffered frame, repaint stays in-place, no partial writes, loop continues.
- Ctrl-C exit: trailing newline, cursor left after the frame; no cursor-hide or alternate-screen state to restore, so no stuck terminal on any exit path.
- Non-live text output byte-identical to the merge base (modulo uptime/age); `--json` document shape unchanged.
- Resize/line-wrap artifacts are the known `top`/`watch`-class limitation and are no worse here than expected.

### Advisory (non-blocking)
- `cmd/pogo/main.go:422` — `--live --json` behavior changed, for the better: the base emitted `ESC[2J`/`ESC[H` interleaved with the JSON stream each tick (corrupting piped output); this PR emits a clean stream of JSON documents with zero control bytes. Worth a line in the PR description, since it currently says `--json` is unchanged.
- `cmd/pogo/main.go:406` — `--live --interval 0` panics in `time.NewTicker` (non-positive interval). Pre-existing on the merge base, identical before/after this PR; noting for a possible follow-up, not for this change.

PR #70, opened 2026-07-10
FIX: pogod attach-socket lifecycle — dead listener while process lives (mg-d216)

drellem2:
Daniel could not `pogo agent attach mayor`: the dial failed with `connect: connection refused` against `mayor.sock`, a socket file that existed, while the mayor process was alive, healthy, and 40h into its loop.

## Root cause

On a unix socket, "file exists + connection refused" means exactly one thing: **nothing is accepting on it.**

`acceptLoop` returned on *any* `Accept` error:

```go
conn, err := l.Accept()
if err != nil {
    return // listener closed
}
```

That comment was the bug — not every `Accept` error is a closed listener. Under fd exhaustion `Accept` returns `EMFILE`/`ENFILE`, which are *transient*. The loop retired itself while the listener stayed bound. The socket file lingered, nothing accepted, and once the 128-slot listen backlog filled, every subsequent attach was refused. Nothing ever rebound it, so a long-running crew agent stayed unattachable for the rest of its life. (This machine had an fd-leak window — see mg-d205 — which is the most likely trigger.)

Reproduced standalone: bind a unix socket, never call `Accept`, dial it 129 times. The file is `srwxr-xr-x`, dial #129 returns `connect: connection refused` — Daniel's symptom exactly.

## Fix

Both halves of the ticket's fix direction:

1. **The accept loop no longer dies silently.** Transient `Accept` errors are retried with bounded backoff (5ms doubling to 1s). Only a deliberate close (`net.ErrClosed`) exits the loop.
2. **The listener is tied to process lifetime.** A supervisor rebinds the attach socket when it stops being served, when the socket file is unlinked underneath a live listener (macOS reaps stale entries under `$TMPDIR` — the likely explanation for the socket mtime landing well after process start), or when the path is taken over by a foreign bind (detected by inode identity). It starts even when the first bind fails, so an agent spawned during fd exhaustion regains attach instead of losing it permanently.

`Cleanup` stays the authority on teardown: it retires the supervisor under the same lock a rebind takes, so a retired agent can never have its socket resurrected — the one race worth getting right here, and it has a dedicated test.

Each repair emits an **`agent_attach_rebound`** event (`reason` ∈ `accept_loop_stopped`, `no_listener`, `socket_file_missing`, `socket_file_replaced`) so operators see the fault occurring instead of inferring it from a refused attach. Documented in `docs/event-log.md`; additive, no `schema_version` bump.

## Tests

`internal/agent/attach_listener_test.go` — real spawned processes, real unix sockets:

- rebinds after the accept loop dies with the socket file left behind (the reported fault, injected in its exact shape)
- rebinds after the socket file is removed
- reclaims the path after a foreign listener takes it over (asserted on inode, since a dial would succeed against the foreign backlog and prove nothing)
- never recreates the socket after `Cleanup`, incl. a direct test of the Cleanup-vs-rebind race
- `acceptLoop` retries a real `EMFILE` `*net.OpError` and terminates only on `ErrClosed`; backoff is bounded

I verified the three rebind tests **fail** without the supervisor, so they pin the regression rather than merely passing. `./build.sh` is green; `go test ./internal/agent/ -race` is clean.

## Scope notes

- This is the durable root-cause fix. The immediate mayor restart is a separate track (Daniel's call) and is unaffected by this PR — a restart would also have cleared the wedge, just not prevented the next one.
- The ticket carries no `gh:` issue ref and names no triage ticket (it originated from an Apple Reminder, via mayor's in-loop diagnosis), so there is no issue to close and no triage recommendation to diff against. The fix direction in the ticket body served as the spec.

Work item: mg-d216

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,18 @@ is the curated, human-readable summary kept in sync at each release cut.
 
 ## [Unreleased]
 
+### Fixed
+
+- **pogod: attach socket no longer dies while the agent lives.** `pogo agent attach <name>`
+  could fail with `connect: connection refused` against a socket file that existed,
+  on an agent whose process was alive and healthy. The accept loop returned on *any*
+  `Accept` error — including transient `EMFILE`/`ENFILE` under fd exhaustion — leaving
+  the socket bound with nothing accepting; once the listen backlog filled, every
+  subsequent attach was refused, permanently. The accept loop now retries transient
+  errors with bounded backoff, and a supervisor tied to process lifetime rebinds a
+  listener that stops serving or whose socket file is unlinked or replaced. Each
+  repair emits an `agent_attach_rebound` event. (mg-d216)
+
 ## [0.4.0] - 2026-07-09
 
 Minor release rolling up the 45 feature and fix commits merged since v0.3.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,8 +19,10 @@ is the curated, human-readable summary kept in sync at each release cut.
   the socket bound with nothing accepting; once the listen backlog filled, every
   subsequent attach was refused, permanently. The accept loop now retries transient
   errors with bounded backoff, and a supervisor tied to process lifetime rebinds a
-  listener that stops serving or whose socket file is unlinked or replaced. Each
-  repair emits an `agent_attach_rebound` event. (mg-d216)
+  listener that stops serving or whose socket file is unlinked or replaced. Both
+  the accept retry and the rebind repair are rate-limited, so a permanently broken
+  socket cannot spin the daemon or flood the event log. Each repair emits an
+  `agent_attach_rebound` event. (mg-d216)
 
 ## [0.4.0] - 2026-07-09
 
--- a/docs/event-log.md
+++ b/docs/event-log.md
@@ -116,6 +116,30 @@ A crew agent that crashed has been automatically restarted by pogod's supervisor
 {"schema_version":1,"timestamp":"2026-04-25T11:02:50.310000000Z","event_type":"agent_restarted","agent":"crew-arch","details":{"previous_pid":47011,"new_pid":47089,"restart_count":2}}
 ```
 
+#### `agent_attach_rebound`
+
+pogod repaired an agent's attach socket while the agent process kept running. The
+socket had stopped serving connections — see `reason` — so `pogo agent attach`
+would have failed against a live, healthy agent. Emitted once per repair; the
+agent is not restarted and loses no state. A steady trickle of these for one
+agent points at whatever keeps breaking the socket (fd exhaustion, a `$TMPDIR`
+reaper) rather than at the attach mechanism itself. Additive — no
+`schema_version` bump.
+
+- **Required envelope:** `schema_version`, `timestamp`, `event_type`, `agent`, `details`
+- **`details` fields:**
+  - `pid` (int, required): PID of the still-running agent process
+  - `socket` (string, required): path of the rebound unix socket
+  - `reason` (string, required): one of
+    - `accept_loop_stopped` — the accept loop exited under a live process. The socket file lingers with nothing accepting, so once the listen backlog fills, attach gets `connection refused`.
+    - `no_listener` — no listener was ever bound (e.g. the bind at spawn failed under fd exhaustion).
+    - `socket_file_missing` — the socket file was unlinked underneath a live listener.
+    - `socket_file_replaced` — a different socket now occupies the path.
+
+```json
+{"schema_version":1,"timestamp":"2026-07-10T09:12:03.410000000Z","event_type":"agent_attach_rebound","agent":"crew-mayor","details":{"pid":23884,"socket":"/var/folders/4n/T/pogo-agents/mayor.sock","reason":"accept_loop_stopped"}}
+```
+
 ### Polecat-specific lifecycle
 
 `agent_spawned` and `agent_stopped` already cover polecats. The two events below give polecat lifecycle a dedicated, work-item-scoped narrative for tools that want to filter by polecat events without inferring from `agent_type`.
--- a/docs/event-log.md
+++ b/docs/event-log.md
@@ -123,8 +123,13 @@ socket had stopped serving connections — see `reason` — so `pogo agent attac
 would have failed against a live, healthy agent. Emitted once per repair; the
 agent is not restarted and loses no state. A steady trickle of these for one
 agent points at whatever keeps breaking the socket (fd exhaustion, a `$TMPDIR`
-reaper) rather than at the attach mechanism itself. Additive — no
-`schema_version` bump.
+reaper) rather than at the attach mechanism itself.
+
+Repairs are rate-limited: a listener that fails again the instant it is rebound
+(a recurring permanent `accept(2)` error) backs off from 50ms to a ceiling of
+30s, so a persistently broken socket cannot flood this log. The backoff resets
+once a listener has stayed healthy for five minutes, so unrelated faults hours
+apart each get an immediate repair. Additive — no `schema_version` bump.
 
 - **Required envelope:** `schema_version`, `timestamp`, `event_type`, `agent`, `details`
 - **`details` fields:**
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -10,6 +10,7 @@ import (
 	"bufio"
 	"context"
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"io"
 	"log"
@@ -134,6 +135,28 @@ type Agent struct {
 	// socketPath is the unix domain socket for attach.
 	socketPath string
 	listener   net.Listener
+
+	// socketIno is the inode of the socket file the current listener is bound
+	// to. The supervisor compares it against whatever now sits at socketPath to
+	// notice a socket that was unlinked underneath us (macOS reaps stale entries
+	// under $TMPDIR) or replaced by a foreign bind. 0 means "unknown, skip it".
+	socketIno uint64
+
+	// listenerDead is closed by acceptLoop when it stops serving. The supervisor
+	// waits on it so a listener that dies while the process lives gets rebound
+	// rather than leaving a socket file nobody accepts on (mg-d216).
+	listenerDead chan struct{}
+
+	// attachStop is closed by Cleanup to retire the supervisor. attachStopped
+	// records the same fact under mu so a rebind that is already in flight
+	// aborts instead of recreating a socket for a torn-down agent.
+	attachStop    chan struct{}
+	attachStopped bool
+
+	// supervisorInterval is this agent's copy of attachSupervisorInterval,
+	// snapshotted on the spawning goroutine so a test that retunes the package
+	// default cannot race a running supervisor. Zero means use the default.
+	supervisorInterval time.Duration
 
 	// attachConns tracks active attach/terminal connections for output fanout.
 	// readOutput fans out PTY output to these in addition to the ring buffer.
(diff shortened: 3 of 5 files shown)


drellem2:
CI (`test`, linux) failed on the first push — `TestAttachRebindsAfterAcceptLoopDies`. Fixed in 153b41a. Worth writing down, because the test failure exposed a soft spot in the *implementation*, not just the assertion:

```
attach_listener_test.go:85: expected a freshly bound socket, inode before=8406852 after=8406852
```

The supervisor rebound correctly. But on Linux, closing the old listener freed its inode and the immediately following `net.Listen` **reused the same inode number**, so my "the inode must have changed" assertion was simply false. macOS happened not to reuse it, which is why this was green locally.

Two changes:

1. **Identify the bound socket with `os.FileInfo` + `os.SameFile` (device + inode)** instead of a raw inode number — same semantics, idiomatic, and it drops a hand-rolled `syscall.Stat_t` unwrap. I documented the reuse caveat on the field rather than papering over it: a *live* listener pins its inode, so a foreign bind cannot reuse it while we hold it — which is exactly the situation the `socket_file_replaced` check exists to catch. The `accept_loop_stopped` and `socket_file_missing` signals are exact and unaffected.

2. **Assert on listener identity, not inode churn.** A rebind installs a new listener and a new dead channel; that pair *is* what "rebound" means, and it is platform-independent. The tests additionally assert the agent owns the file at its own path afterwards.

I re-verified that all three rebind tests still **fail** when the supervisor is disabled, so they continue to pin the regression rather than passing for incidental reasons. `./build.sh` green, `go test ./internal/agent -race -count=3` clean.

drellem2:
## Review round 1: FAIL
Reviewer: mg-1346 · build ticket: mg-d216 · blocking: 1 · advisory: 3

Verified first: `./build.sh` and `./test.sh` are green, `go test ./internal/agent/ -race` is clean, `gofmt`/`go vet` clean. The three rebind tests do fail without the supervisor (I neutered `superviseListener` and confirmed all three fail; `TestAcceptLoopSurvivesTemporaryError` fails against the old return-on-any-error loop). They pin the regression as claimed.

I also reproduced the reported fault end-to-end against a live agent under a sandboxed pogod: a foreign non-accepting listener on the agent's socket path yields `srwxr-xr-x` + `ECONNREFUSED` — Daniel's exact symptom — and pogod reclaimed the path and restored attach within one supervisor tick, emitting `agent_attach_rebound{reason=socket_file_replaced}`. Removing the socket file recovered the same way. Healthy attach is unaffected. The fix direction in mg-d216 is delivered.

One defect blocks the merge.

### Blocking

- `internal/agent/agent.go:1460` — the supervisor's rebind path has no backoff, so a permanently-failing accept loop becomes an unbounded hot loop.

  `acceptLoop` (agent.go:1395) returns on any `Accept` error that is neither `net.ErrClosed` nor `Temporary()`. `superviseListener` observes the closed `dead` channel, rebinds, starts a fresh accept loop, and loops — with no sleep and no failure counter. When the error recurs on the new listener, the cycle repeats at syscall speed.

  The retry predicate at agent.go:1389 rests on `Temporary()`, and `syscall.Errno.Temporary()` is only `EINTR || EMFILE || ENFILE || Timeout()`. `accept(2)` can also return `ENOMEM` and `ENOBUFS` under memory pressure, and `EPERM` under a sandbox policy. All three are classified permanent here:

  ```
  EMFILE   Temporary()=true   -> retried with bounded backoff (safe)
  ENFILE   Temporary()=true   -> retried with bounded backoff (safe)
  ENOMEM   Temporary()=false  -> acceptLoop returns -> rebind, no backoff
  ENOBUFS  Temporary()=false  -> acceptLoop returns -> rebind, no backoff
  EPERM    Temporary()=false  -> acceptLoop returns -> rebind, no backoff
  ```

  Measured: with `acceptLoop` patched to return immediately (standing in for a recurring `ENOMEM`), a single agent drove **10,900 rebinds/sec**, each emitting an `agent_attach_rebound` event — **2.18 MB written to `events.log` in one second, ~7.8 GB/hour** — while pegging a core, for the life of the process.

  Repro: in `acceptLoop`, `return` before `l.Accept()`; spawn any agent; count `rebindListener` calls or `events.log` growth over one second.

  Why it blocks: this is a new unbounded work loop in a long-lived daemon, on the error path this PR introduces, reachable under exactly the resource-exhaustion conditions the PR exists to survive. Under memory pressure it writes multiple GB/hour to `events.log`, which feeds back into the pressure. It also contradicts `docs/event-log.md`, which documents `agent_attach_rebound` as "emitted once per repair" and reads a "steady trickle" as diagnostic signal.

  A pass needs the rebind path bounded — reuse `nextAcceptBackoff` in `superviseListener` (reset after a listener survives some interval), and/or treat `ENOMEM`/`ENOBUFS`/`ECONNABORTED` as retryable at agent.go:1389 rather than relying on `Temporary()` alone. The `no_listener` path is already safe: a failed bind `continue`s to the 30s ticker, which I confirmed against an unbindable path.

### Advisory (non-blocking)

- `internal/agent/attach_listener_test.go:31` — the new tests emit real `agent_attach_rebound` events into the developer's live `~/.pogo/events.log`, because `internal/agent` tests do not isolate the event-log path. One `./test.sh` run left 33 rebound events (`cat-rebind-dead-loop`, `cat-rebind-unlinked`, `cat-rebind-replaced`) in the real log. Test-emitted `agent_spawned` is pre-existing, but this event type is an operator alarm for a production fault, so a test run now manufactures false alarms in the signal this PR adds. Consider isolating `POGO_HOME` (or using `events.EmitTo`) for these tests.

- `cmd/pogod/main.go:625` — `socketDir` is `os.TempDir()/pogo-agents`, not derived from `PogoHome()`. `docs/operations.md:144` states distinct `POGO_HOME` roots "isolate distinct daemons completely", but two daemons with identically-named agents share one socket path. The supervisor turns the pre-existing collision from "the losing daemon silently loses attach" into a permanent 30s reclaim war, each side unlinking the other's live socket and emitting `socket_file_replaced` indefinitely. Pre-existing isolation gap, amplified rather than caused here; worth a follow-up ticket to derive `socketDir` from `PogoHome()`.

- `internal/agent/agent.go:1468` — a socket path that can never bind (e.g. `sun_path` over ~104 bytes) logs a rebind failure every 30s for the life of the agent, with no rate limit or give-up. Observed under a long sandbox path. Cosmetic; the backoff fix above likely covers it.

drellem2:
## Round 1 response — all findings addressed (7d536d6)

Thanks for the end-to-end reproduction and the measurement; the blocking finding is real and I have fixed it rather than argued with it.

### Blocking: unbounded rebind loop — **fixed**

You are right on both counts, including the subtle one. `Errno.Temporary()` is only `EINTR | EMFILE | ENFILE | Timeout()`, so `ENOMEM`/`ENOBUFS`/`EPERM` fell through `acceptLoop` to a rebind path that had no sleep and no failure counter. I reproduced it at **453 rebinds per 600ms** with `acceptLoop` returning immediately.

Both halves of your suggested remedy:

1. **`superviseListener` throttles repair** — 50ms doubling to a 30s ceiling, applied after a failed bind *and* after a rebind whose listener may die again immediately. The backoff **resets once a listener has stayed healthy for 5 minutes** (`attachRebindResetAfter`), so unrelated faults hours apart each still get an immediate repair rather than inheriting a stale 30s penalty. The sleep is interruptible on `stop`/`done`, so `Cleanup` never blocks behind a backoff.
2. **The retry predicate no longer leans on `Temporary()` alone** — new `isRetryableAcceptErr` names `ENOMEM`, `ENOBUFS`, `ECONNABORTED`, `ECONNRESET` explicitly, on top of whatever `Temporary()` reports. `EPERM`, `EBADF`, `EINVAL` stay permanent and fall to the now-bounded rebind path, which is the right place for them: retrying `accept(2)` forever on `EPERM` would be its own hot loop.

Regression tests, both of which fail against the unthrottled code:

- `TestSupervisorThrottlesRebindFlap` — kills each new accept loop the instant the supervisor installs it (your `ENOMEM` stand-in) and bounds the rebinds observed in a 600ms window. Unthrottled: 453. Throttled: ~4.
- `TestIsRetryableAcceptErr` — table over the eleven errnos above, so the `Temporary()` gap cannot silently reopen.
- `TestRebindBackoffIsBounded` — growth and saturation.

I also updated `docs/event-log.md`, which as you noted contradicted the new behavior: it now documents that repairs are rate-limited and that the backoff resets after five healthy minutes.

### Advisory 1: tests emit real `agent_attach_rebound` into `~/.pogo/events.log` — **fixed**

Agreed, and it is worse than cosmetic: a test run manufactures operator alarms for the exact production fault this event exists to announce. Root cause is that `events.Emit` resolves its path once, before any test can re-point `HOME`, so per-test isolation could not work. `TestMain` now points the event log at a throwaway file for the whole `internal/agent` package, which fixes the pre-existing `agent_spawned` leakage too. Verified: a full package run adds **zero** lines to the real log (I diffed before/after).

Note for whoever picks this up: the **33 false `agent_attach_rebound` records already in Daniel's live `~/.pogo/events.log`** (from my pre-fix runs and yours) are still there. I have deliberately not rewritten his event log — it is an append-only operator record and not mine to edit. Flagging so nobody reads them as a real fault.

### Advisory 2: `socketDir` from `os.TempDir()`, not `PogoHome()` — **not fixed, deliberately**

Concur with your analysis, including that the supervisor turns a silent collision into a 30s reclaim war between two daemons. It is a pre-existing isolation gap that this PR amplifies rather than causes, and fixing it means changing where every agent socket lives — out of scope for a socket-lifetime bug fix, and it deserves its own review. I have asked the coordinator to file a follow-up to derive `socketDir` from `PogoHome()`.

### Advisory 3: rebind-failure log every 30s on an unbindable path — **covered**

The rebind backoff now applies to failed binds as well, so the log line is bounded by the same 50ms→30s schedule. I did not add a give-up: a path can become bindable again (fd pressure clears, a `$TMPDIR` reaper stops), and permanently surrendering attach is the failure mode this PR exists to remove. The accept-retry log is separately rate-limited to one line per 30s per error streak.

`./build.sh` green, `go test ./internal/agent -race` clean, CI green on the previous push. Ready for re-review.

drellem2:
## Review round 2: PASS
Reviewer: mg-1346 · build ticket: mg-d216 · blocking: 0 · advisory: 3

Round-1's blocking finding is fixed, and I verified it independently rather than taking the tests' word for it. Re-running my own round-1 probe (`acceptLoop` patched to return immediately, standing in for a recurring `ENOMEM`) against `7d536d6`:

```
round 1 head (217ab56):  10,900 rebinds in 1s   — hot loop, ~2.18 MB/s into events.log
round 2 head (7d536d6):       5 rebinds in 1s, 6 in 3s   — bounded, backoff doubling
```

`isRetryableAcceptErr` now names `ENOMEM`/`ENOBUFS`/`ECONNABORTED`/`ECONNRESET` on top of `Temporary()`, and `EPERM`/`EBADF`/`EINVAL` correctly stay permanent and fall through to the now-bounded rebind path — retrying `accept(2)` forever on `EPERM` would just be a different hot loop, so that split is right.

Both new tests genuinely pin their regressions; I confirmed by neutering the fix, not by reading them:

- `TestSupervisorThrottlesRebindFlap` — fails against unthrottled code: *"supervisor rebound 452 times in 600ms — rebind path is not throttled"*.
- `TestIsRetryableAcceptErr` — fails against the `Temporary()`-only predicate, naming all four missed errnos exactly.

The throttle also reads correctly where it matters: repair stays immediate (the sleep is *after* `rebindListener`, so detection-to-repair latency is unchanged), the backoff resets only after a listener survives 5 minutes, and `sleepInterruptibly` selects on `stop`/`done` so `Cleanup` never blocks behind a backoff.

End-to-end against binaries built from this head, on a sandboxed pogod: healthy attach unaffected; a foreign non-accepting listener reproduces Daniel's exact symptom (`srwxr-xr-x` + `connection refused`); pogod reclaimed the path and restored attach in ~28s (one supervisor tick) and emitted **exactly one** `agent_attach_rebound` event, not a flood.

Advisory 1 (tests polluting the live event log) is fixed at the root, and better than I suggested — `events.Emit` resolves its path once, so per-test isolation could not have worked; `TestMain` re-points the whole package via the existing `events.SetLogPathForTesting` seam. Verified: a full `go test ./internal/agent/` run now writes **zero** lines to the real `~/.pogo/events.log` (144,014 before and after), which also closes the pre-existing `agent_spawned` leakage.

`./build.sh` and `./test.sh` green, `go test ./internal/agent/ -race` clean, `gofmt`/`go vet` clean. `docs/event-log.md` now documents the throttle and the 5-minute reset, so it no longer contradicts the code. Design-faithful to mg-d216: no scope creep, no silent omissions.

### Advisory (non-blocking)

- `cmd/pogod/main.go:625` — `socketDir` derives from `os.TempDir()`, not `PogoHome()`, so two daemons with distinct `POGO_HOME` and same-named agents share a socket path and the supervisor turns a silent collision into a 30s reclaim war. Deferred by agreement: pre-existing, amplified rather than caused here, and moving every agent socket deserves its own review. Follow-up ticket requested.

- `internal/agent/agent.go:1523` — on the *failed-bind* path the rebind backoff is inert: `bindListenerLocked` leaves `listenerDead` nil, so the next loop iteration parks on the 30s ticker regardless of the backoff just computed. Harmless (the ticker bounds it), but the sleep buys nothing there.

- `internal/project` — `TestDiscoverNewReposSkipsProtectedHomeDir` flaked once for me under full-suite load (`TempDir RemoveAll cleanup: directory not empty`, a cleanup race against the periodic indexer). Unrelated to this PR — the package is untouched, and it passes on both `main` and this head on re-run. Noted only so it isn't mistaken for a regression here.

Issue #64, opened 2026-07-08
Refinery count appears shared between host and containerized pogo instances


bath-tub:
## Summary

Two pogo instances run concurrently on this machine — one inside a Docker
container, one on the host. Both report the **same refinery count (18)**. The
identical value across two supposedly independent instances suggests they are
sharing refinery/work state, which would be a bug: separate instances should
have independent state unless a shared store is intentional.

## Environment

- Host pogo instance + containerized pogo instance running at the same time.
- Observed: refinery count = `18` reported by **both** instances, seen when
  invoking `pogostatus-live`.

## Expected

Each instance's refinery count reflects only that instance's own state. Two
independent orchestrators should not report an identical count by coincidence.

## Hypotheses to investigate

1. **Shared bind-mount (most likely, config not code):** the container mounts
   the host's `~/.macguffin` and/or `~/.pogo`, so both daemons read/write the
   same work-item + refinery state and the counts are identical by construction.
   If so, this is expected given the mount — but it also means the two instances
   are *not* isolated, which has knock-on effects (see related bridge issue).
2. **Globally-derived counter:** the refinery count is computed from a shared or
   absolute path rather than instance-scoped state.
3. **Cross-instance contention:** both daemons touch the same state/lockfiles
   under `~/.pogo`, causing one instance's refinery to act on the other's data.

## Repro / diagnosis steps

- Run host pogo and container pogo simultaneously.
- Compare the refinery count each reports.
- Check whether `~/.macguffin` and `~/.pogo` resolve to the same underlying
  volume/inode in both the host and the container (`mount`, `ls -di`, the
  container's volume config).

## Ask

Determine whether the shared count is (a) a bind-mount/config artifact (the two
instances share a store on purpose) or (b) a genuine cross-instance state leak
in host pogo. If (a), document the isolation expectation; if (b), scope the
refinery state to the instance.

## Related

Blocks / informs bath/pogo-slack-bridge#1 — the Slack bridge needs to know
whether the host and container share a maildir before it can be pointed at the
container specifically.

drellem2:
Expected — both instances share the same POGO_HOME, and all pogo state (including the refinery) roots there; POGO_HOME fully isolates an instance. If the container bind-mounts the host ~/.pogo or shares HOME, identical refinery state is by construction, not a cross-instance leak. To isolate: give the container its own POGO_HOME (this also answers pogo-slack-bridge#1 — a distinct POGO_HOME gives it its own maildir).

[...]

PR #71, opened 2026-07-10
FIX: derive agent socketDir from PogoHome() not os.TempDir() (mg-8532)

drellem2:
## Problem

`cmd/pogod/main.go` derived the agent attach-socket directory from `os.TempDir()/pogo-agents`. `$TMPDIR` is **per-user, not per-`POGO_HOME`**, so two daemons on distinct `POGO_HOME` roots with identically-named agents resolved to **one** socket file — while `docs/operations.md` promised distinct roots "isolate distinct daemons completely". The singleton lockfile bars two pogods on the *same* root; nothing barred two on *different* roots from colliding here.

The old symptom was quiet: whichever daemon bound last owned the path, the other silently lost `pogo agent attach`. The mg-d216 attach supervisor made it **loud** — each daemon observes the other's bind as its own socket being replaced, unlinks that live socket and rebinds, forever, on its 30s ticker.

## Fix

`config.AgentSocketDir()` derives the dir from `PogoHome()` → `$POGO_HOME/agents/sockets/`. Distinct roots get distinct paths, so the promised isolation is real.

**One wrinkle: `sun_path`.** A unix socket path cannot exceed 103 usable bytes (darwin; 107 on linux), and a deep enough `POGO_HOME` leaves no room for the `<agent>.sock` leaf — bind would fail with `EINVAL`, i.e. we'd trade a collision for a dead listener. Such a root falls back to `$TMPDIR/pogo-agents-<hash of the cleaned root>`. **The hash keeps distinct roots distinct, so the invariant this PR exists to establish holds in the fallback too.** pogod logs a line when it takes that path. For the default root the new path is *shorter* than the old one (34 bytes vs 60), so the common case gains headroom.

## Verification (before/after, real daemons)

Two sandboxed pogods on distinct `POGO_HOME` roots, each with an agent of the same name:

| binary | socket paths | rebounds in ~90s |
|---|---|---|
| pre-fix (control) | **identical** — `/tmp/pogo-agents/<name>.sock` for both | **3 per daemon**, ~one per 30s tick, all `socket_file_replaced` |
| this PR | distinct, one per root | **0**, stable inode, both sockets connectable throughout |

The control reproduces the permanent 30s reclaim war verbatim. I also drove the fallback branch end-to-end with a 76-byte root: pogod logs, relocates to `/tmp/pogo-agents-346a1be4`, and the agent's attach socket binds and accepts connections.

## Path migration & consumers of the old path

**Every agent socket path moves**, so I audited every consumer:

- **`pogod` is the only place that constructs the dir** (one call site). `internal/agent` receives it via `NewRegistry(socketDir)` and never derives it.
- **No client hardcodes the path.** `pogo agent attach` and `agent status/list` call `client.GetAgent()` and use the daemon-reported `info.SocketPath` (`internal/agent/api.go`). Since the client reaches the daemon on the port resolved from *its own* `POGO_HOME`, attach already routed per-root — only the socket *file* collided. So clients follow the daemon with no change, and a new `pogo` CLI against an old `pogod` still works.
- **No migration of existing socket files.** They are transient: bound by a live pogod, recreated on start, and unlinked on agent cleanup. Deliberately **not** deleting the old `$TMPDIR/pogo-agents/` dir — a still-running old-binary pogod on another root may own those live sockets, and unlinking them is precisely the bug being fixed. They are inert files that the OS tmp reaper collects.
- **Permissions:** `NewRegistry` does `MkdirAll(socketDir, 0700)`, and `MkdirAll` stamps that mode on every parent it creates — which would have made a fresh root's `agents/` dir `0700` instead of its historical `0755` (it is shared with the prompt files). pogod now creates that parent at `0755` first; `sockets/` still lands `0700`, which is what an attach socket wants. Verified on a fresh root.

## Tests

New `internal/config/agent_socket_test.go`. The load-bearing ones:

- **distinct roots → distinct socket dirs**, asserted in *both* the normal and the fallback branch, and never equal to the shared pre-mg-8532 path.
- **bindability**: actually `net.Listen`s a socket named for the longest agent we budget for, under both branches — this checks the budget constant against the kernel rather than against the 104/108 folklore.
- determinism, trailing-slash stability (guards the `filepath.Clean` in the hash), and that the `POGO_HOME=$HOME` legacy normalization reaches the socket dir.

I mutation-tested them: reverting `AgentSocketDir` to the old `os.TempDir()` implementation fails 3 of them, including `distinct POGO_HOME roots share socket dir "/var/folders/.../pogo-agents"`.

`./build.sh` and `./test.sh` both exit 0; gofmt and `go vet` clean.

## Sequencing

Per the ticket, this **must merge before the next controlled pogod rebuild activates the d216 attach supervisor** — otherwise the supervisor amplifies the collision into the permanent reclaim war demonstrated above.

## Notes

- Xref: mg-d216, mg-f227, drellem2/pogo#64 (cross-referenced, not resolved by this PR).
- Ticket carries no `gh:` issue ref and names no review ticket, so there is no `Resolves` line and the mayor is dispatching review.
- `./build.sh` runs `go install`, so this branch's `pogod` is now the installed binary on this host. That is the safe direction (the fix is present if pogod restarts), but worth knowing before the next rebuild.

Work item: mg-8532

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,8 +22,11 @@ RUN apk add --no-cache ca-certificates git tini
 
 RUN addgroup -S pogo && adduser -S pogo -G pogo
 
-RUN mkdir -p /workspace/repos /tmp/pogo-agents && \
-    chown -R pogo:pogo /workspace /tmp/pogo-agents
+# Agent attach sockets used to live in /tmp/pogo-agents, pre-created here. Since
+# mg-8532 they hang off PogoHome() (POGO_HOME, else $HOME/.pogo), which pogod
+# creates itself at startup — as it already did for the rest of the state tree.
+RUN mkdir -p /workspace/repos && \
+    chown -R pogo:pogo /workspace
 
 COPY --from=builder /pogod /usr/local/bin/pogod
 
--- a/cmd/pogod/main.go
+++ b/cmd/pogod/main.go
@@ -621,8 +621,26 @@ Flags:
 		}
 	}()
 
-	// Initialize agent registry
-	socketDir := filepath.Join(os.TempDir(), "pogo-agents")
+	// Initialize agent registry. The socket dir hangs off PogoHome, not
+	// os.TempDir(): $TMPDIR is per-user, so two daemons on distinct POGO_HOME
+	// roots with identically-named agents used to share one socket file and
+	// fight the mg-d216 supervisor over it forever (mg-8532).
+	socketDir := config.AgentSocketDir()
+	if pogoHome := filepath.Clean(config.PogoHome()); strings.HasPrefix(socketDir, pogoHome+string(filepath.Separator)) {
+		// NewRegistry creates socketDir with mode 0700, and MkdirAll stamps that
+		// mode on every parent it creates on the way down — including the agents/
+		// dir the sockets share with the prompt files, which has always been
+		// 0755. Create that parent first so a fresh POGO_HOME ends up with the
+		// same layout as an existing one. socketDir itself still lands at 0700,
+		// which is what an attach socket wants.
+		if err := os.MkdirAll(agent.PromptDir(), 0755); err != nil {
+			fmt.Printf("Cannot create agent state dir %s: %v\n", agent.PromptDir(), err)
+			os.Exit(1)
+		}
+	} else {
+		log.Printf("pogod: POGO_HOME %s is too deep to hold unix sockets (sun_path limit); "+
+			"agent attach sockets live in %s instead — still unique to this POGO_HOME", pogoHome, socketDir)
+	}
 	var initErr error
 	agentRegistry, initErr = agent.NewRegistry(socketDir)
 	if initErr != nil {
--- a/cmd/pogod/main.go
+++ b/cmd/pogod/main.go
@@ -625,8 +625,8 @@ Flags:
 	// os.TempDir(): $TMPDIR is per-user, so two daemons on distinct POGO_HOME
 	// roots with identically-named agents used to share one socket file and
 	// fight the mg-d216 supervisor over it forever (mg-8532).
-	socketDir := config.AgentSocketDir()
-	if pogoHome := filepath.Clean(config.PogoHome()); strings.HasPrefix(socketDir, pogoHome+string(filepath.Separator)) {
+	socketDir, insidePogoHome := config.AgentSocketDir()
+	if insidePogoHome {
 		// NewRegistry creates socketDir with mode 0700, and MkdirAll stamps that
 		// mode on every parent it creates on the way down — including the agents/
 		// dir the sockets share with the prompt files, which has always been
@@ -639,7 +639,8 @@ Flags:
 		}
 	} else {
 		log.Printf("pogod: POGO_HOME %s is too deep to hold unix sockets (sun_path limit); "+
-			"agent attach sockets live in %s instead — still unique to this POGO_HOME", pogoHome, socketDir)
+			"agent attach sockets live in %s instead — still unique to this POGO_HOME",
+			config.PogoHome(), socketDir)
 	}
 	var initErr error
 	agentRegistry, initErr = agent.NewRegistry(socketDir)
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -300,9 +300,10 @@ filesystem-coordination model.
 
 Every pogo state path derives from a single root: `$POGO_HOME`, or `~/.pogo`
 when the variable is unset (`PogoHome()` in `internal/config/config.go`). That
-one function seeds `refinery-state.json`, `schedules.json`, `agents/`,
-`polecats/`, `events.log`, `recovery/`, `projects.json`, and `plugin/` — so the
-root you pick determines where *all* daemon state lives.
+one function seeds `refinery-state.json`, `schedules.json`, `agents/` (including
+the `agents/sockets/` attach sockets), `polecats/`, `events.log`, `recovery/`,
+`projects.json`, and `plugin/` — so the root you pick determines where *all*
+daemon state lives.
 
 **Running N pogo instances requires a distinct `POGO_HOME` per instance.**
 Because every state path hangs off `PogoHome()`, overriding `POGO_HOME` (or
@@ -323,3 +324,13 @@ and pogo normalizes a `POGO_HOME` equal to the user's home directory to
 `$HOME/.pogo` (the documented default) rather than scattering state across the
 home root. See the `PogoHome()` doc comment for the full rationale, including why
 it never falls back to `os.TempDir()`.
+
+One caveat on the attach sockets: a unix domain socket path cannot exceed
+`sun_path` (103 usable bytes on darwin, 107 on linux), and a deep enough
+`POGO_HOME` leaves no room for `agents/sockets/<agent>.sock`. Such a root — a
+scratch dir under `/var/folders` on darwin, say — puts the sockets in
+`$TMPDIR/pogo-agents-<hash of the root>` instead. The hash keeps distinct roots
+distinct, so the isolation guarantee holds either way; pogod logs a line at
+startup when it takes this path. Everything else still lives under the root. If
+you want your sockets under `POGO_HOME` (nicer to inspect and clean up), pick a
+shallow root: `~/.pogo-sandbox` fits comfortably, a 90-byte path does not.
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -334,3 +334,13 @@ distinct, so the isolation guarantee holds either way; pogod logs a line at
 startup when it takes this path. Everything else still lives under the root. If
 you want your sockets under `POGO_HOME` (nicer to inspect and clean up), pick a
 shallow root: `~/.pogo-sandbox` fits comfortably, a 90-byte path does not.
+
+The same limit implies a ceiling on **agent names**: pogo reserves 24 bytes for
+`<agent>.sock` when choosing the socket directory (`MaxAgentNameLen`). Real names
+are far shorter — `pm-dealdesk` is 11, a polecat is named for its work item. But
+a name longer than 24 bytes, under a root deep enough to have consumed the
+headroom (roughly 53+ bytes), produces a socket path past `sun_path`: the agent
+spawns and runs, `pogo agent attach` does not work against it, and pogod logs
+`attach listener failed`. Nothing rejects such a name at spawn today. Under the
+default `~/.pogo` root there is room for a 64-byte name, so this is only
+reachable with both a deep root and a long name.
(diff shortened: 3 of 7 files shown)


drellem2:
## Review round 1: PASS
Reviewer: mg-473a · build ticket: mg-8532 · blocking: 0 · advisory: 5

Verified against the mg-8532 recommendation (derive `socketDir` from `PogoHome()`), not the PR description.

### Verification performed
- `./test.sh` exit 0; `go test -race ./internal/config/ ./internal/agent/ ./cmd/pogod/` clean.
- **Fix, two real daemons on distinct short roots, same agent name `twin`:** distinct socket paths, stable inodes and both sockets connectable across 75s (2.5 supervisor ticks), zero `agent_attach_rebound` events.
- **Control, same setup on `main`:** both daemons resolve to one `$TMPDIR/pogo-agents/twin.sock`, inode churns, 2 `socket_file_replaced` rebinds per daemon per 75s — the reclaim war reproduces exactly as described, so the zero above is meaningful.
- **Fallback branch, two 76-byte roots:** both relocate to distinct `$TMPDIR/pogo-agents-<hash>` dirs, log the relocation, and both sockets accept connections. Per-root distinctness survives the fallback.
- **`maxUnixSocketPathLen = 103` probed directly against this kernel:** 103 binds, 104 returns `EINVAL`. Constant is correct.
- **Consumers:** `cmd/pogod/main.go:628` is the only place the dir is constructed; `internal/agent/agent.go:727` (spawn), `:1035` (respawn) and `:1391` (mg-d216 rebind) all derive from `r.socketDir`; `cmd/pogo/main.go:804,865` use the daemon-reported `info.SocketPath`. A `pogo` CLI built from `main` against the new `pogod` reports and connects to the new path — no client migration needed.
- **Path migration:** no concern. Attach listeners are owned by the pogod process, and agents are its PTY children, so no live socket survives at the old path once the new binary runs. Leftover files under `$TMPDIR/pogo-agents/` are inert; deliberately not unlinking them is the right call.
- **Fresh root:** `agents/` lands `0755`, `agents/sockets/` lands `0700`, as claimed.

### Advisory (non-blocking)
- `Dockerfile:25` — still `mkdir -p /tmp/pogo-agents` and chowns it. That path is dead after this change. Harmless (`cmd/pogod/main.go:585` already required a writable `PogoHome` before this PR, so the container is unaffected), but it is a stale reference to the path this PR removes.
- `internal/config/config.go:635` — `agentSocketLeafBudget` reserves 24 bytes of agent name, but nothing enforces that ceiling: `internal/agent/api.go:809` only rejects an empty name. Repro: a 58-byte `POGO_HOME` (derived dir exactly 73, the budget maximum), a 24-char agent binds; a 30-char agent fails `bind: invalid argument`, `POST /agents/start` still returns 201, and the agent runs with attach silently unavailable. That name bound fine pre-PR under the shorter `$TMPDIR` dir. Narrow (needs a >52-byte root *and* a >24-char name; the default root tolerates 64) and non-fatal, hence advisory — but the budget is currently an unenforced assumption. Consider validating name length at spawn, or documenting the ceiling.
- `internal/config/agent_socket_test.go:93` — `TestAgentSocketDirBindable` says it checks the budget "against the kernel rather than against the 104/108 folklore", but both branches yield short dirs (`shortHome` → `/tmp/phXXXXXXXX`; `deepHome` → the short fallback), so no bind ever approaches `sun_path`. The test would still pass with a too-large `maxUnixSocketPathLen`. The constant is right — I probed it — but the test does not guard it. A case whose derived dir is exactly `maxUnixSocketPathLen - agentSocketLeafBudget` would.
- `cmd/pogod/main.go:629` — the `strings.HasPrefix(socketDir, pogoHome+sep)` test re-derives, by string inspection, the branch `AgentSocketDir` already took. Returning that decision (or exporting the predicate) keeps the two from drifting. Edge case: `POGO_HOME=/` yields the prefix `//` and mis-reports the fallback.
- `docs/operations.md:148` — states sockets live in `agents/sockets/` and that distinct roots isolate "completely", with no pointer to the `sun_path` fallback documented at `docs/CONFIGURATION.md:328`. Isolation does hold in the fallback, so this is not wrong, only incomplete.

drellem2:
## Review round 1 — response (commit `20fc12e`)

Thanks — the independent replication of the reclaim war and the kernel probe of `maxUnixSocketPathLen` were both genuinely useful. Folded in 4 of 5 advisories; advisory 2 is filed as a follow-up with reasoning below.

**Advisory 4 — fallback detected by string inspection.** Fixed. `AgentSocketDir()` now returns `(dir string, insidePogoHome bool)`, so pogod uses the branch decision the function already made instead of re-deriving it. You're right about the `POGO_HOME=/` edge — `strings.HasPrefix(dir, "/"+"/")` mis-reported a fallback for a root that actually fits. `TestAgentSocketDirRootPogoHome` covers it.

**Advisory 3 — the bindability test didn't guard the constant.** Correct, and a fair hit: my comment overclaimed. Neither `shortHome` nor `deepHome` produced a path near `sun_path`. Two tests now pin it:
- `TestMaxUnixSocketPathLenBinds` — binds a path of exactly `maxUnixSocketPathLen`.
- `TestAgentSocketDirBindableAtBudgetLimit` — sizes `POGO_HOME` so the derived dir lands exactly on the leaf budget, binds a `MaxAgentNameLen` socket (a 103-byte path), and asserts that one byte deeper tips into the *fallback* rather than a failed bind.

I mutation-tested both: raising the constant to 104 fails them with `bind: invalid argument`. Neither asserts that 104 *fails*, since 103 is the conservative darwin figure applied on every platform and linux tolerates 107 — over-large is the bug worth catching, under-large is the intent.

**Advisory 1 — stale `/tmp/pogo-agents` in the Dockerfile.** Fixed. A genuine miss: my consumer audit grepped `*.go`/`*.md`/`*.sh` and never looked at the Dockerfile. Dropped the `mkdir`/`chown`; pogod creates the dir under `PogoHome()` itself, as it already did for the rest of the state tree. Docker isn't available on this host, so I made the minimal removal rather than adding a `mkdir -p /home/pogo/.pogo` that would bake in an unverified assumption about `$HOME` in the container — your note that `main.go:585` already required a writable `PogoHome` pre-PR is what makes the removal safe.

**Advisory 5 — `operations.md` incomplete.** Fixed; it now points at `CONFIGURATION.md` for the `sun_path` fallback and the agent-name ceiling.

**Advisory 2 — unenforced `MaxAgentNameLen`.** Confirmed real, and confirmed a regression in the narrow case: the old 60-byte `$TMPDIR` dir fit a 30-char name where the new 73-byte dir does not. **Filed as mg-ef80 rather than fixed here**, because every clean fix is API-visible — rejecting a long name at spawn starts 400ing names that work fine under a shallow root, and making the bind failure fatal changes what `POST /agents/start` means. Neither belongs in a PR being reviewed for path migration. What I did do:
- renamed the constant to exported `MaxAgentNameLen` and documented it as *a ceiling, not a promise*, naming the failure mode explicitly;
- documented the ceiling and its reachability (needs a ~53+ byte root **and** a >24-char name; the default `~/.pogo` tolerates 64) in `docs/CONFIGURATION.md`;
- mg-ef80 records your repro and notes that option (a) alone is insufficient — validating the name still lets the budget math drift — so (a)+(b) is the honest fix.

`./build.sh` and `./test.sh` exit 0; gofmt and vet clean. Re-verified end-to-end after the signature change: two daemons on distinct roots still get distinct sockets, fresh-root modes are still 0755/0700, shallow roots log no spurious fallback, and the deep-root fallback still logs, relocates, and binds a connectable socket.

drellem2:
## Re-confirm at 20fc12e: PASS holds
Reviewer: mg-473a · build ticket: mg-8532 · blocking: 0 · advisory: 0 open (1 deferred to mg-ef80)

The branch moved after the round-1 verdict, so the new head was re-reviewed rather than assumed. The round-1 PASS stands and no new findings were introduced.

### Delta re-verified
- `./test.sh` exit 0; `go test -race ./internal/config/ ./internal/agent/ ./cmd/pogod/` clean.
- **Advisory 3 fix, mutation-tested rather than taken on trust:** with `maxUnixSocketPathLen` bumped 103 → 104, both `TestMaxUnixSocketPathLenBinds` and `TestAgentSocketDirBindableAtBudgetLimit` fail. The constant is now genuinely guarded by the kernel. Declining to assert that 104 fails is right — 103 is the conservative darwin figure and linux tolerates 107, so over-large is the only bug worth catching here.
- **Advisory 4 fix:** `AgentSocketDir()` returning `(dir, insidePogoHome)` removes the string-prefix re-derivation, so the branch decision cannot drift from the branch taken. Re-ran the daemon end to end after the signature change: two shallow roots still produce distinct socket paths, stable inodes and connectable sockets across 75s with zero rebinds; a 76-byte root still logs, relocates to its hashed fallback dir and binds a connectable socket; shallow roots log no spurious fallback; fresh-root modes remain `0755`/`0700`.
- **Advisory 1 fix:** the `Dockerfile` no longer pre-creates `/tmp/pogo-agents`. The minimal removal is the right call — `cmd/pogod/main.go:585` already required a writable `PogoHome` before this PR, so nothing needs to be added in its place.
- **Advisory 5 fix:** `docs/operations.md` now points at the fallback caveat.

### Advisory 2 — deferral accepted
Holding the `MaxAgentNameLen` enforcement out of this PR is correct. Every clean fix is API-visible (rejecting a long name at spawn starts `400`ing names that work fine under a shallow root; making bind failure fatal changes what `POST /agents/start` means), and neither belongs in a change reviewed for path migration. Exporting `MaxAgentNameLen` and documenting it as a ceiling nothing enforces — naming the failure mode at `internal/config/config.go:630` — is the honest interim state. Tracked as mg-ef80.

No blocking findings. Ready to merge.

drellem2:
## Post-merge re-confirm: CONFIRM (clean)
Reviewer: mg-473a · build ticket: mg-8532 · merged as 20fc12e · blocking: 0 · open advisory: 0 (1 deferred to mg-ef80)

Re-verified against the merged `main` checkout rather than the PR branch. `main` is a fast-forward to `20fc12e`, identical to the head re-confirmed above.

- `AgentSocketDir() -> (dir, insidePogoHome)`: correct on both edges. `POGO_HOME=/` yields `/agents/sockets`, which fits, so `inside=true` is the right answer — that is precisely why the old prefix test was wrong. A real pogod on `POGO_HOME=/` dies on the lockfile before reaching the socket code, as it did pre-PR.
- Boundary tests mutation-tested, not read: `maxUnixSocketPathLen` 103 → 104 fails both `TestMaxUnixSocketPathLenBinds` and `TestAgentSocketDirBindableAtBudgetLimit`.
- `Dockerfile` carries no reference to the retired `/tmp/pogo-agents`.
- `docs/operations.md:162` cross-refs the `sun_path` caveat; the anchor resolves to `docs/CONFIGURATION.md:299`.
- Real daemons on merged `main`: two shallow roots give distinct socket paths, stable inodes and connectable sockets with zero rebinds across 75s; a 76-byte root logs, relocates and binds a connectable socket.
- `./test.sh` exit 0; `go test -race ./internal/config/` clean.

`main` is safe for the mg-d216 supervisor deploy. Remaining work is mg-ef80 only.

PR #72, opened 2026-07-10
build.sh: build into ./bin instead of go install into GOBIN (mg-b630)

drellem2:
## Problem

`build.sh` ended in `go install ./cmd/...`, which writes to GOBIN (`~/go/bin`). But `build.sh` is not only a developer command — it also runs unattended in polecat worktrees and as the refinery's default quality gate. So **every** build that touched `cmd/pogod` silently overwrote the machine's installed `pogod` with an unreviewed branch build, and a later `pogod` restart could launch whatever branch happened to compile last.

## Change

- `build.sh` now runs `go build -o ./bin/ ./cmd/...`. `./bin` is already gitignored (`/bin`); `POGO_BUILD_DIR` overrides it.
- Installing into GOBIN is opt-in: `./build.sh --install`.
- Unknown flags are now rejected (exit 2) instead of being silently ignored, so a typo'd `--install` can't quietly do the wrong thing.
- `build_test.sh` (new, wired into `test.sh`) exercises the real `build.sh` against a synthetic Go module with `GOBIN` redirected to a temp dir.
- Docs updated: `CLAUDE.md`, `CONTRIBUTING.md`, `docs/development.md`, `CHANGELOG.md`.

## The gate still gates

The property the refinery depends on is "a compile error fails the script," and that is covered directly: test 5 plants a syntax error in the fixture module and asserts `build.sh` exits non-zero **and** leaves GOBIN empty. Verified on this branch: a full `./build.sh` exits 0, and `shasum ~/go/bin/pogod` is byte-identical before and after.

```
Test 2: Default build writes ./bin and leaves GOBIN untouched
Test 3: --skip-tests skips the test step
Test 4: POGO_BUILD_DIR overrides the build directory
Test 5: A compile error fails the build   <- and leaves GOBIN empty
Test 6: --install populates GOBIN
Test 7: Unknown flags are rejected
=== Results: 10 passed, 0 failed ===
```

## No consumer depended on the `go install`

Checked before removing it: CI (`.github/workflows/ci.yml`) never invokes `build.sh`; releases are cut by goreleaser; `install.sh` downloads release archives rather than building; `Dockerfile` and `run.sh` don't rely on GOBIN. The only references to `build.sh` are as a refinery quality gate, which cares only about the exit status.

## Notes

- This ticket is internal (`mg-b630`) and has no originating GitHub issue, so there is no `Resolves #n` line. The recommendation is inline in the ticket body: *"polecat build.sh should 'go build -o <scratch path>' (or a build dir) rather than 'go install' to GOBIN … Verify no consumer depends on go-install populating GOBIN during a build."*
- The branch is `polecat-b630`, following the repo's existing `polecat-<id>` convention.
- Out of scope, noted not fixed: `run.sh` still sets `GOPATH` and builds a `bin/plugin/pogo-plugin-search` that no longer exists in the tree — it looks like pre-modules archeology. Xref: mg-8532, mg-d216.

Work item: mg-b630

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,15 @@ is the curated, human-readable summary kept in sync at each release cut.
 
 ### Fixed
 
+- **`build.sh` no longer overwrites the installed binaries.** The build step ran
+  `go install ./cmd/...`, writing to GOBIN (`~/go/bin`). Because `build.sh` also runs
+  unattended in agent worktrees and as the refinery's quality gate, every build that
+  touched `cmd/pogod` silently replaced the machine's installed `pogod` with an
+  unreviewed branch build — so a daemon restart could launch whatever branch compiled
+  last. `build.sh` now compiles into `./bin` (gitignored, overridable with
+  `POGO_BUILD_DIR`) and still fails on a compile error. Installing into GOBIN is now
+  opt-in via `./build.sh --install`. (mg-b630)
+
 - **pogod: attach socket no longer dies while the agent lives.** `pogo agent attach <name>`
   could fail with `connect: connection refused` against a socket file that existed,
   on an agent whose process was alive and healthy. The accept loop returned on *any*
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -66,10 +66,14 @@ If you add or advance an integration, update both `README.md` and this table.
 Go project. Build and test:
 
 ```bash
-./build.sh       # Build all binaries (runs fmt + test + install)
+./build.sh       # Build all binaries into ./bin (runs fmt + test + build)
 ./test.sh        # Run tests
 ./fmt.sh         # Format code (go fmt)
 ```
+
+`./build.sh` compiles into `./bin` and does **not** `go install` into GOBIN —
+that would overwrite the machine's live `pogod` with a branch build. Pass
+`--install` if you want the binaries on `PATH`.
 
 Binaries are in `cmd/`: `cmd/pogo`, `cmd/lsp`, `cmd/pose`, `cmd/pogod` (daemon).
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,19 +6,26 @@ Thanks for your interest in contributing to pogo! This guide covers the basics o
 
 1. Fork and clone the repository
 2. Install Go (1.24+)
-3. Run `./build.sh` to build, test, and install
+3. Run `./build.sh` to build and test
 
 ## Development Workflow
 
 ### Building
 
 ```bash
-./build.sh    # Format, test, and install all binaries
-./test.sh     # Run tests only
-./fmt.sh      # Format code only
+./build.sh              # Format, test, and build all binaries into ./bin
+./build.sh --install    # ...and also `go install` them into GOBIN
+./test.sh               # Run tests only
+./fmt.sh                # Format code only
 ```
 
-`./build.sh` runs all three steps (format, test, install) and is the recommended way to verify your changes before committing.
+`./build.sh` runs all three steps (format, test, build) and is the recommended way to verify your changes before committing.
+
+Binaries land in `./bin` (gitignored). `./build.sh` deliberately does *not* write
+to GOBIN: it runs unattended in agent worktrees and as the refinery's quality
+gate, where a `go install` would overwrite the machine's installed `pogod` with
+an unreviewed branch build. Pass `--install` when you actually want the binaries
+on your `PATH`, or set `POGO_BUILD_DIR` to redirect the output directory.
 
 ### Code Style
 
--- a/build.sh
+++ b/build.sh
@@ -1,14 +1,57 @@
 #!/bin/bash
+# build.sh — format, test, and compile every binary under cmd/.
+#
+# Binaries land in ./bin (gitignored); build.sh does NOT `go install` into
+# GOBIN. build.sh is executed in polecat worktrees and as the refinery's
+# quality gate, so a `go install` here silently overwrites the host's live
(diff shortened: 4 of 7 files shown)


drellem2:
## Review round 1: PASS
Reviewer: mg-ffd0 · build ticket: mg-b630 · blocking: 0 · advisory: 5

Verified on `d024bb7`. `./build.sh` exits 0 and `./test.sh` is green (`10 passed, 0 failed` from the new `build_test.sh`).

### Verification

- **The gate still gates.** Beyond the fixture coverage in `build_test.sh:115`, I planted a real compile error in `cmd/pogod/` (gofmt-clean, undefined symbol, so `fmt.sh` could not be what failed) and ran `sh -c './build.sh --skip-tests'` the way the refinery invokes it: exit 1, with the `go build` diagnostic on stdout. Not a no-op.
- **GOBIN untouched.** `shasum` of `~/go/bin/{pogo,pogod,lsp,pose}` is byte-identical before and after a full `./build.sh`. `--install` with `GOBIN` redirected to a temp dir populates all four binaries, so the opt-in escape hatch works.
- **No scratch leak.** `./bin` is ignored (`.gitignore:1`) and `git status` is clean after a build. Worktree teardown uses `git worktree remove --force` (`internal/agent/api.go:1048`, `internal/gitgc/gitgc.go:218`), so an ignored `bin/` cannot block removal. `build_test.sh:34` traps `EXIT` and removes its fixture; temp-dir count is unchanged across a run.
- **`test.sh` unaffected.** The only change is `bash build_test.sh` appended at `test.sh:23`; `set -e` at `test.sh:6` means a `build_test.sh` failure fails the suite.
- **Refinery invocation.** `runGate` (`internal/refinery/merge.go:615`) runs `sh -c "./build.sh"` in the clone and keys purely on exit status. Confirmed exit 0 on success, 1 on compile error, 2 on an unknown flag, 0 on `--help`. Gate semantics preserved.
- **No consumer depended on the `go install`.** Independently confirmed: `.github/workflows/ci.yml` never invokes `build.sh`, `Dockerfile` runs its own `go build`, `install.sh` fetches release archives, `run.sh` does not read GOBIN. The only `build.sh` callers are the refinery gate (`merge.go:464`) and the contributor docs.

Architecturally the change moves toward `docs/design/sandbox-design.md:90`, which wants gate execution confined to the worktree with no parent-tree writes — a `go install` into `~/go/bin` was exactly such a write. `build_test.sh` follows the existing shell-test convention (`shell/bashrc_test.sh`): `pass`/`fail` counters, `mktemp` fixture, `trap ... EXIT`, wired into `test.sh`.

### Advisory (non-blocking)

- `build_test.sh:45` — the fixture pins `go 1.25.0` in its generated `go.mod`, duplicating the repo's directive. It will drift silently the next time `go.mod` is bumped. A lower floor, or deriving the line from the repo's `go.mod`, would be sturdier.
- `build.sh:46` — rejecting unknown flags with exit 2 is scope the recommendation did not ask for. It is harmless (no caller passes flags) and the PR description declares it, so it is not a silent change — noting it for the record rather than objecting.
- Each agent worktree now retains ~115 MB in `./bin` until `gitgc` sweeps it, where previously every build shared one `~/go/bin`. The recommendation explicitly sanctioned "a build dir", so this is expected, not a regression.
- `internal/refinery/merge.go:464` runs both `./build.sh` and `./test.sh`, and `build.sh` itself runs `test.sh` — so `build_test.sh`'s fixture builds now execute twice per gate. Pre-existing double-run, marginally more expensive.
- `docs/development.md:12` says Go 1.21+, `CONTRIBUTING.md:8` says 1.24+, `go.mod` says 1.25.0. Pre-existing inconsistency in two files this PR touches; out of scope here.

### Faithfulness

The diff delivers both halves of the mg-b630 recommendation — `go build` into a build dir instead of `go install` to GOBIN, and a verified survey that no consumer relied on the install. No silent omissions.

drellem2:
## Round 1 acknowledgement (builder) — no new commits

Thanks for the independent verification, particularly planting a real compile error in `cmd/pogod/` with a gofmt-clean undefined symbol. That rules out `fmt.sh` as the thing that failed, which my fixture-based test 5 alone could not have proven.

**Deliberately not pushing a round-2 commit.** `d024bb7` is the commit you verified; none of the five advisories are blocking, and amending the branch after a pass would mean the refinery merges code that differs from what was reviewed. Disposition of each:

1. **Fixture pins `go 1.25.0` (`build_test.sh:45`)** — agreed it can drift, but it fails safe: a newer toolchain happily builds a module declaring an older directive, so a `go.mod` bump leaves the fixture building rather than breaking. Worth lowering the floor in a follow-up; not worth re-review churn now.
2. **Unknown-flag rejection (`build.sh:46`)** — fair call that it exceeds the literal recommendation. The reasoning: the old `for/case` loop silently ignored anything it did not recognize, so a typo'd `--instal` would have quietly produced a *non*-installing build while the developer believed otherwise. Since this PR makes `--install` the only path to GOBIN, silent flag-dropping became a footgun of the same family as the one being fixed. Declared in the PR body; happy to drop it if the mayor prefers strict scope.
3. **~115 MB per worktree in `./bin`** — expected; the recommendation sanctioned "a build dir". `gitgc` sweeps it, and `git worktree remove --force` is unaffected by an ignored directory, as you confirmed.
4. **`build_test.sh` fixture builds run twice per gate** — pre-existing double-run of `test.sh` via `merge.go:464`. The fixture is a single trivial `main` package, so the second pass is cache-warm and cheap.
5. **Go version skew (1.21+ / 1.24+ / 1.25.0)** — pre-existing and genuinely out of scope; I have flagged it to the mayor as a separate item rather than widening this diff.

No changes to the tree. `d024bb7` stands as the reviewed and submittable commit.

PR #73, opened 2026-07-10
FOLLOW-UP: enforce MaxAgentNameLen at agent spawn (mg-ef80)

drellem2:
## What

`internal/config.MaxAgentNameLen` (24) is the byte budget `AgentSocketDir()` reserves for the `"/<name>.sock"` leaf when choosing the attach socket directory — but nothing enforced it. Under a `POGO_HOME` deep enough to consume the remaining headroom, a longer agent name pushed the socket path past `sun_path`: the bind failed, pogod logged `attach listener failed`, and `POST /agents/start` **still returned 201** for an agent that could never be attached to.

Implements the ticket's recommended fix, **option (c) — both halves**:

**(a) Reject the name at spawn.** `ValidateAgentName` refuses an empty or over-long name, wired into `Registry.Spawn` (the choke point all three spawn paths funnel through) and into each handler, so a rejected polecat name never gets as far as creating a worktree, agent dir, or expanded prompt. The check is **unconditional**, not conditional on the root's depth: a name that works on a shallow root and silently loses attach on a deep one is worse than one refused everywhere.

**(b) Make an unbindable socket fatal to the spawn.** As the ticket notes, (a) alone still lets a deep root + long name fail if the budget arithmetic ever drifts. So a bind failure that no retry can clear now fails the spawn instead of logging. The listener binds *before* the PTY plumbing goroutines start and before the agent enters the registry, so the teardown is just a kill.

The fatal set is deliberately narrow — `EINVAL`/`ENAMETOOLONG`, the sun_path-overflow class. A **transient** failure (`EMFILE` under fd exhaustion) still keeps the agent alive and lets the mg-d216 supervisor rebind once fds free up; a false positive here would cost an agent that would otherwise have recovered. `TestSpawnSurvivesTransientBindFailure` pins that contract.

Spawn errors now map to **400** (bad name) and **500** (unbindable socket); the legacy catch-all **409** stays for genuine registry conflicts.

## Verification

Driven end-to-end against a sandboxed pogod on the reviewer's exact geometry (58-byte `POGO_HOME` → 73-byte socket dir, so a 24-char name lands on exactly 103 bytes and a 30-char name overruns at 109):

- 30-char name → `400`, message names the limit; no agent dir, no socket, no registry entry. Same on `POST /agents`, `/agents/start`, and `/agents/spawn-polecat`.
- 24-char name at the budget limit → `201`, and its 103-byte socket path is bound and **accepts connections**; no `attach listener failed` in the log.

Each new test was confirmed to **fail without the fix** (checked by neutering the length check and the errno classifier). `./build.sh` green; `go test -race ./internal/agent/` clean.

## Note on the test-file churn

Making an unbindable socket fatal surfaced that **90 of this package's 311 tests had never bound an attach socket.** They root their `Registry` at `t.TempDir()`, which on darwin returns a `/var/folders/...` path already longer than `sun_path` — so the bind failed, the failure was only a log line, and the tests passed while silently never exercising attach. (A handful of tests that *did* care already worked around this with a short `/tmp` dir, with a comment saying why.)

So the second commit promotes that workaround to the package default: `shortSocketDir` moves to `testsupport_test.go` and every test registry roots there. Those 90 tests now genuinely exercise attach binding. **Production was never affected** — pogod takes its socket dir from `config.AgentSocketDir`, which falls back to a short hashed path when `POGO_HOME` is too deep.

## Behavior change (API-visible, as the ticket anticipated)

A crew agent whose **configured** name exceeds 24 bytes will now fail to start, auto-start, or wake, with an explicit error — where it previously started without a working `pogo agent attach`. Recorded in `CHANGELOG.md` under an operator note. The default `~/.pogo` root has room for a 64-byte name, so no default deployment could reach the old failure.

## Out of scope (noted, not fixed)

- `ValidateAgentName` does not reject path separators, so a name like `../x` still escapes `filepath.Join(socketDir, name+".sock")` and `filepath.Join(PromptDir(), name)`. Pre-existing, unrelated to this ticket's recommendation, and worth its own review.
- `Respawn` still logs (rather than fails) on a fatal bind error. Unreachable in practice: the name and socket dir are fixed at registry creation, so a respawn cannot newly overflow a path that bound at spawn.

## Provenance

There is **no originating GitHub issue** for this work item — it came from advisory 2 (non-blocking) of the mg-8532 review by reviewer 473a on #71, which is merged. Nothing to `Resolves`.

Xref: #71, mg-8532, mg-d216.
Work item: mg-ef80

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,34 @@ is the curated, human-readable summary kept in sync at each release cut.
 
 ## [Unreleased]
 
+### Changed
+
+- **pogod: agent names longer than 24 bytes are refused at spawn.** `MaxAgentNameLen`
+  was the byte budget pogod reserves for `<agent>.sock` when choosing the attach
+  socket directory, but nothing enforced it. Under a `POGO_HOME` deep enough to
+  consume the remaining headroom, a longer name produced a socket path past the
+  `sun_path` limit: the bind failed, pogod logged `attach listener failed`, and
+  `POST /agents/start` still returned 201 for an agent that could never be
+  attached to. Such a name now gets a 400 from every spawn endpoint, under every
+  root — a name's fate must not depend on how deep the operator's root happens to
+  be. Real names are far shorter (`pm-dealdesk` is 11; a polecat is named for its
+  work item), and the default `~/.pogo` root always had room for a 64-byte name,
+  so no default deployment could reach the old failure. **Operator note:** a crew
+  agent whose configured name exceeds 24 bytes will now fail to start (and fail to
+  auto-start or wake) with an explicit error, where it previously started without
+  a working `pogo agent attach`. Rename it to 24 bytes or fewer. (mg-ef80)
+
 ### Fixed
+
+- **pogod: a spawn whose attach socket cannot bind now fails instead of lying.**
+  Backstop for the name check above: if binding the attach socket fails for a reason
+  no retry can clear — the path overruns `sun_path` — the spawn is torn down and
+  reports an error rather than returning a live agent nobody can reach. Transient
+  bind failures (fd exhaustion) still keep the agent, and the mg-d216 supervisor
+  rebinds once they clear. Relatedly, `AgentSocketDir`'s fallback root is now
+  guaranteed to leave room for the reserved name budget: it previously derived from
+  `os.TempDir()` unchecked, so a `TMPDIR` over ~52 bytes produced a socket directory
+  in which no legal agent name could bind. (mg-ef80)
 
 - **`build.sh` no longer overwrites the installed binaries.** The build step ran
   `go install ./cmd/...`, writing to GOBIN (`~/go/bin`). Because `build.sh` also runs
--- a/cmd/pogod/main_stallnudger_test.go
+++ b/cmd/pogod/main_stallnudger_test.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"os"
 	"path/filepath"
 	"strings"
 	"testing"
@@ -8,6 +9,22 @@ import (
 
 	"github.com/drellem2/pogo/internal/agent"
 )
+
+// shortSocketDir returns a socket dir short enough that "<dir>/<name>.sock"
+// fits AF_UNIX's sun_path limit. t.TempDir() on darwin lives under
+// /var/folders/... and already exceeds it on its own, so a registry rooted
+// there cannot bind an attach socket for any agent — which fails the spawn
+// outright since mg-ef80. Production takes this dir from config.AgentSocketDir,
+// which guarantees the fit.
+func shortSocketDir(t *testing.T) string {
+	t.Helper()
+	dir, err := os.MkdirTemp("/tmp", "pogo-pogod-sock-")
+	if err != nil {
+		t.Fatalf("MkdirTemp: %v", err)
+	}
+	t.Cleanup(func() { os.RemoveAll(dir) })
+	return filepath.Join(dir, "s")
+}
 
 // TestStallNudgerNeverInterruptsBusyAgent is the end-to-end proof of gh
 // drellem2/pogo #61 review point (a): the priority wake reuses newStallNudger,
@@ -17,7 +34,7 @@ import (
 // proves the wait-idle primitive) by exercising the exact function the stall
 // watcher and priority wake call.
 func TestStallNudgerNeverInterruptsBusyAgent(t *testing.T) {
-	reg, err := agent.NewRegistry(filepath.Join(t.TempDir(), "sockets"))
+	reg, err := agent.NewRegistry(shortSocketDir(t))
 	if err != nil {
 		t.Fatalf("NewRegistry: %v", err)
 	}
@@ -66,7 +83,7 @@ func TestStallNudgerNeverInterruptsBusyAgent(t *testing.T) {
 // registered/running, the nudger delivers via macguffin mail so the signal is
 // durable rather than dropped.
 func TestStallNudgerFallsBackToMailWhenOffline(t *testing.T) {
-	reg, err := agent.NewRegistry(filepath.Join(t.TempDir(), "sockets"))
+	reg, err := agent.NewRegistry(shortSocketDir(t))
 	if err != nil {
 		t.Fatalf("NewRegistry: %v", err)
 	}
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -335,12 +335,18 @@ startup when it takes this path. Everything else still lives under the root. If
 you want your sockets under `POGO_HOME` (nicer to inspect and clean up), pick a
 shallow root: `~/.pogo-sandbox` fits comfortably, a 90-byte path does not.
 
-The same limit implies a ceiling on **agent names**: pogo reserves 24 bytes for
-`<agent>.sock` when choosing the socket directory (`MaxAgentNameLen`). Real names
-are far shorter — `pm-dealdesk` is 11, a polecat is named for its work item. But
-a name longer than 24 bytes, under a root deep enough to have consumed the
-headroom (roughly 53+ bytes), produces a socket path past `sun_path`: the agent
-spawns and runs, `pogo agent attach` does not work against it, and pogod logs
-`attach listener failed`. Nothing rejects such a name at spawn today. Under the
-default `~/.pogo` root there is room for a 64-byte name, so this is only
-reachable with both a deep root and a long name.
+The same limit implies a hard ceiling on **agent names**: pogo reserves 24 bytes
+for `<agent>.sock` when choosing the socket directory (`MaxAgentNameLen`). Real
+names are far shorter — `pm-dealdesk` is 11, a polecat is named for its work
+item — so you are unlikely to meet this limit. A name longer than 24 bytes is
+rejected at spawn with HTTP 400 (`pogo agent start` and `pogo agent spawn-polecat`
+print the error and exit non-zero).
+
+The rejection is unconditional, not conditional on your root's depth. Only a root
+deep enough to have consumed the socket directory's headroom (roughly 53+ bytes)
+would actually push such a name's socket path past `sun_path` — the default
+`~/.pogo` root has room for a 64-byte name — but a name that works on one machine
+and silently loses attach on another is worse than a name that is refused
+everywhere. If pogod cannot bind an agent's attach socket at all, the spawn now
+fails outright rather than returning a running agent that `pogo agent attach`
+cannot reach.
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -335,6 +335,14 @@ startup when it takes this path. Everything else still lives under the root. If
 you want your sockets under `POGO_HOME` (nicer to inspect and clean up), pick a
 shallow root: `~/.pogo-sandbox` fits comfortably, a 90-byte path does not.
 
+`$TMPDIR` is itself unbounded, so if it is long enough to squeeze out the
+reserved name budget (roughly 52+ bytes), the sockets degrade one step further to
+`/tmp/pogo-agents-<hash of the root>`, which fits under any root. The hash — and
+with it the per-root isolation — is unchanged. This only matters if you run pogod
+with an unusually deep `TMPDIR`; the guarantee it protects is that **the 24-byte
+agent-name budget below holds under every root and every `TMPDIR`**, so a legal
+name never fails to bind.
+
 The same limit implies a hard ceiling on **agent names**: pogo reserves 24 bytes
 for `<agent>.sock` when choosing the socket directory (`MaxAgentNameLen`). Real
 names are far shorter — `pm-dealdesk` is 11, a polecat is named for its work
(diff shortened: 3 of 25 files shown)


drellem2:
## Review round 1: FAIL
Reviewer: mg-14de · build ticket: mg-ef80 · blocking: 2 · advisory: 5

Verified against the ticket's recommendation (option (c): reject at spawn **and** make an unbindable socket fatal). Both halves are implemented, tested, and behave as described. `./build.sh` is green, `go test -race ./internal/agent/` is clean, and each new test was confirmed to fail with the length check neutered.

Driven end-to-end against a sandboxed pogod on the reviewer's geometry (58-byte `POGO_HOME` → 73-byte socket dir): a 30-byte name is refused with 400 on `POST /agents`, `/agents/start`, and `/agents/spawn-polecat`, leaving no agent dir or socket behind; `pogo agent start` and `pogo agent spawn-polecat` print the error and exit 1. A 24-byte name at the limit spawns, binds its 103-byte socket, and accepts connections, with no `attach listener failed` in the log. Existing crew names (longest is `pm-onethird`, 11 bytes) are unaffected.

Two things block the merge.

### Blocking

- `internal/config/config.go:680` — **`AgentSocketDir`'s fallback branch is never checked against `agentSocketDirFits`, so `MaxAgentNameLen` is not the promise this PR relies on.** The primary branch returns `PogoHome()/agents/sockets` only when it fits. The fallback returns `filepath.Join(os.TempDir(), "pogo-agents-"+hash)` unconditionally. When `os.TempDir()` is longer than ~52 bytes the fallback dir itself has no room for the reserved leaf, and a name well inside the 24-byte ceiling fails to bind — which `Spawn` now treats as fatal.

  Reproduced on this branch, with a 60-byte `TMPDIR` and a 73-byte `POGO_HOME` (deep enough to force the fallback):

  ```
  POST /agents {"name":"ef80"}                    -> 201
  POST /agents {"name":"cccccccccccccccccccc"}    -> 500
    attach socket cannot be bound: agent cccccccccccccccccccc at
    /tmp/ttt…ttt/pogo-agents-09cc822b/cccccccccccccccccccc.sock:
    listen: bind: invalid argument
  ```

  That name is 20 bytes — legal per `ValidateAgentName`. On `main` it spawns (with attach unavailable); on this branch it cannot spawn at all. So the failure mode this PR set out to remove — a name's fate depending on how deep the operator's paths happen to be — still exists, relocated from silent attach loss to a hard 500.

  This is why it blocks rather than being a nit: the PR chooses a name-only check over the path-aware check the ticket asked for, and `internal/agent/name.go:22-24` justifies that choice by asserting "a name either works under every `POGO_HOME` or fails under all of them." The assertion is false as written, and `internal/config/config.go:639-641` ("it is a promise and not merely a ceiling") repeats it. The reasoning is sound; the invariant it rests on is not yet established.

  Cheapest fix that makes the invariant true: have the fallback return a root that always fits — e.g. `/tmp/pogo-agents-<hash>` (25 bytes) rather than `os.TempDir()`-relative — or assert `agentSocketDirFits` on the fallback and fail pogod startup loudly instead of 500ing every spawn. Either way the doc comments then describe the code.

- `CHANGELOG.md` — **the branch conflicts with `main` and will bounce the refinery.** `main` moved to `d024bb7`, which added a `### Fixed` entry at the top of `[Unreleased]`; this branch inserts `### Changed` plus its own `### Fixed` entry at the same anchor. `git merge-tree --write-tree main HEAD` reports `CONFLICT (content): Merge conflict in CHANGELOG.md`, and `internal/refinery/merge.go:155` aborts the merge request on a failed `rebase origin/main`. Mechanical, but it has to be resolved before the mayor can submit. Rebasing also picks up the `build.sh` fix, so a branch build stops overwriting the host's installed `pogod`.

### Advisory (non-blocking)

- `internal/agent/name.go:32` — `ValidateAgentName` accepts path separators, so `../x` still escapes both `filepath.Join(socketDir, name+".sock")` and `filepath.Join(PromptDir(), name)`. Pre-existing and correctly called out as out of scope in the PR description, but this function is now the obvious home for the check. Worth its own ticket rather than this round.
- `internal/agent/agent.go:1066` — `Respawn` still logs on a fatal bind error. Unreachable today for the reason the PR gives; it will stop being unreachable if the socket dir ever becomes per-agent.
- `internal/agent/agent.go:754` — the fatal teardown runs `cmd.Wait()` while holding `r.mu` (taken at `agent.go:620` with a deferred unlock). `SIGKILL` makes it prompt, so this is not a hang risk, but it serializes the whole registry behind a process reap.
- `internal/agent/agent.go:694` — `writeContextFilePrompt` may have written `AGENTS.override.md` into the agent's working directory by the time the fatal-bind path returns; the teardown does not remove it. The same is true of the existing `pty.StartWithSize` error path, so this is a pre-existing gap the new early-return widens slightly.
- `internal/agent/testsupport_test.go:24` — `shortSocketDir` hardcodes `/tmp` and ignores `TMPDIR`. Correct on darwin and linux, and it inherits the convention from the helper it replaces, but a sandboxed test runner without a writable `/tmp` would now fail the whole package rather than a single test.

The test-file churn is not scope creep: making an unbindable socket fatal is what exposes that 90 of these tests never bound a socket, and rooting them at a bindable dir is the minimum required to keep them honest. Good change.

drellem2:
## Round 1 response — both blockers fixed

Rebased onto `d024bb7`; branch head is now `f9f4e47`. `./build.sh` green, `go test -race ./internal/config/ ./internal/agent/` clean.

### BLOCKING 1 — `AgentSocketDir` fallback not checked against `agentSocketDirFits` → **fixed** (`f9f4e47`)

You're right, and the framing was the useful part: the deviation from VERIFY item 2 (name-only check instead of path-aware) is only defensible if `AgentSocketDir` actually reserves 24 bytes under every root. It didn't, so `name.go:22-24` and `config.go:639-641` were false as written, and the exact failure class I set out to remove survived — relocated from silent attach loss to a hard 500. I reproduced it before touching anything:

```
TMPDIR=/tmp/<55 t's> (60 bytes), POGO_HOME deep
AgentSocketDir() = "/tmp/ttt…ttt/pogo-agents-09cc822b" (81 bytes)
  -> 22 bytes of headroom for a 30-byte leaf
```

I took your first route — make the invariant true — rather than weakening the comments. The fallback still **prefers `os.TempDir()`** (per-user on darwin, and where these sockets already live) but only when it fits; otherwise it degrades to `/tmp`, which at 4 bytes leaves room under any budget these constants could plausibly grow to. Per-root distinctness (the hash) survives the degradation. If `/tmp` isn't writable, `NewRegistry`'s `MkdirAll` fails and pogod exits loudly at startup — which beats 500ing every spawn.

I preferred this over the "assert and fail startup" alternative because it keeps every existing deployment's socket dir exactly where it is (no migration), and reserves the loud-failure path for a genuinely unusable `/tmp`.

`TestAgentSocketDirAlwaysFits` pins it across a matrix of roots (shallow / deep / `/`) × TMPDIRs (short / at the budget / **one byte past it** / pathological), asserting both that the dir fits *and* that a `MaxAgentNameLen` name actually binds under it. Confirmed to fail on the old unchecked fallback, precisely at the one-byte boundary. Two companion tests pin that `os.TempDir()` is still preferred when it fits, and that distinct roots stay distinct after degrading.

Re-ran your end-to-end on the same geometry (TMPDIR 60 bytes, POGO_HOME 73 bytes — same `09cc822b` hash you saw):

| request | before | after |
|---|---|---|
| `{"name":"cccccccccccccccccccc"}` (20 bytes, legal) | `500` bind: invalid argument | **`201`**, socket bound at `/tmp/pogo-agents-09cc822b/…` and accepts connections |
| `{"name":"bbbb…"}` (30 bytes) | — | **`400`**, unchanged |

No `attach listener failed` in the log.

The comments now cite the guarantee instead of asserting it, and `MaxAgentNameLen`'s doc names the test that keeps it honest.

### BLOCKING 2 — CHANGELOG conflict → **fixed** (rebase)

Rebased onto `origin/main` (`d024bb7`) and re-resolved `CHANGELOG.md`, keeping both `### Fixed` entries. `git merge-tree --write-tree origin/main HEAD` is now clean. Confirmed the rebase picks up the `build.sh` fix — my `./bin` is populated and GOBIN is no longer written.

I extended my own `### Fixed` entry to cover the `AgentSocketDir` fallback repair, since it's user-visible for anyone running a long `TMPDIR`.

### Advisories

Per your note these don't need fixing this round, and I've left 1–5 as-is. Positions for the record:

1. **Path separators in `ValidateAgentName`** — agreed it's the obvious future home. Not folding it in: it's a real escape (`../x` → `filepath.Join(socketDir, name+".sock")` and `filepath.Join(PromptDir(), name)`), pre-existing, and deserves its own ticket and review rather than riding this one. Happy to file it.
2. **`Respawn` logs on fatal bind** — agreed, unreachable today; the name and socket dir are both fixed at registry creation.
3. **`cmd.Wait()` under `r.mu`** — accurate. SIGKILL keeps it prompt so it's not a hang, but it does serialize the registry behind a reap. Narrowing that lock scope is a bigger change than this PR should carry; noting it rather than doing it badly under time pressure.
4. **`AGENTS.override.md` not removed on the fatal-bind teardown** — correct, and as you say the pre-existing `pty.StartWithSize` error path has the same hole. Worth fixing in one place, not two.
5. **`shortSocketDir` hardcodes `/tmp`** — fair. It inherits the convention from the helper it replaces, and production now has the same `/tmp` last resort, so the two are at least consistent. A runner without a writable `/tmp` would fail the package; I'd rather fix that for helper and fallback together than half of it here.

Ready for round 2.

drellem2:
## Review round 2: PASS
Reviewer: mg-14de · build ticket: mg-ef80 · blocking: 0 · advisory: 2 (new) + 5 (carried, acknowledged)

Both round-1 blockers are fixed at `f9f4e47`. No new blocking findings. Merging is fine.

### Blocking findings from round 1 — both resolved

- `internal/config/config.go:700` — **resolved.** The fallback now prefers `os.TempDir()` only when `agentSocketDirFits` accepts it and otherwise degrades to `/tmp/pogo-agents-<hash>` (25 bytes, leaving 78 for the 30-byte leaf). Per-root distinctness survives the degradation. Re-ran the round-1 repro on the same geometry (60-byte `TMPDIR`, 73-byte `POGO_HOME`, same `09cc822b` hash): the 20-byte name that returned 500 now returns 201 with a connectable socket, the 24-byte name at the ceiling returns 201, and a 30-byte name still returns 400 — with no `attach listener failed` in the log. `TestAgentSocketDirAlwaysFits` fails against the old unchecked fallback at exactly the one-byte boundary and at the pathological row, so it pins the invariant rather than restating it. The doc comments now cite the guarantee and name the test that keeps it honest, instead of asserting it.

  The builder took the first of the two routes offered and picked the better one: keeping every existing deployment's socket dir where it is, with no migration, and reserving loud startup failure for a genuinely unusable `/tmp`.

- `CHANGELOG.md` — **resolved** by rebase onto `origin/main` (`d024bb7`), both `### Fixed` entries preserved. `git merge-tree --write-tree origin/main HEAD` is clean.

### Verification

`./build.sh` green (and, post-rebase, it builds into `./bin` and no longer overwrites the host's installed `pogod`). `go test ./...` green; `go test -race ./internal/config/ ./internal/agent/` clean. Re-checked the round-1 geometry (58-byte `POGO_HOME`) end-to-end: sockets stay under `POGO_HOME`, 24-byte name 201, 30-byte name 400, zero bind failures.

### Advisory (non-blocking, new this round)

- `internal/config/agent_socket_test.go:325` — `TestAgentSocketDirAlwaysFits` leaves an empty `/tmp/pogo-agents-<hash>` behind for each subtest that reaches the degraded branch (16 dirs in `/tmp` after one `go test ./internal/config/`), because `bindOK` at `:65` creates the directory and only the socket file is unlinked on `Close`. `TestAgentSocketDirBindable` already shows the fix: `t.Cleanup(func() { os.RemoveAll(dir) })`.
- `internal/config/config.go:700` — the `/tmp` last resort puts attach sockets under a world-writable parent. `NewRegistry`'s `MkdirAll(dir, 0700)` does not correct the mode of a directory that already exists, and `net.Listen` creates the socket file at `0755`, so a local user who pre-creates `/tmp/pogo-agents-<hash>` (the hash is derived from a guessable `POGO_HOME`) at `0777` could reach or replace an agent's attach socket, which brokers a PTY. Not a regression and not in this ticket's scope — on linux `os.TempDir()` is already `/tmp`, so the pre-`mg-ef80` fallback had the same exposure for a deep root — but this change makes the path reachable on darwin too. Worth its own ticket alongside the `ValidateAgentName` path-separator advisory below.

### Advisory carried from round 1 (acknowledged by the builder, correctly deferred)

- `internal/agent/name.go:35` — `ValidateAgentName` accepts path separators; `../x` still escapes `filepath.Join(socketDir, name+".sock")` and `filepath.Join(PromptDir(), name)`. Pre-existing; own ticket.
- `internal/agent/agent.go:1066` — `Respawn` still logs on a fatal bind error. Unreachable today.
- `internal/agent/agent.go:754` — the fatal teardown runs `cmd.Wait()` while holding `r.mu`. Not a hang risk under `SIGKILL`.
- `internal/agent/agent.go:694` — `AGENTS.override.md` is not removed on the fatal-bind teardown; the pre-existing `pty.StartWithSize` error path has the same hole.
- `internal/agent/testsupport_test.go:24` — `shortSocketDir` hardcodes `/tmp`, ignoring `TMPDIR`. Now consistent with production's own `/tmp` last resort.

The change delivers the ticket's option (c) in full, with the ceiling now backed by a real guarantee rather than a comment.

PR #74, opened 2026-07-10
Add Cursor as a pogo provider (mg-c146)

drellem2:
Adds `internal/cursor/provider.go` as pogo's fourth harness provider, built to the `internal/pi` template (descriptor + calibration doc + e2e). Selection is via the existing `[agents] provider` config — no config-surface change.

Every value in the descriptor was measured against a live, authenticated **Cursor CLI 2026.07.09-a3815c0** in a 200×50 PTY. Full evidence: `docs/investigations/cursor-nudge-calibration.md`.

## The ticket's open question is resolved — no architect consult needed

**Wrinkle 1** asked whether `.cursor/rules`-in-worktree is a viable escape hatch for the `AGENTS.md` collision. **It is**, with one non-obvious constraint.

Cursor has no `--append-system-prompt` (the only prompt-ish flag on the CLI is `--no-prompt`) and no `AGENTS.override.md`-style precedence file, so a separate rules namespace is the *only* additive injection point. The persona goes to `.cursor/rules/pogo-persona.mdc`.

The constraint: **frontmatter is mandatory, and its absence fails silently.** A behavioural probe (repo `AGENTS.md` and the pogo rule each instructing a distinct token; the model asked only to "greet me", never to read files), 3 runs per variant:

| `.mdc` frontmatter | Persona injected |
|---|---|
| *(none)* | **0/3** — silently ignored |
| `alwaysApply: false` | 3/3, but attachment is the *model's* decision, not a contract |
| `alwaysApply: true` | **3/3** — Cursor's documented "Always" rule type |

A plain copy — what `writeContextFilePrompt` did before this PR — would have delivered **nothing**, quietly. Collision verified absent: with both files present a single reply carried both tokens, and `AGENTS.md` was byte-identical afterwards.

> Methodological note recorded in the doc: an earlier probe asked the model to "list every marker in your rules/context" and **all three variants passed** — Cursor has file-read tools and simply grepped the worktree. A context-file injection test that lets the agent read the file under test proves nothing.

## Other measured findings that shaped the descriptor

- **Workspace trust needs a `PostSpawnHook`.** `--force` governs command approval, not trust; `--trust` is *rejected* outside `--print` (`Error: --trust can only be used with --print/headless mode`) and the process exits before the TUI renders. So the template is `agent --force` and `TrustDialogHook` dismisses the dialog. It presses **`a`**, not Enter — Enter selects the *highlighted* row, and a future menu reorder would select `[q] Quit` and silently kill every polecat.
- **`SubmitDelay` is load-bearing here, unlike pi's.** Cursor's composer has paste-burst detection: a combined `body+"\r"` write does **not** submit (probe timed out at 95s); a 50ms split write does.
- **Initial task via argv.** `agent [prompt...]` survives the trust dialog (3/3). A typed nudge would race a modal that is dead silent once drawn — the same trap that made Codex's nudge type its task into the trust dialog.

## Scope deviation, flagged

The ticket asked for an *offline* e2e per pi's pattern. **That is not possible for Cursor** and the PR does not pretend otherwise. pi's rig works because pi supports custom OpenAI-compatible providers; Cursor is closed-source SaaS speaking proprietary connectrpc/protobuf to `api2.cursor.sh`, and the one base-URL override in the bundle (`CURSOR_API_BASE_URL`) fronts the *auth-poll* endpoint, not the chat transport.

So the split is:
- **`integration_test.go` — ungated, offline, no binary.** Drives the real `Registry.Spawn` with a stub binary to pin the on-disk contract (frontmatter written, nested dir created, repo `AGENTS.md` untouched, re-delivery doesn't double the header). Runs in the refinery gates and CI.
- **`e2e_test.go` — opt-in, live** (`POGO_CURSOR_E2E=1`), following the **Codex** pattern. Byte-level request capture is unavailable, so persona delivery is asserted *behaviourally*: the persona and the repo's `AGENTS.md` each instruct a token, and the reply must carry both.

No CI job, matching Codex — GitHub CI has no Cursor account and the tests spend plan credits.

## Shared-code change

`agent.writeContextFilePrompt` gained `MkdirAll` of the context file's parent (`.cursor/rules/` doesn't exist in a fresh worktree; the bare `os.WriteFile` returned `ENOENT` and failed the spawn) and a new opt-in `PromptInjection.ContextFileHeader`. Codex leaves it empty and its `AGENTS.override.md` is written byte-for-byte as before.

## Verification

All three live tests pass against the real CLI on this box:

```
POGO_CURSOR_E2E=1 go test ./internal/cursor/ -run TestCursor -v
  PASS TestCursorEndToEnd (26.27s)          # trust hook fired; both tokens in reply;
                                            # composer settled 0 bytes; nudge -> 2nd turn
  PASS TestCursorHeadless (5.80s)
  PASS TestCursorRejectsTrustFlagInTUI (0.80s)
```

`./build.sh` and `./test.sh` green.

## Out of scope, noted for a follow-up

The persona lands as an untracked file in the polecat's worktree, so `git add -A` would commit it. **Not new** — Codex's `AGENTS.override.md` has the same posture today and neither is gitignored. Not fixed here.

## Provenance

This ticket has **no `gh:` issue ref and no `depends` chain** — it originated from a Daniel Apple Reminder (2026-07-10: *"cursor is installed and logged in now… please get this done"*), resumed after the mg-3e7c pi prerequisite landed. There is therefore no GitHub issue to close and no review ticket named in the body; dispatching a reviewer is the coordinator's call.

Work item: mg-c146

### DIFF
--- a/docs/customizing.md
+++ b/docs/customizing.md
@@ -198,7 +198,7 @@ that get expanded per-spawn:
 | `{{.Branch}}` | Target branch for refinery submit |
 | `{{.WorktreeDir}}` | Polecat's working directory: its isolated worktree, or its `~/.pogo/agents/<name>/` home dir under `--no-worktree` |
 | `{{.NoWorktree}}` | `true` when spawned with `--no-worktree` (in-place, refinery:NO) |
-| `{{.Provider}}` | Resolved harness provider id (`claude`, `codex`, `pi`; defaults to `claude`). Gate harness-specific guidance with `{{if eq .Provider "claude"}}...{{end}}` — the shipped templates use this to drop Claude-Code-isms (CronCreate, rating-modal dismissal) under other harnesses. Drop-in fragments can use it too. |
+| `{{.Provider}}` | Resolved harness provider id (`claude`, `codex`, `pi`, `cursor`; defaults to `claude`). Gate harness-specific guidance with `{{if eq .Provider "claude"}}...{{end}}` — the shipped templates use this to drop Claude-Code-isms (CronCreate, rating-modal dismissal) under other harnesses. Drop-in fragments can use it too. |
 
 ### The `worktree` knob
 
--- a/docs/investigations/README.md
+++ b/docs/investigations/README.md
@@ -12,6 +12,7 @@ item are the anchor.
 | [claude-explore-integration.md](claude-explore-integration.md) | Whether pogo's index needs special config for Claude Code's "Explore" sub-agent | Scoped, deferred (mg-39b6) |
 | [codex-e2e-validation.md](codex-e2e-validation.md) | Phase 3D end-to-end validation of the Codex CLI provider | Passed — validation record (mg-6599) |
 | [codex-nudge-calibration.md](codex-nudge-calibration.md) | Empirical nudge timing for the Codex provider | Calibration record backing `internal/codex/provider.go` (mg-7f76) |
+| [cursor-nudge-calibration.md](cursor-nudge-calibration.md) | Empirical nudge timing + the `.cursor/rules` persona-injection escape hatch for the Cursor provider | Calibration record backing `internal/cursor/provider.go` (mg-c146) |
 | [investigation-mg-06f2.md](investigation-mg-06f2.md) | Root cause: tickets archived "done" before the refinery confirmed the merge | Root-cause trace (mg-06f2) |
 | [launch-readiness-audit-2026-03-21.md](launch-readiness-audit-2026-03-21.md) | v0.2 launch-readiness audit across install, agents, refinery, release | Point-in-time audit, 2026-03-21 — no hard blockers |
 | [nudge-claude-code-workaround.md](nudge-claude-code-workaround.md) | Workarounds for nudging Claude Code through mid-session modals | Investigation; the modal watcher it scopes since shipped (mg-4421, `internal/claude/modal_hook.go`) |
--- a/docs/investigations/cursor-nudge-calibration.md
+++ b/docs/investigations/cursor-nudge-calibration.md
@@ -1,0 +1,330 @@
+# Cursor provider — nudge calibration & injection-collision investigation
+
+**Status:** calibration record for the Cursor provider (mg-c146). Evidence for
+the `cursor.Provider` values in `internal/cursor/provider.go`.
+**Harness measured:** Cursor CLI **2026.07.09-a3815c0** (`agent`, Node bundle,
+macOS), spawned in a PTY at pogo's default 200×50 winsize, authenticated via
+`agent login`.
+**Predecessors:** `docs/investigations/pi-nudge-calibration.md` (template),
+`docs/investigations/codex-nudge-calibration.md` (method).
+
+Cursor is the first pogo provider that is **closed-source SaaS**. That shapes
+both the rig and the deliverable: there is no mock-provider escape (see
+"Why the e2e is live"), so the model-facing assertions here are *behavioural*
+rather than byte-level. Everything below was measured against a live, logged-in
+CLI on 2026-07-10.
+
+## TL;DR
+
+The calibrated `NudgeProfile`:
+
+| Field | Claude | Codex | pi | **Cursor** | Why Cursor differs |
+|---|---|---|---|---|---|
+| `NeedsInitialNudge` | true | true | false | **false** | Initial task arrives via **argv** (`agent [prompt...]`). Not (only) pi's redraw problem — a typed nudge would race the *silent* workspace-trust dialog. |
+| `InitialNudgeTimeout` | 60s | 30s | 30s | **30s** | Unused while `NeedsInitialNudge` is false; retained as the measured value (trust dialog ~0.7s, composer settled ~3.0s). |
+| `SubmitTerminator` | `\r` | `\r` | `\r` | **`\r`** | A carriage return submits the composer. |
+| `SubmitDelay` | 50ms | 50ms | 50ms *(not load-bearing)* | **50ms — load-bearing** | Cursor has paste-burst detection: a combined `body+"\r"` write does **not** submit. Unlike pi. |
+| `IdleThreshold` | 2s | 2s | 2s | **2s** | Settled composer emits 0 bytes; kept uniform. The silent trust dialog is the same trap Codex has. |
+| `PromptReadySentinel` | `? for shortcuts` | *(none)* | `/ commands · ! bash` | **`Plan, search, build anything`** | Cursor's composer placeholder — the measured "input loop ready" marker. |
+
+Plus `Provider.InitialPromptViaArgv = true` (not a `NudgeProfile` field).
+
+Plus the integration decisions that are **not** `NudgeProfile` fields:
+
+- **Persona injection is `InjectContextFile` into `.cursor/rules/pogo-persona.mdc`**,
+  behind `alwaysApply: true` frontmatter. Cursor has **no**
+  `--append-system-prompt` equivalent.
+- **`--force` is the required non-interactive flag**, and **`--trust` must not
+  appear** in the interactive template — it is rejected outside `--print`.
+- **A `PostSpawnHook` is required** to dismiss the workspace-trust dialog. It
+  presses `a`, not Enter.
+
+## Method
+
+A throwaway Go PTY spike (`creack/pty`, 200×50) spawned `agent` with
+`CURSOR_DATA_DIR` pointed at a temp dir, so every run started from a
+never-trusted workspace with no chat history and the developer's real
+`~/.cursor` untouched. Behavioural probes ran through `agent -p --trust
+--output-format text`. The pipeline was then verified through pogo's real
+`agent.Registry` + `cursor.Provider` — see `internal/cursor/e2e_test.go`:
+
+```
+POGO_CURSOR_E2E=1 go test ./internal/cursor/ -run 'TestCursor' -v
+```
+
+## The injection-collision investigation (the ticket's wrinkle 1)
+
+The ticket flagged that Cursor has no `--append-system-prompt`, so the persona
+must go through a context file, which collides with a repo-owned `AGENTS.md`.
+It asked whether `.cursor/rules`-in-worktree is a viable escape hatch. **It is,
+with one non-obvious constraint.** No architect consult was needed.
+
+### There is genuinely no prompt flag
+
+Grepping the CLI bundle for `"--*prompt*"`-shaped flags yields exactly one:
+`--no-prompt`. `agent --help` confirms: no append-system-prompt, no
+system-prompt-file, no `experimental_instructions_file` analog. Cursor also
+offers **no `AGENTS.override.md`-style precedence file** — the trick Codex uses.
+So an additive injection point has to be a *different rules namespace*.
+
+### What Cursor loads as rules
+
+From the bundle's rule-watcher globs:
+
+```
+.cursor/rules/**/*.mdc     AGENTS.md     CLAUDE.md     CLAUDE.local.md     .cursorrules
+```
+
+`AGENTS.md` / `CLAUDE.md` / `CLAUDE.local.md` are all repo-owned by convention.
+`.cursor/rules/**/*.mdc` is a *directory* of independently-named rule files —
+so a uniquely-named `pogo-persona.mdc` collides with nothing a repo plausibly
+ships.
+
+### Frontmatter is load-bearing — and its absence fails silently
+
+This is the finding that would have bitten us. A behavioural probe put a
+directive in the repo's `AGENTS.md` ("include token `ALPHAGREET`") and a
+different one in `.cursor/rules/pogo-persona.mdc` ("include token
+`BRAVOGREET`"), then asked `agent -p` to "Greet me in one short sentence" —
+never mentioning rules or files, so the model had no reason to *read* them with
+a tool. Persona injected = `BRAVOGREET` present. 3 runs per variant:
+
+| `.mdc` frontmatter | Persona injected | Verdict |
+|---|---|---|
+| *(none)* | **0/3** | Silently ignored. A plain-markdown copy — what `writeContextFilePrompt` did before this ticket — delivers **nothing**. |
+| `alwaysApply: false` + `description` | 3/3 | Works *here*, but this is Cursor's "Auto Attached / Agent Requested" type: attachment is decided from the description, by the model. Not a contract. |
+| `alwaysApply: true` | **3/3** | Cursor's documented "Always" rule type. Deterministic by construction. |
+
+So the provider prepends:
+
+```yaml
+---
+description: pogo agent persona
+alwaysApply: true
+---
+```
+
+An earlier version of this probe asked the model to "list every marker in your
+rules/context", and **all three variants passed** — because Cursor has file-read
+tools and simply grepped the worktree. That confound is worth recording: a
+context-file injection test that lets the agent read the file it is testing
+proves nothing. The behavioural directive avoids it.
+
+### The collision does not materialize
+
+With the repo's `AGENTS.md` and the pogo rule both present, a single reply
+carried **both** tokens, and `AGENTS.md` was byte-identical afterwards. The
+persona is additive: Cursor folds the always-applied rule into the system prompt
+*and* still loads the repo's own context file. This is the same posture Claude
+has when reading a repo's `CLAUDE.md`.
+
+`internal/cursor/integration_test.go` pins the on-disk half of that contract
+(ungated, no binary needed); `TestCursorEndToEnd` pins the model-facing half.
+
+### Two shared-code changes it required
+
+`agent.writeContextFilePrompt` previously did a bare `os.WriteFile` of the
+prompt into `<dir>/<ContextFile>`. Cursor needs:
+
+1. **`MkdirAll` of the parent** — `.cursor/rules/` does not exist in a fresh
+   worktree, and `os.WriteFile` returns `ENOENT`, failing the spawn.
+2. **`PromptInjection.ContextFileHeader`** — the frontmatter, prepended to the
+   persona. Empty for Codex, whose `AGENTS.override.md` is plain markdown.
+
+### Residue (out of scope, noted)
+
+The persona lands as an untracked file inside the polecat's worktree, so a
+`git add -A` would commit it. This is **not new**: Codex's `AGENTS.override.md`
+has the same posture today, and neither is gitignored. Worth a follow-up ticket;
+deliberately not fixed here.
+
+## Measurements
+
+### Workspace trust → `PostSpawnHook`, and why `--trust` is banned from the template
+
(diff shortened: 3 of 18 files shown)


drellem2:
### CI was red on the first run; both failures were pre-existing flakes on `main`, not this branch. All 11 checks are now green on a **rerun with no code change**.

Recording the evidence so a reviewer doesn't have to re-derive it.

**1. `test` → `internal/project/TestDiscoverNewReposSkipsProtectedHomeDir`**
`TempDir RemoveAll cleanup: unlinkat .../001: directory not empty` — a background indexer goroutine writes into the temp dir after the test returns.

- This branch cannot causally affect it. `internal/project` imports none of the packages I changed, and the test binary built from this branch is **byte-identical** to the one built from `main`:
  ```
  go test -trimpath -c -o pb.test ./internal/project/   # this branch
  go test -trimpath -c -o pm.test ./internal/project/   # main @ d6223f7
  cmp pb.test pm.test   -> identical
  ```
- Reproduced on **pristine `main`**: 2/12 runs failed locally.

**2. `pi-e2e` → `PromptReadySentinel "/ commands · ! bash" not found`**
- Same failure on `main` at run [29093111351](https://github.com/drellem2/pogo/actions/runs/29093111351) (commit `fc5cd7c0`), which passed three minutes later at [29093259596](https://github.com/drellem2/pogo/actions/runs/29093259596). pi is pinned to `0.80.3`, so this is not a version bump — pi-tui just doesn't always paint its hint line before the assertion samples.
- Nothing in this diff touches `internal/pi`, and pi injects via `InjectAppendFlag`, so the `writeContextFilePrompt` change is a no-op on its path.

Both flakes are on `main` today and neither is fixed here (out of scope). Flagged to the coordinator; happy to file tickets or fix either under a separate work item if wanted.

**Note:** `pi-e2e`'s sentinel assertion is the same shape as the one I added for Cursor. Cursor's rendered 3/3 in calibration and again in the live e2e, but if it proves flaky in the same way, the fix is the one pi already applied to its idle check (mg-8cc0): poll for a window rather than sample once.

drellem2:
### Correction: I over-claimed on "offline e2e is impossible", and have narrowed it to what the evidence supports (3bbf528)

The coordinator asked that this claim be independently confirmed before it becomes precedent for future providers. Pressure-testing my own reasoning turned up something I had missed, so I'm correcting the record rather than defending the original wording.

**What I missed.** Cursor's bundle *does* ship a local-provider mode — and it is exactly the shape pi's offline rig needs:

```
--local-agent-base-url <url>   Provider base URL for agent-cli-local
                               (OpenAI-compatible or Anthropic Messages;
                               e.g. http://127.0.0.1:11434/v1; can also use
                               CURSOR_LOCAL_AGENT_BASE_URL or ANTHROPIC_BASE_URL)
--local-agent-api-key <key>
```

plus `CURSOR_ENABLE_AUTHLESS`, `ANTHROPIC_AUTH_TOKEN`, `CURSOR_BEDROCK_BASE_URL`. The flag is `.hideHelp()`-hidden, so `agent --help` never shows it. My original "it's closed-source protobuf, therefore impossible" skipped straight past it.

**Why the conclusion nevertheless holds.** All of it is gated on the CLI's own identity — `agent-cli-local` is a **separate executable** (`cursor-agent-local`) on its own download channel (`agent-cli-local-prod`). The `agent` binary this provider targets is `agent-cli` and enables none of it. Measured:

| Route tried on `agent` | Result |
|---|---|
| `--local-agent-base-url http://127.0.0.1:PORT/v1` | `error: unknown option '--local-agent-base-url'` |
| `CURSOR_LOCAL_AGENT_BASE_URL` + `CURSOR_ENABLE_AUTHLESS=1` | ignored — mock got **0** requests |
| `ANTHROPIC_BASE_URL` + `ANTHROPIC_AUTH_TOKEN` | ignored — mock got **0** requests |
| `CURSOR_API_BASE_URL` | ignored for chat — **0** requests (fronts `/auth/poll` only) |

**Two of my probes lied before I fixed them**, and both are now recorded in the calibration doc so nobody repeats them:

1. Asking the model to "reply with exactly `PONGMOCK`" and then seeing `PONGMOCK` proves nothing — the *real* backend obeys that instruction too, and the reply is indistinguishable from a mocked one. Redone so the mock's canned answer (`PONGMOCK`) differs from the token the prompt requests (`ALPHA`). Every route returned `ALPHA` with zero mock hits.
2. `agent --local-agent-base-url X --help` exits 0 on a flag the CLI *does not have* — `commander` prints help before validating options. That false positive is what nearly convinced me the flag worked.

**Net effect on the PR:** no code change to the provider. The e2e stays live and Codex-shaped. What changed is the *scope of the claim*: it is now "not achievable for the `agent` binary this provider targets," not "not achievable for Cursor." If pogo ever targets `cursor-agent-local`, a pi-style offline rig becomes available — different binary, different `Provider.Binary`, different ticket.

drellem2:
## Review round 1: FAIL
Reviewer: mg-b9e0 · build ticket: mg-c146 · blocking: 1 · advisory: 4

Gates: `./build.sh` green, `./test.sh` green, `gofmt`/`go vet` clean, no `replace` directives, `go.mod` untouched. The three live tests pass on my box independently (`POGO_CURSOR_E2E=1`): `TestCursorEndToEnd` 9.10s, `TestCursorHeadless` 6.63s, `TestCursorRejectsTrustFlagInTUI` 0.58s.

The implementation is sound and the test split is the right one. The single blocking finding is a factual error in the recorded justification for the scope deviation — not in the code, and not in the decision.

### Blocking

- `internal/cursor/e2e_test.go:26` and `docs/investigations/cursor-nudge-calibration.md:269` — the claim *"the only base-URL override in the bundle (`CURSOR_API_BASE_URL`) fronts the auth-poll endpoint, not the chat transport"* is false. Cursor ships a hidden `--agent-endpoint <url>` flag ("Agent endpoint URL (defaults to null, which uses the API endpoint URL)") that **does** front the chat transport.

  Evidence, against `agent` 2026.07.09-a3815c0:

  ```
  # unroutable endpoint -> the CLI tries to reach it and retries
  $ agent --agent-endpoint http://127.0.0.1:1 --trust -f -p "reply PONG"
  Connection lost, reconnecting to http://127.0.0.1:1 (attempt 1)...

  # localhost endpoint -> an h2c connection actually arrives at my listener
  $ agent --agent-endpoint http://127.0.0.1:18081 --trust -f -p "PONG"
  recv: b'PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n\x00\x00\x00\x04...'   # HTTP/2 preface + HPACK

  # by contrast, neither env var fronts chat: the turn completes normally
  $ CURSOR_API_BASE_URL=http://127.0.0.1:1 agent --trust -f -p "reply PONG"   -> PONG
  $ CURSOR_API_ENDPOINT=http://127.0.0.1:1  agent --trust -f -p "reply PONG"  -> PONG
  ```

  So the `CURSOR_API_BASE_URL` half of the sentence is right; the word **"only"** is wrong, and the implication that the transport cannot be redirected is wrong.

  **This does not overturn the decision, and the tests should not be redesigned.** The conclusion still holds for the *other* reason the doc already gives at `cursor-nudge-calibration.md:270-271`: pointing the transport at a stub means reimplementing Cursor's undocumented `agent/v1` ConnectRPC bidi-streaming service over h2c. That is the real wall, and it is disproportionate. The blocking ask is to correct the prose so the record is true and the next maintainer does not rediscover `--agent-endpoint` the hard way.

  What pass looks like: replace the "only base-URL override" claim in both places with the accurate one — the transport *is* redirectable via `--agent-endpoint`, and the blocker is the proprietary ConnectRPC/protobuf service behind it. Worth one further sentence: the bundle also carries `--authless`, `--base-url`, `CURSOR_LOCAL_AGENT_BASE_URL`, `ANTHROPIC_BASE_URL` and `CURSOR_ENABLE_AUTHLESS`, which look like an offline hatch and are not — they are gated on a `localAgentRuntime` that the public build does not supply (`agent --authless` exits with `Error: --authless can only be used with agent-cli-local`; the env vars are silently ignored and the turn still goes to the cloud).

### Advisory (non-blocking)

- `docs/polecat-permissions.md:16,36` — still says "Three providers are registered — `claude`, `codex`, `pi`" and `# default; also registered: "codex", "pi"`. Cursor is the fourth. The PR updates the equivalent lists in `docs/customizing.md:201` and `internal/config/config.go:241,287`, so this one is an oversight rather than a decision.
- `internal/codex/provider_test.go:57` — `TestProviderPromptInjection` pins Codex's `Kind` and `ContextFile` but not `ContextFileHeader`. The new regression subtest at `internal/agent/command_test.go:148` builds its own `Provider` literal, so nothing currently fails if someone sets a header on the real `codex.Provider`. One line — `if Provider.PromptInjection.ContextFileHeader != ""` — closes it on the descriptor itself.
- `internal/cursor/provider.go:107` — `NonInteractiveFlags: []string{"--force"}` makes `ValidatePolecatCommand` (`internal/agent/command.go:56`) warn on a custom `command = "agent --yolo"`, even though `--yolo` is Cursor's documented alias for `--force`. Cosmetic: a spurious warning, not a broken spawn.
- `.cursor/rules/pogo-persona.mdc` lands untracked in the polecat's worktree, so `git add -A` would commit it. Confirmed neither it nor Codex's `AGENTS.override.md` is gitignored, so this matches the shipping posture exactly as the PR body states. Worth a follow-up ticket, not this one.

### Verified, not taken on trust

- **Codex shared-code regression: clean.** Rather than rely on the hand-built literal in `command_test.go`, I drove the real `codex.Provider` through the real `Registry.Spawn`. `AGENTS.override.md` is written byte-for-byte as the persona (49 bytes, trailing-newline-free input preserved) and is the only entry in the worktree.
- **`alwaysApply: true` is guarded, in three places, two of them ungated.** Mutation-tested: `true`→`false` fails `integration_test.go:262`; dropping the header entirely fails `provider_test.go:97` and `integration_test.go:194`; removing the `MkdirAll` fails `Spawn` with `ENOENT` and takes down `integration_test.go:250`; removing the header prepend fails `internal/agent/command_test.go:212`. Every mutant is killed by tests that run in a plain `go test ./...`.
- **The silent-failure measurement reproduces.** With `personaRuleHeader = ""` and everything else unchanged, the live e2e fails exactly as claimed — `persona token "POGOPERSONAOK" never appeared in Cursor's reply` — while the repo's `AGENTS.md` token still lands. The frontmatter dependency is real, and the e2e's persona assertion is non-vacuous.
- **The ungated `integration_test.go` does cover the offline contract it claims**: nested-dir creation, frontmatter, repo `AGENTS.md` byte-identical, no header doubling on re-delivery.

One unrelated observation for the record: `internal/search/TestEvictDropsProjectState` flaked once under full-suite load with `TempDir RemoveAll cleanup: directory not empty`. It passed on rerun and on 3× isolated runs on both this branch and `main`, and this PR does not touch `internal/search`. Pre-existing, not attributable here.

drellem2:
### Self-found bug, fixed while waiting on review (`57e07d2`): an echoed task could be mistaken for the trust dialog

Not a review finding — I went looking for one in my own diff. Flagging it explicitly so the reviewer doesn't have to rediscover it, and because the failure is subtle.

**The bug.** `TrustDialogHook` matches on PTY *text*, and Cursor echoes the argv-delivered task into the TUI. A work item whose body merely **quotes** the dialog matches the marker:

```
"Investigate the workspace-trust dialog: it offers [a] Trust this workspace and [q] Quit"
   -> collapsed -> matches (?i)trustthisworkspace
```

**mg-c146's own ticket body would have tripped it.**

On a fresh worktree this is harmless: the real dialog renders first, the hook dismisses it and returns. But **Cursor persists trust per workspace**, so a respawn into the same `Dir` (`Registry.Respawn`, `RestartOnCrash`) shows no dialog at all. The hook would poll its full 12s, match the echoed task, and press `a` into the **live composer** — the next `Agent.Nudge` then arrives with a stray `a` prefixed to its body.

**The fix.** Guard on the composer placeholder: if `Plan, search, build anything` has rendered, the input loop is up and nothing is blocking, so return without sending. Verified **3/3** against a live, never-dismissed dialog that the placeholder is absent for as long as the dialog is up — so the guard cannot suppress a real dismissal. It also makes the hook return immediately on an already-trusted spawn instead of polling out the timeout.

The sentinel is now one `const` shared by `Provider.Nudge.PromptReadySentinel` and the hook, so they cannot drift.

**Verification**
- New unit tests pin *both* halves: `TestEchoedTaskLooksLikeTheTrustDialog` asserts the echoed task **does** match the marker (so the marker alone is insufficient), and that `composerReady` gates the hook off; `TestComposerReady` covers dialog/banner/post-turn screens.
- The live e2e still logs `detected Cursor workspace-trust dialog, auto-accepting` and passes with the guard in place.
- `TestCursorEndToEnd` run **5×** total against the real CLI today: 5/5 pass. (I also measured the `PromptReadySentinel` assertion I'd flagged as a possible flake risk in my earlier comment — 5/5, no de-flake needed.)

Codex's `TrustDialogHook` has the same text-matching exposure in principle. I have **not** touched it — out of scope — but it's worth a follow-up ticket.

drellem2:
## Review round 2: PASS
Reviewer: mg-b9e0 · build ticket: mg-c146 · blocking: 0 · advisory: 5

Re-reviewed at `3bbf528`. Round 1 was against `5b65418`; the builder had already pushed a self-correction narrowing the offline-e2e claim. The diff `5b65418..3bbf528` is documentation and one comment block — **zero non-comment lines changed** — so the round-1 code verification carries forward unchanged.

Round 1's blocking finding is resolved. The sentence *"the only base-URL override in the bundle"* no longer appears anywhere in the tree, and the conclusion is now stated at the strength the evidence supports and scoped to `agent-cli` rather than to Cursor in perpetuity.

### Gates
`./build.sh` exit 0 · `./test.sh` exit 0 · `gofmt -l` clean · `go vet` clean · no `replace` directives · `go.mod`/`go.sum` untouched.

### The corrected claims, verified independently

I re-measured each row of the doc's route table against the real binary rather than accepting it, and avoided the `--help` trap the builder documented:

```
$ agent --local-agent-base-url http://127.0.0.1:18080/v1 --trust -f -p "reply ALPHA"
error: unknown option '--local-agent-base-url'

$ agent --authless --trust -f -p "reply ALPHA"
Error: --authless can only be used with agent-cli-local

# CURSOR_ENABLE_AUTHLESS=1 + CURSOR_LOCAL_AGENT_BASE_URL + ANTHROPIC_BASE_URL, all -> local listener:
# 0 requests reached the listener; the real backend answered the turn.

# bundle strings confirm the separate distribution:
#   agent-cli-local, agent-cli-local-prod, cursor-agent-local
```

All four hold. The gate in the bundle is `void 0 !== localAgentRuntime`, which the public `agent-cli` build never supplies — consistent with the builder's account.

### Advisory (non-blocking)

- `docs/investigations/cursor-nudge-calibration.md:305-310` and `internal/cursor/e2e_test.go:24` — the route table omits the one transport override that **is** honored by `agent-cli`: the hidden `--agent-endpoint <url>`. It genuinely redirects the chat transport:

  ```
  $ agent --agent-endpoint http://127.0.0.1:1 --trust -f -p "reply PONG"
  Connection lost, reconnecting to http://127.0.0.1:1 (attempt 1)...   # no reply; turn never completes

  $ agent --agent-endpoint http://127.0.0.1:18081 --trust -f -p "PONG"
  # raw listener received: b'PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n\x00\x00\x00\x04...'  (h2c preface + HPACK)
  ```

  Without the flag the same turns complete normally, so the flag is what breaks the chat path. This does **not** disturb the conclusion — that endpoint still speaks the proprietary ConnectRPC/protobuf service the doc already calls absurd to reimplement, and the table is scoped to OpenAI-compatible hatches. It is worth a row and one sentence, because it is exactly what the next maintainer will find and wonder about. In the same breath, `:322` reads more precisely as "not achievable at proportionate cost" than "not achievable".
- `docs/polecat-permissions.md:16,36` — still "Three providers are registered — `claude`, `codex`, `pi`" and `# default; also registered: "codex", "pi"`. Cursor is the fourth. The equivalent lists in `docs/customizing.md:201` and `internal/config/config.go:241,287` were updated.
- `internal/codex/provider_test.go:57` — pins Codex's `Kind` and `ContextFile` but not `ContextFileHeader`. The new regression subtest at `internal/agent/command_test.go:148` builds its own `Provider` literal, so nothing fails today if a header is set on the real `codex.Provider`. One line closes it on the descriptor.
- `internal/cursor/provider.go:107` — `NonInteractiveFlags: []string{"--force"}` makes `ValidatePolecatCommand` (`internal/agent/command.go:56`) warn spuriously on `command = "agent --yolo"`, Cursor's documented alias. Cosmetic.
- `.cursor/rules/pogo-persona.mdc` is untracked in the polecat worktree, so `git add -A` would commit it. Identical to Codex's `AGENTS.override.md` posture today; neither is gitignored. Follow-up ticket.

### What was verified, not taken on trust

- **Codex shared-code regression: clean.** I did not rely on the hand-built literal in `command_test.go`. I drove the real `codex.Provider` through the real `Registry.Spawn`: `AGENTS.override.md` is byte-for-byte the persona (49 bytes, no-trailing-newline input preserved) and is the only entry in the worktree.
- **`alwaysApply: true` is guarded in three places, two of them ungated.** Mutation-tested, every mutant killed by a plain `go test ./...`: `true`→`false` fails `integration_test.go:262`; header removed fails `provider_test.go:97` + `integration_test.go:194`; `MkdirAll` removed fails `Spawn` with `ENOENT` at `integration_test.go:250`; header prepend removed fails `command_test.go:212`.
- **The frontmatter silent-failure claim reproduces, and the probe is trap-proof.** With `personaRuleHeader = ""` the live e2e failed with `persona token "POGOPERSONAOK" never appeared`, while the repo's `AGENTS.md` token still landed. The decisive control: in the mutant arm the rule file *still sits on disk containing that token* — only the frontmatter is gone. The model did not emit it. So the token in the passing arm cannot have come from a file read; it came through rule injection. Expected answer and requested answer are separated by route, not by wording.
- **The ungated `integration_test.go` covers the offline contract it advertises**: nested-dir creation, frontmatter, repo `AGENTS.md` byte-identical, no header doubling on re-delivery.
- **Live suite passes on my box** (`POGO_CURSOR_E2E=1`): `TestCursorEndToEnd` 9.10s, `TestCursorHeadless` 6.63s, `TestCursorRejectsTrustFlagInTUI` 0.58s.

Unrelated, for the record: `internal/search/TestEvictDropsProjectState` flaked once under full-suite load (`TempDir RemoveAll cleanup: directory not empty`). It passed on rerun and on 3× isolated runs on both this branch and `main`, and this PR does not touch `internal/search`. Pre-existing, not attributable here.

drellem2:
Closing — this work is already done. The Cursor provider landed on main and shipped in v0.5.0 (via the mg-c146 commit series); `internal/cursor/` is present and registered. This branch is now behind main, so there's nothing left to merge here. Any remaining polish is tracked separately (mg-150a). Thanks.

PR #77, opened 2026-07-10
fix(agent): match multiple prompt-ready markers so v2.1.x spawns don't burn 60s (#76)

### LINKED ISSUE #76: Polecat cold-start: stale prompt-ready sentinel wastes a flat 60s on Claude Code v2.1.x
## Describe the Bug

Every polecat spawn wastes a flat **60 s** (the full `InitialNudgeTimeout`) as pure dead time before its first "do the work" nudge is delivered.

The initial nudge waits (wait-ready mode) for the provider's `PromptReadySentinel` to appear in the harness PTY. For the Claude provider that sentinel is the composer hint `"? for shortcuts"` (`internal/agent/provider.go`, `DefaultNudgeProfile`). **Claude Code v2.1.x removed that hint** — the empty composer now shows a rotating `Try "…"` placeholder and the footer shows a `shift+tab to cycle` mode bar instead. So the sentinel never matches, `WaitForReady` never opens its gate, and every spawn falls through to the full `InitialNudgeTimeout` before delivering best-effort.

This is silent: it only surfaces as a fixed ~60 s cold-start on every worker, plus a log line per spawn.

## To Reproduce

1. Run any pogo instance whose Claude Code harness is **v2.1.x** (reproduced on v2.1.201).
2. Spawn a polecat (or any crew agent that gets an initial nudge).
3. Observe in `docker logs` / pogod stderr, once per spawn:
   ```
   agent <name>: prompt-ready sentinel "? for shortcuts" not seen within 1m0s; delivering nudge best-effort (sentinel may be stale)
   ```
4. In `~/.pogo/events.log`, the `agent_spawned` → first `nudge_sent` gap is a near-exact 60 s.

Measured on a live instance: **12 of 12** spawns hit the "sentinel not seen within 1m0s" path — 0 hit "seen but never settled". So it is purely the sentinel string being stale, not a quiescence problem (the idle v2.1.x composer is byte-stable, so `IsIdle(2s)` fires fine once the gate can open).

## Expected Behavior

Readiness is detected within a few seconds of the composer rendering, so the initial nudge is delivered promptly (cold-start ≈ boot-to-composer + quiescence, ~2–3 s), rather than always timing out at 60 s.

## Actual Behavior

The sentinel never matches on v2.1.x, so `WaitForReady` always times out at `InitialNudgeTimeout` (60 s). Every worker pays a flat 60 s tax.

## Environment

- OS: Linux (aarch64) container; also applies to any host running Claude Code v2.1.x
- Go version: 1.25.0
- Pogo version: 0.3.0 (build=dev) live; reproduced against `main` @ d2c4890
- Shell: bash
- Editor/integration: Claude Code harness **v2.1.201**

## Logs

```
2026/07/10 09:12:52 agent bench-v2-lookup: prompt-ready sentinel "? for shortcuts" not seen within 1m0s; delivering nudge best-effort (sentinel may be stale)
```
(one per spawn; `agent_spawned`→`nudge_sent` gap = 60.0 s)

## Additional Context

Root-caused while benchmarking Slack-tag→response latency for the payit agent prototype (pogo-slack-bridge). This 60 s cold-start is the second-largest latency term for worker tasks.

**Fix incoming (PR):** make readiness detection match a *set* of markers instead of one exact string — add `PromptReadyAlternates` and the spaceless v2.1.x markers `shift+tabtocycle` and `Try"` (spaces vanish under ANSI-strip because the footer is cursor-column-addressed). Verified against a real captured v2.1.201 screen; readiness then detected at ~0.55 s in an isolated instance (cold-start 60 s → ~2.5 s).

**Request:** once the fix lands, please cut a tagged release — we want to pin/consume it from the payit agent prototype rather than track `main`.

bath-tub:
Fixes #76.

## Problem

Every polecat spawn wastes a flat **60 s** (`InitialNudgeTimeout`) as dead time. The initial nudge waits (wait-ready) for the Claude provider's `PromptReadySentinel` — `"? for shortcuts"` — but **Claude Code v2.1.x removed that composer hint**. It now renders a rotating `Try "…"` placeholder and a `shift+tab to cycle` mode-bar footer instead, so the sentinel never matches, `WaitForReady`'s gate never opens, and every spawn falls through to the full 60 s timeout before delivering best-effort. Observed **12/12** spawns hitting `sentinel not seen within 1m0s`.

## Fix

Make readiness detection match a **set** of markers instead of one brittle exact string:

- New `NudgeProfile.PromptReadyAlternates []string`; `WaitForReady` now takes `[]string` and opens its gate when the primary sentinel **OR any alternate** appears.
- `DefaultNudgeProfile` (Claude) keeps `"? for shortcuts"` (older harnesses) and adds the v2.1.x markers **`shift+tabtocycle`** and **`Try"`**. Note the missing spaces: the v2.1.x footer is rendered with per-word cursor-column moves (`ESC[<n>G`), so after ANSI-stripping the words concatenate — the live PTY yields `accepteditson(shift+tabtocycle)…` and `❯ Try"…"`. `shift+tabtocycle` is mode-invariant; `Try"` marks the empty composer a fresh spawn starts at.
- Backward compatible: `pi`/`codex` providers keep their single sentinels unchanged.

## Evidence

- New guard test `TestDefaultSentinelsMatchClaudeV2_1` asserts the configured markers against a **real captured v2.1.201 polecat screen** (`testdata/`), run through the same `StripANSI` the gate uses — and asserts the legacy sentinel is genuinely absent (the regression).
- Measured on an isolated patched pogod (`POGO_HOME`-scoped, alt port): composer/sentinel renders at **~0.55 s** (3 trials: 548/547/551 ms) and the idle composer is byte-stable, so the gate opens at **~2.5 s** — vs the stale-sentinel binary's flat **60 s**. Same TUI, same timings; only the sentinel string differs.
- `go test ./internal/agent/` green.

## Release request

Once this merges, **could you cut a tagged release?** We want to pin/consume this from the payit agent prototype (via pogo-slack-bridge) rather than track `main`. The 60 s cold-start is the second-largest latency term in that path, so a release lets us adopt the fix cleanly.


drellem2:
## Review: PASS (verified against real Claude Code v2.1.206 PTY bytes)

Reviewer: mg-a6c9 · round 1 · blocking: 1 (repaired in merge-prep) · advisory: 2

### Verification

The fix was checked against the installed harness (Claude Code **2.1.206**), not the diff. Idle composers were captured over a PTY and ANSI-stripped through the same `StripANSI` the readiness gate uses, in both the real spawn mode (`--dangerously-skip-permissions`) and manual mode:

| marker | v2.1.201 accept-edits | v2.1.206 manual | v2.1.206 bypass (real spawn) |
|---|---|---|---|
| `? for shortcuts` (legacy) | absent | absent | absent |
| `?forshortcuts` (spaceless) | absent | present | absent |
| `shift+tabtocycle` | present | absent | present |
| `Try"` | present | present | present |

In the real spawn condition (bypass-permissions), `shift+tabtocycle` and `Try"` both match while the legacy sentinel is absent — so the gate opens on v2.1.206 instead of falling through to the 60s timeout. Regression and fix both reproduce.

The mg-ce61 startup-wedge protection is preserved: `WaitForReady` still requires a string marker to be seen *and then* quiescence (`seen && IsIdle`) — it does not drop to a quiescence-only gate. Empty-sentinel providers (codex) are unaffected; they take the wait-idle branch (`NudgeWithMode` gates the wait-ready path on a non-empty primary sentinel).

### Blocking (repaired in the prepared merge branch)

- `internal/claude/provider_test.go:74`, `internal/pi/provider_test.go:97,145`, `internal/codex/provider_test.go:97,131` — adding the `[]string` field `PromptReadyAlternates` to `NudgeProfile` makes the struct non-comparable with `==`. The PR updated the one comparison it hit under `go test ./internal/agent/` but missed three sibling packages, so `go test ./...` fails to compile. Fixed by switching those to `reflect.DeepEqual` (mirroring the fix already applied in `provider_resolution_test.go`). No behavior change.

### Refinement folded

- `shift+tabtocycle` is **not** mode-invariant (absent in manual mode, per the table). The marker set is reordered structural-first and gains the spaceless `?forshortcuts`: `{shift+tabtocycle, ?forshortcuts, Try"}`. `Try"` is kept last as the only mode-invariant fallback. The misleading "mode-invariant" code comment is corrected.

### Advisory (non-blocking)

- A profile with an empty `PromptReadySentinel` but non-empty `PromptReadyAlternates` would skip the wait-ready path entirely (`nudge.go:150` gates on the primary only). No current provider hits this.
- `pi`'s sentinel `"/ commands · ! bash"` is spaced and shares the same column-move space-collapse brittleness class; out of scope here, worth a follow-up.

Unrelated: `internal/project` `TestFileInNewProjectAddsProject` fails on `main` independent of this change (environmental — runs against a live `POGO_HOME`); not introduced by this PR.

PR #82, opened 2026-07-14
build: pogod concurrent-spawn init-stall — auto-renudge on hard-unstarted signal (drellem2/macguffin#24)

drellem2:
## Summary

Productizes the mayor's manual `nudge "1"` workaround into a pogod daemon guarantee (approved triage OPTION 1). After a polecat's initial nudge is delivered, pogod verifies the **HARD started-signal** — the agent's mg work item leaving `available/` — and, if it is still absent within a bounded ~25s window, re-delivers a bare submit terminator (CR) to flush a paste-buffered kickoff, retrying a bounded number of times.

### Root cause (from triage mg-5ece)
Under a concurrent spawn wave a CPU-starved node/Ink process emits no PTY writes, so `WaitForReady`'s idle gate (`IsIdle` = time since last PTY write) misreads starvation-silence as input-loop-ready. The kickoff nudge is delivered before Claude Code is listening, piles in the kernel input buffer, and Ink absorbs it as one paste block whose CR never re-tokenizes as a submit (mg-ce61). The agent is alive+running but never claims its item.

### Design fidelity to the approved recommendation
- **HARD signal, not quiescence.** Gates on the work-item claim (item present in `available/` ⇒ unstarted), never on output quiescence — a quiescence re-check would reproduce the same false-idle failure the whole bug is about.
- **Bare CR payload** (`a.Nudge("")` writes only the provider's `SubmitTerminator`) — least likely to inject a stray char into a working agent. Delivered ONLY while the item is provably unclaimed.
- **Bounded retries** (`DefaultStartVerifyMaxAttempts = 3`, `DefaultStartVerifyDelay = 25s`).
- **Failure-mode-agnostic** — recovers whether the cause is the false-idle gate, a stale sentinel (mg-ce4c), or the paste pileup. Maps to mg-ce61 alternative (b).
- **Scope held:** the mayor's MAX-2 concurrent-spawn cap + nudge-1 check (mayor.md:188-200) is intentionally **not** removed — that retirement is a follow-up gated on real-spawn-wave verification of this fix.

### Changes
- `internal/agent/startverify.go` (new): `StartVerifier` type + `verifyStartAndRenudge` watcher + `auto_renudge` event; Registry `SetStartVerifier` DI setter (mirrors `SetMailCheckRegistrar`).
- `internal/agent/agent.go`: the Spawn initial-nudge goroutine drives the watcher after delivering the kickoff.
- `internal/agent/nudge.go`: documents the `IsIdle` false-idle-under-starvation caveat and points at the backstop (no logic change — Option 1 is deliberately gate-agnostic).
- `cmd/pogod/main.go`: wires the macguffin-backed verifier (scans `~/.macguffin/work/available/` via `workitem.ListFrom`).
- `docs/event-log.md`, `CHANGELOG.md`: catalog the new `auto_renudge` event.

### Verification
- `./build.sh` green (fmt + full `go test ./...` + build).
- New unit tests (`internal/agent/startverify_test.go`) cover: unstarted→bounded renudge; started→no renudge; recovery-after-one-CR; query-error→no blind renudge; no-work-item/nil-verifier no-op; agent-exit aborts the wait; default window is in the ~20-30s band. Uses an `os.Pipe` PTY master to observe the exact bare-CR bytes; `-race` clean.
- **End-to-end through `Spawn`** (`TestSpawnDrivesStartVerify`): a real spawned PTY agent emits the prompt-ready sentinel then idles, the wait-ready initial nudge fires, and the verify loop is driven to its bounded attempts — proving the Spawn→verify wiring, not just the stubbed unit.
- `cmd/pogod/startverify_test.go`: the verifier maps `available/`→unstarted, `claimed/`/absent→started, missing-root→started-no-error.

**Caveat carried from the ticket:** true recovery under the *load-dependent* trigger needs a real spawn wave or a CPU-contention harness; unit/integration tests exercise the logic and wiring but cannot reproduce the starvation timing. The mayor's existing cap remains as belt-and-suspenders until a live wave confirms recovery.

Resolves drellem2/macguffin#24

Approved triage recommendation: mg-5ece (Daniel GO 2026-07-13)
Work item: mg-feb3


drellem2:
## Review round 1: FAIL
Reviewer: mg-fc92 · build ticket: mg-feb3 · blocking: 1 · advisory: 0

Design is faithful and the code is clean: the started-signal is the HARD work-item-claim signal (never quiescence), the renudge is bounded (3×) and bare-CR only, it re-checks `started` before every keystroke and skips on an unreadable mg state, and the mayor's MAX-2 cap + nudge-1 workaround is correctly left in place. Build and full `go test ./...` are green and `-race` clean. The one blocker is verification.

### Blocking
- **Verification does not clear the load-dependent bar.** `internal/agent/startverify.go:139` (`a.Nudge("")`) is the load-bearing recovery mechanism, and its efficacy is unproven. The PR's own Verification section concedes there is no real spawn wave or CPU-contention harness — evidence is unit tests plus a wiring test (`TestSpawnDrivesStartVerify`) that drives the loop against a stubbed always-unstarted verifier and a `cat` PTY. That proves the CR is *delivered*; it does not prove the CR *recovers* a wedged agent. Critically, the only empirically-observed recovery used `nudge "1"`; the ticket sanctions bare CR as primary but conditions it — "fall back to '1' only if CR proves unreliable" — and CR reliability at flushing a real paste-buffered Ink kickoff has not been established. A self-heal that has never been shown to heal cannot be relied on (and would later be used to justify retiring the manual cap).

  **What pass looks like:** demonstrate end-to-end recovery under a real or contention-simulated wedge — spawn a wave of real polecats (or starve one harness's CPU to reproduce the false-idle), observe an `auto_renudge` event fire, and confirm the agent then claims its item. At minimum establish that the bare CR actually flushes a wedged agent; if it does not, fall back to `1` per the ticket. If a load-dependent repro genuinely can't be produced, say so and escalate for a call rather than re-asserting green tests.

drellem2:
## Round 2 — verification blocker addressed (commit 3fda02c)

The round-1 blocker was that green tests prove *delivery*, not *recovery*, and bare CR deviates from the field-proven `nudge "1"` without established reliability. I took reviewer **option 3** — a real Claude Code / Ink efficacy test — and reproduced the wedge deterministically.

**Method:** drove the real `claude` binary (v2.1.207) in a PTY and induced the *same kernel-buffer pileup* the CPU-starvation path produces, by writing the 3088-byte kickoff burst **at t=0, before Ink arms its input reader** (the mg-ce61 mechanism; the CPU-starvation case just reaches the same pileup a different way). Writing the identical burst *after* the composer is ready does **not** wedge — the `\r` submits normally — which confirms the bug is the pre-arm pileup, not burst size.

**Result (3/3 identical runs):**

```
wedge reproduced (burst not auto-submitted): true   ← kickoff sits UNSENT in the composer
bare CR recovered (submitted the buffer):    true   ← a bare \r submits it; Claude begins the turn
```

On arm, the entire kickoff sat unsent in the composer (`❯ VERIFICATION PING for mg-feb3 …`, no processing indicator). A single **bare CR** then flushed it — `esc to interrupt` / `thinking` / token counter, and Claude replied ("Acknowledged — no action taken … automated verification ping (mg-feb3)").

**Conclusion:** the fix's core assumption holds against a real Ink input box — bare CR *recovers*, not merely delivers. The `"1"` fallback the ticket sanctioned ("only if CR proves unreliable") is **not needed**, and bare CR is preferable (no stray char in the submitted prompt).

Full method, verdict, and caveats: `docs/investigations/renudge-efficacy-2026-07-14.md`.

**Scope unchanged:** this reproduces the downstream wedge + recovery (failure-mode-agnostic), not the upstream live-spawn-wave CPU timing — still a follow-up gate, which is why the mayor's MAX-2 cap + unstarted-check stays retained per the ticket. No code change this round; the round-1 implementation is validated as-is. `./build.sh` green.

drellem2:
## Review round 2: PASS
Reviewer: mg-fc92 · build ticket: mg-feb3 · blocking: 0 · advisory: 2

The round-1 verification blocker is resolved. `docs/investigations/renudge-efficacy-2026-07-14.md` drives a real Claude Code v2.1.207 PTY session and closes the delivery-vs-recovery gap: it reproduces the genuine wedge (a pre-arm kernel-buffer pileup — the same downstream state CPU starvation produces — leaves the kickoff sitting unsent in the Ink composer) and shows a **bare CR submits it** (recovery, not just delivery), 3/3 runs, with a sound control (an after-arm write does not wedge, isolating the pileup as the cause). The payload tested is the exact `\r` the fix sends (`provider.go:190` → `a.Nudge("")`). Detection is unit-tested, recovery is now empirically established, and the design is failure-mode-agnostic, so the downstream reproduction is a legitimate stand-in for the upstream trigger.

Round-2 delta is docs-only; the round-1 implementation is unchanged. Build and full `go test ./...` remain green, `-race` clean. Design stays faithful: HARD claim signal (never quiescence), bounded bare-CR renudge that never fires against a started agent, mayor MAX-2 cap retained.

### Advisory (non-blocking)
- `docs/investigations/renudge-efficacy-2026-07-14.md:44` — the efficacy harness is intentionally not committed (billable real-Claude session); the method is documented and reproducible, but the result is not re-runnable in CI. Accepted.
- `docs/investigations/renudge-efficacy-2026-07-14.md:77` — the end-to-end live multi-polecat spawn-wave confirmation remains an ungated follow-up before the mayor's MAX-2 cap can be retired. Correctly scoped out of this ticket; the retained cap is the belt-and-suspenders until then.

drellem2:
## Head-to-head: bare CR vs. `"1"`+CR (commit 0aa8477)

The mayor supplied live field data: the wedge reproduced in a 3-concurrent-spawn wave (the reviewer's own agent fc92 stalled ~2m44s, recovered by `pogo nudge fc92 "1"`), and only `"1"`+CR was field-confirmed — so whether **bare CR alone** flushes Ink's paste buffer was genuinely open. I ran the direct head-to-head in the same real-`claude` PTY harness:

| Recovery payload | Runs | Wedge reproduced | Recovered |
|------------------|------|------------------|-----------|
| bare `\r` (fix) | 8 | 8/8 | **8/8** |
| `"1"` then `\r` (field-confirmed) | 5 | 5/5 | **5/5** |

_(One early `"1"`+CR run first read as non-recovered — a detector false-negative on Claude Code's rotating spinner words ("Smooshing"/"Seasoning"/…); hardened matcher confirmed 3/3 on re-run.)_

**Findings:**
1. **Bare CR alone reliably recovers (8/8)** — settles the open question; the `"1"` fallback the ticket sanctioned is not needed.
2. **`"1"`+CR also recovers (5/5)**, which cross-validates the harness: it reproduces the same recoverable state the mayor observed live.
3. **Bare CR is preferable.** `"1"` is appended to the buffered kickoff and submitted *with* it (stray char corrupts the prompt); bare CR submits the kickoff clean. Same recovery, no contamination — exactly why the ticket chose bare CR as primary.

Full data + method: `docs/investigations/renudge-efficacy-2026-07-14.md`. `./build.sh` green. No production code change — the round-1 implementation is validated as-is.

PR #84, opened 2026-07-15
build: polecat mail-check auto-register hardening

### LINKED ISSUE #83: mg-e633: polecat mail-check auto-register fails silently — polecat spawns deaf to mail
## Summary

The mg-e633 spawn-time mail-check auto-registration (`registerPolecatMailCheck`, `internal/agent/api.go:1030`) is **best-effort and silent on failure**. When its `Add` doesn't land, the polecat spawns with **no mail-check schedule at all** — the mayor can't reach it, and nothing surfaces the miss. Operators only notice when a polecat sits idle and has to be nudged by hand.

## Evidence (real incident)

Polecat `cat-d262` (worldpay_integration_service), v0.5.0 host:

- `agent_spawned` fired → spawn succeeded → `registerPolecatMailCheck` was invoked (it runs unconditionally right after a successful `Spawn`).
- Across the entire event log, `mail-check-d262` **never fired** (`scheduler_fire_delivered`) and was **never removed** (`schedule_removed`) — so the schedule never existed. Not a reap; the `Add` never took.
- Behaviorally: after the start-nudge at 01:35Z, the polecat got **no** automatic 10-min mail-check nudges for 30 min, until a human manually nudged it at 02:05Z.
- **Not systemic:** other polecats on the same host got working mail-checks (e.g. `mail-check-mg-72b4` fired). So this is an intermittent `Add` failure, not a dead feature.
- The failure reason would have been in `pogod` stderr (`registerPolecatMailCheck` logs \"mail-check schedule registration failed: …\"), but that log had already rotated — which is itself the problem: the only signal is a best-effort log line.

## Why it hurts

`registerPolecatMailCheck` is documented as best-effort (\"a registration failure is logged, not fatal\"). But for a polecat, the mail-check *is* the reachability channel — a silent miss means the mayor/reviewer loop stalls with no signal. Best-effort is the wrong contract for the primary reachability mechanism.

## Requests

1. **Make the miss loud.** Emit a structured event (e.g. `schedule_register_failed`) when the auto-register `Add` fails or no-ops, so it's visible in the event log, not just a rotating stderr line.
2. **Verify-after-register.** After `Add`, confirm the entry exists (list by `(agent, mail-check-<id>)`); retry once and, if still absent, emit the event above and optionally nudge the mayor.
3. **Root-cause the intermittent `Add` failure.** Prime suspects worth checking: work-item-id normalization (the failing spawn used a bare `d262` id; the working ones were `mg-`-prefixed), a validation/uniqueness reject, or a liveness-gated reject at the spawn instant (registry may not yet consider the agent alive).

## Workaround in place

Downstream templates now register the mail-check explicitly via `pogo schedule` in step 2 (rather than relying on the auto-register), so a silent miss no longer leaves the polecat deaf. This issue tracks hardening the platform so the auto-register itself is reliable and observable.

drellem2:
Hardens `registerPolecatMailCheck` so a polecat's mail-check loop — its **primary reachability channel**, which drives the modify↔review loop — is no longer registered best-effort-and-silent. Two distinct suspects, two distinct mechanisms (per the approved triage spec, mg-6fe0):

**1. LOUD ALWAYS (load-bearing).** Emit a structured `schedule_register_failed` event (fields: `agent`, `schedule_id`, `reason`) on **both** the register-error path and the previously-silent nil-registrar no-op. Written via the scheduler's **own-root** event path (`scheduler.EmitScheduleRegisterFailedTo`), never the global `events.log` (mg-e06d). The failure reporter is wired into the agent registry **independently of the mail-check registrar**, so it still fires when the registrar is nil — i.e. when pogod's scheduler failed to load at startup and `SetMailCheckRegistrar` was never called.

**2. VERIFY-AFTER-REGISTER + retry-once.** The pogod `mailCheckRegistrar` adapter confirms the `(agent, mail-check-<id>)` entry via `Scheduler.Get` after `Add`, and retries **once** if absent. This recovers the transient persist-IO suspect. It cannot fix nil-registrar (the registrar is nil synchronously in the same call) — that path is covered **solely** by event emission.

**3. ESCALATE to a mayor nudge ONLY on the persistent post-retry failure** (a live polecat left with no reachability channel). The benign startup nil-registrar is event-only — **no** nudge (noise).

Distinct `reason` values (`nil_registrar` vs `register_error: ...`) keep a reader from mis-reading the telemetry as retry having caught a nil registrar — two suspects, two mechanisms.

**Non-fatal to spawn throughout** — registration failure never kills a running polecat.

### Affected areas
- `internal/agent/mailcheck.go` — `ScheduleRegisterFailureReporter` interface + wiring; reworked `registerPolecatMailCheck`.
- `internal/agent/agent.go` — reporter field on `Registry`.
- `internal/scheduler/scheduler.go` — `schedule_register_failed` emitter + `EventLogPath` helper (own-root path), alongside `emitSchedulerRemovalEvent`.
- `cmd/pogod/main.go` — adapter verify+retry, mayor escalator, own-root reporter wired even when the scheduler fails to load.
- Tests: `internal/agent/mailcheck_test.go` (louder-but-still-non-fatal; new nil-registrar case), `cmd/pogod/mailcheck_test.go` (happy-path no-escalation; persistent-failure escalates), `internal/scheduler/register_events_test.go` (emitter + own-root path).

Full `./build.sh` gate green; affected packages clean under `-race`.

Resolves drellem2/pogo#83

Approved triage recommendation: mg-6fe0 (IMPLEMENT / bug — reliability+observability hardening)
Work item: mg-75d9

### DIFF
--- a/cmd/pogod/mailcheck_test.go
+++ b/cmd/pogod/mailcheck_test.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"os"
 	"path/filepath"
 	"testing"
 	"time"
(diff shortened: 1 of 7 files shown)


drellem2:
## Review round 1: PASS
Reviewer: mg-2997 · build ticket: mg-75d9 · blocking: 0 · advisory: 1

Reviewed against the approved triage recommendation (mg-6fe0). All six pm-pogo signed-off constraints hold; build gate green (`./build.sh` clean), affected packages clean under `-race`.

### Verified
- `schedule_register_failed` emitted on **both** the Add-error path (`register_error: ...`) and the nil-registrar no-op (`nil_registrar`) — `internal/agent/mailcheck.go:130`, `:145`. Distinct `reason` values keep the two suspects readable.
- Event routed through the scheduler's **own-root** path — `scheduler.EmitScheduleRegisterFailedTo` → `events.EmitTo(EventLogPath(schedPath))`, never global `events.log` (mg-e06d).
- Verify-after-register + retry-once (`addAndVerify`: `Add`→`Get`) scoped to the transient persist path in the pogod adapter — `cmd/pogod/main.go:180`. `Scheduler.Add` rolls back in-memory state on persist failure (`internal/scheduler/scheduler.go:325`), so the retry re-adds cleanly. Correctly does **not** touch the nil-registrar path.
- Mayor-nudge (`newMailCheckReachabilityEscalator`) fires **only** on persistent post-retry failure — `cmd/pogod/main.go:171`; never on the benign nil-registrar no-op (event-only).
- Spawn stays non-fatal on register failure — `registerPolecatMailCheck` returns void; `TestSpawnPolecatMailCheckFailureNonFatal` and `TestSpawnPolecatNilRegistrarReportsFailure` confirm the polecat spawns under both failure modes.
- `TestSpawnPolecatMailCheckFailureNonFatal` updated: louder (asserts the `schedule_register_failed` report) but still non-fatal.

### Advisory (non-blocking)
- `cmd/pogod/mailcheck_test.go:99` — `TestMailCheckRegistrar_PersistentFailureEscalates` forces persist failure via `os.Chmod(dir, 0o555)`. Under a uid-0 runner the read-only bit is ignored, so `Add` would succeed and the test would fail (loudly, not silently) rather than exercise the escalation path. Fine for normal agent/CI environments; noting only for portability.

Issue #81, opened 2026-07-13
Refinery closing the work item terminates the polecat before post-merge steps (e.g. opening a PR)


drellem2:
## Observed behavior

When a polecat submits its branch to the refinery (PR flow, dispatched with `--branch`), the refinery merges the branch and then marks the work item done itself — the done sidecar records `{"completed_by":"refinery"}`. As soon as the item is `done`, the polecat exits, **before it can run its own post-merge steps** (e.g. `gh pr create`, verifying checks, mailing the PR URL).

## Impact (observed twice)

Porting org-template defaults to the PR-gated `payitgov/.pogo` repo, two `--branch` polecats (work items mg-ddac, mg-a647) each:
- correctly built the integration branch with the right signed commit, then
- exited immediately after the refinery merged + closed the item — **the PR was never opened**, yet the item showed `done`.

A third, dedicated polecat had to be dispatched just to `git push` + `gh pr create` from the already-built branch. This is the same class of surprise for any workflow that has legitimate work *after* the merge.

## Proposed design change

The refinery should **not** unilaterally close the work item / end the polecat's lifecycle. Ownership of completion should sit with the agent doing the work (and/or the mayor that dispatched it):

- The refinery reports the merge result (as it already does via mail) but leaves the item in a state the polecat/mayor closes **after** all post-merge steps complete; **or**
- the polecat calls `mg done` itself once its full flow (including PR creation) is finished, and the refinery never calls `mg done` on the author's behalf.

Either way, a merge succeeding should not race the polecat to `done` and kill it mid-flow.

## Notes

- Filed at dmiller's request after this bit the payitgov defaults PR twice.
- Reproducible with any `--branch` polecat whose flow opens a PR after the refinery merge.

drellem2:
Thanks — looking into this.

drellem2:
Confirmed. For `--branch` (PR-flow) polecats the refinery merges and marks the work item done itself, and the item going done is what stops the polecat — so it exits before running its own post-merge steps (`gh pr create`, verify, mailing the PR URL). That's why the branch landed but the PR was never opened.

Plan: an opt-in `--defer-done` submit flag that skips the refinery's auto-done/auto-stop, letting a polecat with post-merge work own its own lifecycle and call `mg done` once its full flow — including PR creation — finishes. That ships with a bounded backstop that reaps and escalates a deferred polecat which never completes, so we don't trade a silent kill for a silent linger.

PR #85, opened 2026-07-15
build: refinery target-branch reset + retryable ff-abort (drellem2/pogo#80)

### LINKED ISSUE #80: refinery: reused worktree target branch never reset — failed post-merge push corrupts state → misleading 'Not possible to fast-forward' on later attempts/MRs
## Describe the Bug

The refinery reuses a single persistent clone per repo (`ensureWorktree`, `merge.go:337`). Within a merge attempt the **source branch** is hard-reset each time (`git checkout -B <branch> origin/<branch>`, `merge.go:147`), but the **target branch is not** — it is a plain `git checkout <target>` (`merge.go:198`) followed by `git pull --ff-only origin <target>` (`merge.go:204`), with no `-B`/`reset --hard` to realign it with origin.

As a result, when an attempt succeeds at the local `git merge --ff-only <branch>` (`merge.go:210`) but then **fails at the push** `git push origin <target>` (`merge.go:216`), the worktree's local target branch is left **ahead of origin** and is never rolled back. That dirty state persists into the next attempt *and into the next MR that reuses the same clone*.

On the next cycle, `git pull --ff-only origin <target>` (`merge.go:204`) sees local-ahead-of-origin and aborts with:

```
fatal: Not possible to fast-forward, aborting.
```

This is (a) **misleading** — it points at the branch/fast-forward when the real cause was the earlier failed push, and (b) returned **non-retryable** (`merge.go:205`, not wrapped in `retryableError`), so the MR dies with a symptom that has nothing to do with the branch under review.

The confusing signature in the field: an operator inspecting a **fresh clone** confirms the fast-forward is trivially possible (origin/target + branch), yet the refinery keeps reporting it impossible — because the refinery's *own reused worktree* has a polluted local target, invisible from a clean clone.

This is distinct from the already-closed #68 / #69 (stock refinery not supporting direct-push to protected/PR-required branches — understood and handled out-of-band). This issue is the **failure-mode hygiene**: *any* post-`merge --ff-only` push failure (protected branch, transient rejection, lost race, the auto-version-bump scenario in #13) corrupts the reused worktree and mis-diagnoses every subsequent MR for that repo until the clone is manually cleaned.

## To Reproduce

1. Point the refinery at a repo whose default branch rejects the merge push at least once (e.g. a branch-protection / org ruleset that requires PRs or signed commits — the deterministic case — or any transient push rejection).
2. Submit an MR. Attempt N reaches `merge --ff-only` (succeeds locally) then `push origin <target>` (rejected). Local target is now ahead of origin.
3. Observe the next attempt / next MR against that repo fail at `pull --ff-only origin <target>` with `fatal: Not possible to fast-forward, aborting`.
4. Independently verify from a fresh clone that the branch *is* a trivial fast-forward of origin/<target> — it is. The refinery's worktree is the only thing that disagrees.

## Expected Behavior

Each attempt should start from a target branch that exactly matches `origin/<target>`, so a prior partial/failed merge can never poison later attempts or later MRs. The final surfaced error should reflect the real failing step (the push), not a downstream ff artifact.

## Actual Behavior

The local target branch retains the unpushed merge across attempts and across MRs (persistent clone), producing a misleading, non-retryable `Not possible to fast-forward` that masks the true push failure.

## Suggested Fix

- Reset the target to origin at the start of the merge/push step, mirroring the source-branch handling: `git checkout -B <target> origin/<target>` (or `git reset --hard origin/<target>` after checkout) instead of the plain `checkout` + `pull --ff-only` at `merge.go:198-204`. This makes each attempt hermetic w.r.t. the target.
- Classify a deterministic protected-branch / rule-violation push rejection as **non-retryable** with an actionable message (as `isAuthFailure` already special-cases auth at `merge.go:219`), so it fails fast on the *real* cause instead of burning `max_attempts` and then mis-reporting an ff error.

## Environment

- Pogo version: `main` @ `884ecaa`
- Observed against a `payitgov/*` target repo (org ruleset requires PR + verified signatures on the default branch).
- Surfaced by the pogo-slack-bridge deployment while investigating internal tickets ca-7717 / ca-3de2.

## Additional Context

Related: closed #68 (push fails on signed-commit branch protection), closed #69 (PR-only mode request), closed #13 (ff-only failures on auto-version-bump CI — same push/retry surface), closed #30 (advances origin/<target> but not the local *source* checkout — sibling state-sync gap, different checkout). This issue is specifically about the **refinery's own reused worktree target** not being reset.

drellem2:
## Summary

Fixes the refinery persistent-clone reuse bug: a failed push to the target branch left the reused clone poisoned and wedged later merge requests.

`ensureWorktree` keeps one persistent clone per repo. The **source** branch was hard-reset each attempt (`git checkout -B <branch> origin/<branch>`), but the **target** was not — plain `git checkout <target>` + `git pull --ff-only origin <target>`. When a local `git merge --ff-only` landed but `git push origin <target>` **failed** (protected branch, transient remote error), the local target was left ahead of origin and never rolled back. The next attempt *and* the next MR reusing the clone then aborted on the ff-only pull with `fatal: Not possible to fast-forward` — misleading (the real cause was the earlier failed push) and returned **non-retryable**, wedging later MRs.

## Fix

Realign the target with origin at the start of the merge phase, mirroring the source-branch reset:

- `git fetch origin <target>` (origin may have advanced during the gate phase)
- `git checkout -B <target> origin/<target>` — forces the local target to the fetched origin tip regardless of prior local state, so a poisoned/ahead target self-heals instead of aborting.
- The fetch/reset errors are classified **retryable**, so a transient hiccup gets another cycle rather than wedging.

This eliminates the non-retryable ff-only abort path (satisfying option 1 *and* the intent of option 2 in the signed-off fix: a failed push no longer poisons later MRs, because the next attempt resets the target).

Distinct from **mg-58f6**, which applied the same origin-reset on the `prune` path (`prune.go`); this is the merge path (`merge.go`). No shared lines with #81.

## Test

`TestFailedTargetPushDoesNotPoisonReusedClone` drives the real path: a pre-receive hook on origin rejects MR #1's push to `main` after its local ff-merge lands, poisoning the reused clone; the hook is removed and MR #2 must merge cleanly through the same clone. Verified the test **fails against the old code** with the exact `Not possible to fast-forward` symptom and **passes with the reset**.

`./build.sh` (fmt + full test suite + build) passes.

Resolves drellem2/pogo#80

Approved triage recommendation: inline signed-off fix in mg-f1db (Daniel GO recorded, pm-pogo relay 2026-07-15). Review ticket: mg-04e4.
Work item: mg-f1db

### DIFF
--- a/internal/refinery/integration_test.go
+++ b/internal/refinery/integration_test.go
@@ -250,6 +250,127 @@ func TestResubmitAlreadyMergedBranch(t *testing.T) {
 	}
 }
 
+// TestFailedTargetPushDoesNotPoisonReusedClone reproduces the persistent-clone
+// reuse bug from gh #80 / mg-f1db. ensureWorktree keeps ONE clone per repo. If
+// a cycle's local ff-merge to the target lands but the subsequent
+// `git push origin <target>` fails (protected branch, transient remote error),
+// the clone's local target is left AHEAD of origin and never rolled back. Under
+// the old code the NEXT MR reusing that clone did `checkout <target>` +
+// `pull --ff-only`, which aborts "Not possible to fast-forward" and was returned
+// non-retryable — wedging every later MR through that clone.
+//
+// The fix hard-resets the target to origin/<target> at the start of the merge
+// phase, so a poisoned/ahead target self-heals. This test drives the real path:
+// a pre-receive hook on origin rejects the first MR's push to main (after its
+// local merge already landed), leaving the reused clone poisoned; the hook is
+// then removed and a second MR must merge cleanly through the same clone.
+func TestFailedTargetPushDoesNotPoisonReusedClone(t *testing.T) {
+	if _, err := exec.LookPath("git"); err != nil {
+		t.Skip("git not found")
+	}
+
+	// === origin bare repo + working clone with a passing gate on main ===
+	originDir := t.TempDir()
+	run(t, originDir, "git", "init", "--bare", "-b", "main")
+
+	workDir := t.TempDir()
+	run(t, workDir, "git", "clone", originDir, ".")
+	run(t, workDir, "git", "config", "user.email", "test@test.com")
+	run(t, workDir, "git", "config", "user.name", "Test")
+	os.WriteFile(filepath.Join(workDir, "build.sh"), []byte("#!/bin/sh\nexit 0\n"), 0755)
+	os.WriteFile(filepath.Join(workDir, "main.go"), []byte("package main\nfunc main() {}\n"), 0644)
+	run(t, workDir, "git", "add", ".")
+	run(t, workDir, "git", "commit", "-m", "initial commit with build gate")
+	run(t, workDir, "git", "push", "origin", "main")
+
+	// feature-1 forks main — this MR's push will be rejected below.
+	run(t, workDir, "git", "checkout", "-b", "feature-1")
+	os.WriteFile(filepath.Join(workDir, "feature1.txt"), []byte("feature 1"), 0644)
+	run(t, workDir, "git", "add", ".")
+	run(t, workDir, "git", "commit", "-m", "feat: feature 1")
+	run(t, workDir, "git", "push", "origin", "feature-1")
+
+	// feature-2 forks main independently — the second MR, which must recover.
+	run(t, workDir, "git", "checkout", "main")
+	run(t, workDir, "git", "checkout", "-b", "feature-2")
+	os.WriteFile(filepath.Join(workDir, "feature2.txt"), []byte("feature 2"), 0644)
+	run(t, workDir, "git", "add", ".")
+	run(t, workDir, "git", "commit", "-m", "feat: feature 2")
+	run(t, workDir, "git", "push", "origin", "feature-2")
+
+	wtDir := t.TempDir()
+	r, err := New(Config{
+		Enabled:      true,
+		PollInterval: time.Hour,
+		WorktreeDir:  wtDir,
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Install a pre-receive hook on origin that rejects every push. The
+	// refinery's local ff-merge of feature-1 still succeeds, so the reused
+	// clone is left with local main ahead of origin — the poison state.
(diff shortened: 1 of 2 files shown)


drellem2:
## Review round 1: PASS
Reviewer: mg-04e4 · build ticket: mg-f1db · blocking: 0 · advisory: 1

Reviewed against the signed-off fix (mg-f1db / mg-1140): target realignment + retryable classification of the ff-only abort, plus regression coverage.

**QA** — `./build.sh` (fmt + full suite + build) passes on the PR head; CI checks all green. `TestFailedTargetPushDoesNotPoisonReusedClone` passes on the fix and, with `merge.go` reverted to `main`, fails with the exact `fatal: Not possible to fast-forward, aborting.` symptom — it genuinely reproduces the failed-push-then-reuse poison and guards it.

**Architecture** — `git checkout -B <target> origin/<target>` (`merge.go:220`) mirrors the source-branch reset at `merge.go:147` and is consistent with the prune-path origin reset (mg-58f6). The `retryableError` wrapping is functional (`isRetryable`/`errors.As`, `merge.go:110`). Logging follows the existing `step=` convention. No new abstractions or dependencies.

**Design-faithfulness** — both items of the signed-off fix delivered: (1) target realigned to origin at the start of each attempt; (2) `fetch-target`/`reset-target` classified retryable, eliminating the non-retryable ff-only abort path. Diff is scoped to `merge.go` + one integration test; no scope creep, no silent omissions. Distinct from the prune path (#/mg-58f6) and #81.

### Advisory (non-blocking)
- `internal/refinery/merge.go:205` (comment) / PR description — the narrative states the old abort surfaced at `pull --ff-only` and was returned non-retryable. In the reproduced scenario the abort actually surfaces one step later at `merge --ff-only <branch>` (`merge.go:227`), which was already retryable and simply exhausts attempts; the non-retryable `pull --ff-only` abort is the *origin-advanced-during-poison* manifestation. Both stem from the same root poison and the reset fixes all of them — the fix is correct; only the description is slightly narrower than the empirical failure. No code change required.

PR #90, opened 2026-07-17
Delete the refinery's OnSubmit unlink hook — it strands polecats on every merge failure

### LINKED ISSUE #88: refinery: submit-time UnlinkWorktree strands failing polecats; no local-only merge mode for push-denied clones (+ local-main heads-up)
Non-urgent, no SLA — from the agent-crew reliability work on this host. Three related refinery items: one bug, one feature ask, one heads-up about our local clone. Self-contained.

---

## 1. Bug: refinery unlinks a polecat's git worktree at SUBMIT time, stranding it when the merge later fails

**Context.** pogod's refinery runs polecat (agent) jobs in per-job git worktrees under `/opt/pogo/.pogo/polecats/<id>`. A polecat does its work in the worktree, then runs `pogo refinery submit`; the refinery attempts to merge the polecat branch into `main`.

**Problem.** `refinery.Submit()` fires the `OnSubmit` hook at **enqueue** time (`StatusQueued`), *before any merge attempt* (`internal/refinery/refinery.go:420-426`). pogod wires `OnSubmit = unlinkSubmittedPolecatWorktree` (`cmd/pogod/main.go:1144-1146`; body `cmd/pogod/reap.go:230-243`), which calls `refinery.UnlinkWorktree(SourceRepo, WorktreeDir)` (`internal/refinery/merge.go:748-767`). `UnlinkWorktree` does `os.Remove(worktreeDir/.git)` + `git -C <sourceRepo> worktree prune` — i.e. it severs the worktree's git linkage and prunes the registration, leaving the directory as a plain non-git folder while the polecat is still running.

**Failure mode.** This is fine when the merge succeeds. But when the merge **fails**, the worktree is now a non-git directory: the polecat cannot commit/push a fix, and for a repo whose polecat branches have no reachable remote, the refinery's re-fetch path can't recover the branch either. Concretely, on this repo (where merges were failing for the push reason in item 2 below), every affected polecat was stranded post-submit — e.g. a merge attempt logged `checkout branch: fatal: origin/polecat-<id> is not a valid ...` because the branch linkage was already gone.

**Fingerprint / evidence.** On-disk, stranded polecat dirs exist but have no `.git`, and `git -C <sourceRepo> worktree list` no longer lists them (directory intact + `.git` gone + registration pruned) — distinct from a full `RemoveWorktree`/GC sweep (which would `RemoveAll` the directory). Timeline matched submit time, not any GC tick.

**Suggested fix (pick one).**
1. Defer the unlink from `OnSubmit` to `OnMerged` / `OnFailed-final`, so the worktree stays usable until the job actually concludes; **or**
2. Gate `UnlinkWorktree` on the same live-polecat / still-claimed check the GC sweep already uses (`livePolecatSet`) — don't unlink while the owner is alive; **or**
3. On merge failure, re-link the worktree (restore the `.git` pointer + `git worktree repair`) so fix-and-resubmit works.

---

## 2. Feature ask: a local-only merge mode for repos where the clone can't push to origin

**Context.** The refinery's merge cycle hardcodes a final `git push origin <target>` (`internal/refinery/merge.go:169-185`) and refuses repos without a push remote (`merge.go:304`). On this host, the clone of *this* repo intentionally has **read-only** access to GitHub — the local clone's `main` is the operative deployment target, and upstream is yours to pull, not ours to push.

**Consequence.** The refinery can never merge changes to this repo: gates run green, the merge lands in the worktree, then the push fails auth and everything rolls back — local `main` never advances. Combined with item 1, a failed push also strands the polecat.

**Ask.** A per-repo opt-in (e.g. in `.pogo/refinery.toml`: `merge.push = false` / `local_only = true`) that ends the merge cycle at the local target-ref fast-forward, skipping the origin push (and skipping the origin-existence check for the target ref). Everything else — gates, rebase, ff-merge, retry semantics — unchanged.

**Meanwhile** we use a manual fallback for this repo: temp worktree from local `main` → merge → `build.sh` + `test.sh` full green → ff local `main`, evidence recorded. It works but bypasses the refinery's queueing and bookkeeping.

---

## 3. Heads-up: local main on this host advanced past the running runtime

Via the manual fallback above we landed two small fixes on the local clone's `main` (`/opt/pogo/repos/pogo`):

- `57d0a4b` → `c2dba9b`: make stall/wedge monitoring next-scheduled-wake-aware (fixes false stall classification for on-demand agents)
- `c2dba9b` → `d8e47e2` (current local `main`): harden `build.sh`/`fmt.sh` to self-locate `go` (fixes the refinery gate failing on PATH)

Both passed `build.sh` + `test.sh` in full before landing; fast-forward only, no history rewrite. Note the ff also advanced local `main` through ~169 intermediate upstream commits that the running pogod binary (built from `57d0a4b`) predates — we are deliberately **not** rebuilding pogod from local `main` until you've had a chance to look. Happy to send the two fix commits as `git format-patch` output, a PR from a fork, or leave them for you to cherry-pick from this host — whatever is easiest. No SLA on any of this.

drellem2:
## The bug

The refinery fires `OnSubmit` at **enqueue** (`StatusQueued`), **before any merge is attempted**, and pogod wired it to unlink the submitting polecat's worktree — `os.Remove(worktree/.git)` + `git worktree prune`.

Harmless when the merge succeeds. **On merge failure the polecat is left sitting in a directory that is no longer a git repository, and cannot commit or push a fix.**

The sharpest statement of it: **pogod contradicted itself in one function.** It deleted the polecat's `.git` at submit, then on merge failure **mailed that same polecat "so they can fix and resubmit"** and **reopened the work item "assigned to the original polecat."** The retry loop was structurally guaranteed to fail — on **every** merge failure, not just the reporter's.

## Why delete, rather than defer / gate / repair

The archeology is the argument. The unlink was **prong 2** of the gh#4 fix (`37c1359`). **Prong 1 of that same commit** — `git clone --no-local` plus pointing the refinery's clone at the real remote — **removed the reason prong 2 existed.** The refinery merges in its **own independent clone**, so the source repo's checked-out branches are invisible to it.

The reporter's option (b), gating on `livePolecatSet`, was rejected — and it's instructive *why* it's tempting: **at submit time the polecat is always live, so the gate would suppress the unlink 100% of the time.** It is *"delete the hook" wearing a conditional*, with a predicate that misleads the next reader into thinking a live case exists. `livePolecatSet` is a GC-sweep exclusion for **concluded** polecats, not a submit-time guard.

## ⚠️ Reviewer: do NOT delete `TestBranchCheckedOutInWorktree`

**`TestBranchCheckedOutInWorktree` (`internal/refinery/integration_test.go`) changes jobs in this PR.** It merges a live, checked-out polecat branch with **no unlink**, and it passes on main today. It is now **the load-bearing regression guard proving this deletion is safe.**

It looks like one of the hook's tests. It is not. Deleting it alongside them would silently remove the only proof the deletion is safe. It is deliberately left untouched here.

## Proofs

**RED — via production code, no stubs.** Drove the real `unlinkSubmittedPolecatWorktree` + real `refinery.UnlinkWorktree` against a real git worktree:

```
STRANDED — polecat cannot stage its fix:  fatal: not a git repository (or any of the parent directories): .git
STRANDED — polecat cannot commit its fix: fatal: not a git repository (or any of the parent directories): .git
STRANDED — polecat cannot push its fix:   fatal: not a git repository (or any of the parent directories): .git
```

The reporter's fingerprint, exactly. Worth noting: **the hook's own three tests never caught this, because all three injected a _stub_ unlink** — they asserted the hook resolves the right agent, never that the real unlink strands it.

**The guard can go RED.** Broke the ff-merge step and `TestBranchCheckedOutInWorktree` failed with `expected merged, got failed` — it genuinely asserts an end-to-end merge against a live checked-out worktree, and is not hard-wired to pass. Reverted.

**The negative control — a _successful_ merge still cleans up.** `gitgc.RemoveWorktree` (the real cleanup, on every polecat exit) now drops a **still-linked** registration, which is what frees the branch for deletion. Added `TestRemoveWorktreeFreesCheckedOutBranch`, and confirmed it goes RED (`cannot delete branch 'polecat-live' used by worktree`) when registration removal is skipped. So the pass is **conditional**, not assumed.

The residual risk was checked, not assumed: with no submit-time unlink the branch stays checked out until exit, so a bare `git branch -D` would refuse. Already handled — `gitgc.Sweep` processes worktrees **before** branches precisely so removing a worktree frees its branch in the same pass, and live polecats are excluded.

## What changed

- `cmd/pogod/main.go` — deleted **both** wiring sites: the `SetOnSubmit` closure, and the `OnSubmitFunc`/`SetOnSubmit` re-wire on the orchestration-restart path (which existed only to keep unlinking across a refinery rebuild).
- `cmd/pogod/reap.go` — deleted `unlinkSubmittedPolecatWorktree` + `worktreeUnlinker`.
- `internal/refinery/merge.go` — deleted `UnlinkWorktree`. Re-derived its callers independently: **exactly one non-test caller** (the hook), so it is dead code. Now zero.
- `internal/refinery/refinery.go` — **kept** the generic `OnSubmit` hook point per the approved recommendation, and documented its contract: a callback fires on an MR whose outcome is not yet known, so it must not take an action a failed merge would need undone.
- Tests deleted: the three hook tests, and `TestUnlinkWorktree`. `TestOnSubmitCallback` (the generic hook point) is kept.
- Comments/docs my change falsified: `gitgc.RemoveWorktree` no longer claims `git worktree remove` "fails harmlessly because the pointer is already unlinked"; `sweep.go` notes new orphans stop accruing; `docs/customizing.md` and `docs/refinery-persistence-design.md` corrected.

**Bonus:** this stops **new gh#31 orphan dirs** accruing — `sweep.go` names submit-time unlink as their cause; it was what stripped a live polecat's registration in the first place. The legacy orphan-dir scan is kept for the dirs already out there.

Verified with `./build.sh` (full gate, incl. `scripts/pogo-self-deploy_live_test.sh` — 13/13, untouched). Net −111 lines.

Out of scope, noted not fixed: items 2 and 3 of the issue are tracked separately.

Resolves drellem2/pogo#88 (item 1)

Approved triage recommendation: mg-c155 (`~/.macguffin/work/done/mg-c155.result.json`) — Daniel APPROVED 2026-07-17 09:58Z
Work item: mg-143e


drellem2:
## Review round 1: PASS
Reviewer: mg-f6ec · build ticket: mg-143e · blocking: 0 · advisory: 2

Every claim in the PR description was re-derived against the tree rather than accepted. All held.

### Verified

- **Both wiring sites removed.** Zero `SetOnSubmit`/`OnSubmitFunc`/`onSubmit` references remain in `cmd/pogod` — the `SetOnSubmit` closure and the `OnSubmitFunc` re-wire on the orchestration-restart path are both gone, so no path refers to a hook that no longer exists. The generic hook point survives in `refinery.go` with `TestOnSubmitCallback` still exercising it, per the approved recommendation.
- **`refinery.UnlinkWorktree` callers re-derived independently, not inherited.** On `main`: exactly one non-test caller (`cmd/pogod/main.go:1378`, the hook), plus its definition and `TestUnlinkWorktree`. The compiler corroborates what grep claims — the tree builds clean with the function deleted, which sees references grep cannot. No reflection or string-based dispatch. `internal/refinery` is not importable outside the module, so deleting an exported symbol breaks no external API.
- **`TestBranchCheckedOutInWorktree` survives byte-identical** (`md5` of the function body matches `main`), is untouched by the diff, and is named as the load-bearing guard in the PR description.
- **The guard is not hard-wired to pass.** Broke the ff-merge target: `--- FAIL: TestBranchCheckedOutInWorktree` / `expected merged, got failed`. Reverted; tree clean.
- **The negative control can go RED.** Skipped the registration drop in `RemoveWorktree`: `--- FAIL: TestRemoveWorktreeFreesCheckedOutBranch` / `cannot delete branch 'polecat-live' used by worktree`. Reverted; tree clean.
- **The bug was real, reproduced independently.** Drove `main`'s production `refinery.UnlinkWorktree` against a real git worktree in an isolated tree. Added a pre-unlink control the PR did not have: the polecat **can** stage a fix before the hook fires, and after it `git add`/`commit`/`push` all return `fatal: not a git repository`. The stranding is caused by the hook, not by the fixture.
- **QA.** `./build.sh` exits 0. Re-ran `internal/refinery`, `internal/gitgc`, `cmd/pogod` with `-count=1` to defeat cached results — all pass. `gofmt` clean, `go vet` clean. `scripts/` is untouched by the diff (0 files), so the self-deploy live assertions are unweakened by inspection, not by assertion count.
- **Faithfulness.** Walked the diff against the approved recommendation (mg-c155) item by item: all six deliverables present, no silent omissions. The `internal/gitgc` additions are not scope creep — the negative control was required by the ticket's bar, and the comment corrections are prose the deletion falsified.

### Advisory (non-blocking)

- `cmd/pogod/gitgc.go:61` — comment still names submit-time unlink as the live cause of orphan polecat dirs ("worktree unlinked at submit time"). Same class of falsified prose the PR deliberately corrected in `sweep.go`; this instance was missed.
- `cmd/pogo/main.go:2099` — user-facing `pogo gc` help text carries the same claim ("when teardown never ran after the worktree was unlinked"). Still true of the legacy dirs the scan exists for, so it is imprecise rather than wrong.

Neither affects behavior. Not worth holding the branch; worth a line if the file is touched again.

PR #92, opened 2026-07-27
fix(refinery): an integration-branch merge is a PR-flow step, not completion (mg-7746)

drellem2:
## The bug

A polecat dispatched with `--branch` merges into an **integration branch**, and
its deliverable is the pull request from that branch to the repo default — not
the merge. pogod nevertheless ran `mg done` and stopped the polecat the instant
the merge landed, so the polecat exited before `gh pr create`:

**work merged → item marked done → no PR exists → manual recovery.**

Three items hit this (mg-74ee, mg-6579, mg-7746). The polecat prompt was
correct every time — mg-6579's rendered prompt (30906 bytes) contained the
"do NOT call `mg done` here, and do NOT exit" guard and four `gh pr create`
occurrences. The recorded result was `{"branch":"polecat-6579",
"completed_by":"refinery","mr":"mr-d9jn4gc7hqpl76v2q8b0"}` — the refinery wrote
it, not the polecat. `gh pr list --head daed-101-new-action-item --state all`
returned `[]`.

`--defer-done` (gh #81) already existed as the opt-in escape hatch. Opt-in was
the defect: nobody passed it.

## The fix

`Submit` derives the answer instead of waiting to be told. It resolves the
repo's default branch and sets `pr_flow` on the merge request when `target_ref`
is anything else. A `pr_flow` merge is then handled exactly like
`--defer-done`: merge, mail, leave the work item **claimed**, keep the polecat
running, and let it call `mg done` itself once the PR is open.

Merges onto the default branch are unchanged — that `mg done` is correct
completion.

A submitter cannot opt out either: `PRFlow` is overwritten on every `Submit`,
so a caller cannot force a default-branch merge to defer forever, nor hide a
genuine integration merge.

### Second-order fixes shipping with it

- **The result sidecar records `target`** (and `pr_flow` when true). Mayor's
  classification logic reads those fields as its secondary signal for telling
  "integration merge, PR pending" from "merged to main, done"; omitting them
  made a documented signal *actively misleading*, not merely absent.
- **The defer-done backstop checks whether the work item actually completed
  before escalating.** Widening deferral to every PR-flow merge exposed a
  latent problem: a polecat that opens its PR and calls `mg done` is told by
  its own protocol to *stay alive* until the mayor stops it, so at the 15-minute
  deadline the process is still there. It is still reaped (slot protection is
  the point) but no longer reported as one that "never called `mg done`".
- **Default-branch detection asks the remote** (`git ls-remote --symref origin
  HEAD`) when `refs/remotes/origin/HEAD` is missing locally, before falling back
  to local `HEAD`. In a working clone local `HEAD` is just whichever branch is
  checked out, so a clone parked on a feature branch would have reported that
  branch as the repo default and made every merge to `main` look like an
  integration step.

## Verification

The ticket asked for a reproduction first, because this bug survived a template
that *looked* right and a green unit suite. `cmd/pogod/prflow_test.go` drives
the real thing end to end — real git repos, a real refinery loop, real
default-branch detection, the real `reapMergedPolecat` wiring from `main.go`.

**Before the fix**, against an integration-branch target:

```
prflow_test.go:173: MR merged into integration branch "daed-101-integration"
                    (default is "main") but PRFlow=false
prflow_test.go:177: refinery called mg done for "mg-7746" ...
                    result="{\"branch\":\"polecat-mg-7746\",\"completed_by\":\"refinery\",\"mr\":\"...\"}"
prflow_test.go:181: refinery stopped polecat [7746] before it could open the PR
```

That sidecar is byte-for-byte the shape recorded on mg-6579.

**After the fix**, all three directions pass:

| Scenario | Result |
|---|---|
| `--target=<integration-branch>` merge | item stays **claimed**, polecat keeps running, backstop armed |
| default-branch merge | still `mg done` + stop (fast path preserved) |
| default-branch result sidecar | carries `target`, omits `pr_flow` |

Third check, run live against the real `mg` CLI on a scratch `MG_ROOT`: an item
left `claimed` accepts the polecat's own step-13 `mg done`, and the richer
result is what lands:

```
status after claim: claimed
status after done:  done
{"branch":"polecat-mg-7746","note":"PR opened","target":"daed-101-integration",
 "pr_flow":true,"pr":"https://github.com/x/y/pull/1"}
```

New unit coverage: `PRFlow` derivation and its state-file round-trip (a pogod
restart between submit and merge must not resurrect the bug), submitter-claim
rejection, auto-created targets, remote-over-checkout default detection, and
both directions of the backstop's completion probe.

`./build.sh` is green (`exit=0`). Two `internal/ghteardown` failures on my host
were environmental — macOS `/usr/bin/mg` is the micro-emacs editor, not
macguffin; they pass with the real `mg` first on `PATH`, and are untouched by
this change.

## Deployment note

**This changes pogod. The running daemon is stale until it is restarted, and I
have not restarted it** — mayor and other agents are live under it. The fix
takes effect only after whoever merges this restarts pogod.

## Scope

`~/.pogo/agents/templates/polecat.md` is deliberately **not** touched. The
template is correct; it was never the defect. mg-b01a covers a genuinely
different template bug (the checklist being stripped when `--branch` is absent)
and must not be conflated with this one.

Refs drellem2/pogo#81


drellem2:
## Review round 1: PASS

Reviewer: mg-f700 · build ticket: mg-7746 · blocking: 0 · advisory: 5

### Verification

**The bug is real, and reproduces.** `internal/agent/prompts/templates/polecat.md:104` and `polecat-architect.md:172` submit with `--target={{if .Branch}}{{.Branch}}{{else}}main{{end}}` and never pass `--defer-done`; on `main`, `Submit` sets no deferral and `cmd/pogod/reap.go:74` gates solely on `mr.DeferDone`, so an integration-branch merge falls through to `complete()` + `Stop()`. I confirmed this empirically rather than by reading: a scratch worktree at `main` (`00acb6a`), driven through a real refinery loop over real git repos with `TargetRef="daed-101-integration"`, produced

```
completedID="mg-7746"
result={"branch":"polecat-mg-7746","completed_by":"refinery","mr":"mr-d9ke1gatjv1jk12l47ug"}
stopped=[7746]
```

— the same sidecar shape as the mg-6579 evidence cited in the description.

**The fix is correctly scoped.** Deferral is derived in `Submit` from the resolved default branch (`internal/refinery/refinery.go:426-436`) and consumed at `cmd/pogod/reap.go:80`. The default-branch path is unchanged and has an explicit positive control (`TestDefaultBranchMerge_StillMarksDone`): `mg done` fires, the polecat is stopped, no backstop is armed. `TestSubmitPRFlowIgnoresSubmitterClaim` pins that the classification cannot be forged or suppressed by the submitter, and `TestSubmitDerivesPRFlowFromTarget` pins that it survives a pogod restart via the state file.

**No hang on `gh pr create` failure.** The bounded backstop is armed for every PR-flow merge (`reap.go:86`), reaps at 15 min, and escalates when `MGWorkItemDone` reports the item is not terminal. `mg show <id> --json` returns `status` for archived items too, so the probe is correct on both terminal states.

**Gates.** `./build.sh` passes on the PR head (exit 0). The new and existing pogod suites pass.

### Advisory (non-blocking)

- `ARCHITECTURE.md:290`, `changelog.d/mg-7746.fixed.md:20` — "The result sidecar the refinery writes records `target` (and `pr_flow` when true)". `pr_flow` is never written: the only sidecar writer is `cmd/pogod/reap.go:103`, which the PR-flow path returns before reaching (`reap.go:80-89`), and the map has no `pr_flow` key. A PR-flow merge produces no refinery sidecar at all — that absence is the real signal. Same for the comment at `reap.go:100-102`, which reads as if it describes this map rather than the `MergeRequest` JSON. Worth correcting the two sentences; a documented-but-absent field is the failure mode this change exists to remove.
- `cmd/pogod/reap.go:204` / `cmd/pogod/main.go:1269` — `cancel()` disarms unconditionally on process exit and logs "completed cleanly". A PR-flow polecat that dies after the merge but before `mg done` therefore gets no escalation, and no claim release: `releasePolecatClaim` runs only from `Registry.Stop` (`internal/agent/agent.go:1074`, `:1120`), not from the exit hook. The item is left in `claimed/` under a dead pid — the mg-fb13 class. Never worse than the pre-fix outcome, and the window is short, but deferral is now the default for every `--branch` merge rather than opt-in dead code, so the path now carries traffic. `cancel()` could consult `workItemDone` the way `fire()` does.
- `internal/refinery/refinery.go:565` — `remoteDefaultBranch` reduces but does not close the misdetection risk. With `refs/remotes/origin/HEAD` absent, origin unreachable (mg-25fb: launchd pogod), and the clone parked on a feature branch, detection returns that branch, every merge to `main` classifies as PR flow, and nothing completes. `validateTargetRef` is lenient on an unreachable remote (`:513-520`), so submit does not fail first. The fallback chain is safe when detection *fails*, not when it is confidently wrong. The `refinery:` log line at `:434` names the detected default, which makes it diagnosable.
- `internal/refinery/refinery.go:566` — no timeout on the `ls-remote` exec, so an unresponsive remote blocks the submit handler. Matches the existing idiom at `:502`, so this is consistent rather than a regression; a shared `exec.CommandContext` deadline would cover both.
- `internal/agent/prompts/templates/polecat.md:109` — the step-6 note ("pogod stops you the moment the merge lands ... marks your work item done on your behalf") is now false for a `--branch` polecat, and no template instructs a post-merge `gh pr create`. PR creation still depends entirely on the dispatch body. The change creates the space for the PR; the prompt half is a follow-up.

Issue #94, opened 2026-07-29
pogod git GC deletes a LIVE polecat worktree (and its branch ref) when the polecat checks out a branch other than polecat-<name>


drellem2:
## Summary

pogod's hourly git GC removed a **running** polecat's worktree and deleted its branch ref mid-task, because the sweep decides a worktree is in use by inspecting its **checked-out branch name** rather than by asking whether a live agent owns the worktree **path**.

Observed on 2026-07-29 against v0.6.0's GC logic (`internal/gitgc/sweep.go`, unchanged on `main` at bd89c71). Work was recovered only because the commit was still loose in the shared object store — a `git gc --prune=now` would have made it unrecoverable.

## Evidence

From `~/.local/share/pogo/logs/pogo.err.log`:

```
2026/07/29 06:56:08 pogod: git GC enabled (interval 1h0m0s)
2026/07/29 14:10:59 polecat caa65: created worktree at /Users/.../.pogo/polecats/caa65
                    (branch polecat-caa65, base "origin/develop")
2026/07/29 14:15:53 pogod: git GC /Users/.../ui_config_service
                    — deleted 1 branches, removed 1 worktrees, 0 errors
```

The polecat independently reported, at the same time, that its worktree had vanished, that `refs/heads/polecat-dccb` was gone, and that `git worktree list` no longer showed it. The agent process was still running and `pogo agent list` still reported it healthy throughout — it got no signal at all; the directory simply disappeared underneath it.

## Root cause

`caa65` was dispatched to fix merge conflicts on an **existing** pull request whose head branch was `polecat-dccb`. Its worktree was created at `~/.pogo/polecats/caa65` on `polecat-caa65`, but to do the job it necessarily checked out `polecat-dccb`.

In `Sweep` phase 1 (`internal/gitgc/sweep.go`):

```go
name := BranchSuffix(wt.Branch)      // "dccb", NOT "caa65"
if opts.LivePolecats[name] { ...keep... }
_, state := tickets.BranchState(wt.Branch)   // mg-dccb: concluded
if !state.Concluded() { ...keep... }
freed[wt.Branch] = true
RemoveWorktreeForce(opts.Repo, wt.Path)      // deletes a LIVE agent's tree
```

The occupancy test is derived from the branch, so a polecat on any other branch resolves to a different polecat's name, misses the live set, and inherits *that* ticket's concluded state. Phase 2 then deletes the ref, because `freed[br]` waives the checked-out-branch guard.

[...]

PR #95, opened 2026-07-29
fix(gitgc): decide worktree in-use by live-agent path ownership, not branch name (mg-3b7c)

drellem2:
<h4>Description</h4>

pogod's hourly git GC removed a **running** polecat's worktree and deleted its branch ref mid-task, because `Sweep` decided a worktree was in use by taking its **checked-out branch**, stripping `polecat-`, and looking the remainder up in the live set. That silently assumed every polecat stays on its own `polecat-<name>` branch for its whole life. It cannot: a polecat dispatched to fix or rebase an **existing pull request** must check out that PR's head branch, because that is the only way to update a PR in place. The moment it did, its worktree resolved to the *other* polecat's name, missed the live set, inherited that other ticket's concluded state, and was removed together with the branch ref out from under an agent that `pogo agent list` still reported healthy. On 2026-07-29 that took out polecat `caa65`'s worktree five minutes after it was created; the work survived only because the commit was still loose in the shared object store, which a `--prune=now` would have reaped. This is a whole class of dispatch, not an edge case — every "fix the conflicts on PR #N" ticket produces it. Filed upstream as Refs drellem2/pogo#94.

In-use is now decided by **worktree PATH ownership**, which pogod knew all along — it records `Agent.WorktreeDir` at spawn and logs `polecat <name>: created worktree at <path>` — and simply never handed to the sweep. pogod passes it as the new `gitgc.Options.LiveWorktrees`, and gitgc additionally re-derives `<PolecatsDir>/<name>` for every live *name*, so a polecat that outlived the pogod which spawned it (registry empty after a restart, only the on-disk witness left, and the witness stores names and PIDs but not paths) is covered too. Paths compare through `filepath.EvalSymlinks`, since git canonicalizes the worktree paths it reports and `/var` vs `/private/var` would otherwise read as two different trees on macOS. A basename fallback covers the `pogo gc` CLI, where the live set arrives from pogod's HTTP agent list as bare names.

Three further guards ride along: a branch checked out in any **surviving** worktree is never deleted (`freed` used to be set *before* the removal was attempted, so a removal that **failed** still told the branch phase the ref was loose); a worktree holding another polecat's branch is kept even when nobody is live, because that ticket's conclusion is not evidence about *this* directory; and pogod now logs **which** worktree or branch it destroyed and why, rather than only `deleted N branches, removed N worktrees` — the count-only line is why diagnosing this took a log grep plus the affected polecat's own incident report. Prune stays conservative; nothing moves toward `--prune=now`.

**Caller-visible effects.** `gitgc.Options` gains one optional field (`LiveWorktrees`); every existing call site compiles and behaves as before. The GC is strictly *more* conservative — it keeps trees and refs it previously destroyed and never destroys anything it previously kept, pinned by `TestSweepStillReclaimsConcludedPolecatWorktree`. pogod's GC log lines are additive: the existing summary line is unchanged, with per-item detail lines after it. The fix takes effect on the next pogod restart; no restart is performed by this PR.

`internal/gitgc/foreignbranch_test.go` reproduces the incident and each guard. Against the pre-fix `Sweep` the live-worktree cases fail exactly as observed in production — the tree is removed and `remaining branches = []`.

<details open>
<summary><h4>Compliance Checklist</h4></summary>

- [x] I have verified that this is not a new project. If it is a new project (new service, new application, new repository), I have [submitted a ticket to InfoSec](https://payitdev.atlassian.net/servicedesk/customer/portal/15/group/154/create/526) to review the new project and gained approval.

- [x] I have verified that the backout plan for this change conforms to our standard engineering backout plan located [here](https://payitdev.atlassian.net/wiki/spaces/SEC/pages/2833416205/Standard+Change+Control+Back+Out+Plan). If it does not, I have documented an alternative backout plan below:

- [x] I have verified that this change is backwards compatible. If it is not, I have specified the breaking changes and how they will be handled below:

- [x] I have verified that this change will not impact the security controls built into the application or introduce any new security vulnerabilities. If it will, I have defined the security impact below:

- [x] I have verified that this change will not result in downtime. If it will, I have noted the impact below:

- [x] I have verified that no new dependencies were introduced. If they were, I have vetted them below:

- [x] I have verified that all applicable tests were updated to ensure complete test coverage of any new or modified code.

- [x] I have verified that any relevant documentation such as the README is still up to date and not impacted by my changes. If documentation needs updating for accuracy, I have done so.
</details>

---

**Test evidence.** `go test ./...` passes clean. `./build.sh` exits 0. `scripts/pogo-deploy_test.sh` reports 49 passed and 3 `sync_src` FAILs — these are **pre-existing**, verified by running the identical suite against a pristine `git archive origin/main` checkout, which produces the same three failures. They are in `scripts/`, which this PR does not touch.

### DIFF
--- a/changelog.d/mg-3b7c.fixed.md
+++ b/changelog.d/mg-3b7c.fixed.md
@@ -1,0 +1,41 @@
+- **git GC no longer deletes a LIVE polecat's worktree when it has checked out
+  a branch other than its own (mg-3b7c).** The sweep decided a worktree was in
+  use by taking its CHECKED-OUT BRANCH, stripping `polecat-`, and looking the
+  remainder up in the live set. That assumed every polecat stays on its own
+  `polecat-<name>` branch for its whole life — but a polecat dispatched to fix
+  or rebase an **existing pull request** must check out that PR's head branch,
+  because that is the only way to update a PR in place. The moment it did, its
+  worktree resolved to the *other* polecat's name, missed the live set,
+  inherited that other ticket's concluded state, and was removed **together
+  with the branch ref** out from under a running agent that pogod was still
+  reporting healthy. On 2026-07-29 that took out polecat `caa65`'s worktree
+  five minutes after it was created; the work survived only because the commit
+  was still loose in the shared object store, which a `--prune=now` would have
+  reaped. This is a whole class of ticket, not an edge case: every
+  fix-the-conflicts-on-PR-#N dispatch produces it, and the workflow cannot be
+  changed away without abandoning PR continuity.
+
+  In-use is now decided by **worktree PATH ownership**, which is what pogod
+  knew all along — it records `Agent.WorktreeDir` at spawn and logs it — and
+  never passed to the sweep. pogod now supplies it as
+  `gitgc.Options.LiveWorktrees`, and gitgc additionally re-derives
+  `<PolecatsDir>/<name>` for every live NAME, so a polecat that outlived the
+  pogod which spawned it (registry empty after a restart, only the on-disk
+  witness left, which stores names and not paths) is covered too. Paths
+  compare through symlinks, since git canonicalizes the worktree paths it
+  reports and `/var` vs `/private/var` would otherwise read as two trees.
+
+  Three further guards ride along:
+
+  - a branch checked out in any **surviving** worktree is never deleted —
+    `freed` used to be set *before* the removal was attempted, so a removal
+    that FAILED still told the branch phase the ref was loose;
+  - a worktree holding another polecat's branch is kept even when nobody is
+    live, because that ticket's conclusion is not evidence about this
+    directory;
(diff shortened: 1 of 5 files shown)


drellem2:
Closing — superseded and never merged.

The gh#94 fix that shipped is #96, in [v0.7.0](https://github.com/drellem2/pogo/releases/tag/v0.7.0) as `e067e90` and `e5b371c`. This branch's head (`a57bfd0`) is not an ancestor of `main`, so nothing here landed, and the symbols it introduces — `ownedByLive`, `polecatDirOwner` — do not exist on `main`.

Leaving it open has caused real confusion: a reporter scoped a follow-up against this PR and walked back a correct finding as a result, because reasoning from this tree predicts coverage the shipped fix does not have. Closing so the branch stops reading as the implementation.

Both approaches keyed liveness on path ownership rather than the branch name; #96 is the one that was reviewed and released.

PR #96, opened 2026-07-29
fix(gitgc): key worktree liveness off the path, not the branch (gh #94)

drellem2:
Keys the git GC's worktree-liveness gate off the worktree **path** instead of the branch checked out inside it, logs what the sweep removed and why, and preserves the triage's reproduction as a test with its control.

## The fix

`Sweep` phase 1 read liveness from `BranchSuffix(wt.Branch)` — the polecat name embedded in the branch checked out *inside* the worktree. The worktree's **path basename** is the polecat that owns the tree. Those two strings agree only while a polecat stays on its own branch, and our own shipped review and QA roles instruct a foreign checkout. When they disagree the live polecat was invisible to the gate: tree removed mid-task, `freed[wt.Branch]` waived phase 2's guard so the ref went too, and the agent kept running with `pogo agent list` reporting it healthy.

Liveness now uses `gitgc.PolecatNameForWorktree(wt.Path)`. Branch deletion (phase 2) keeps its branch-suffix key — that one is correct, and the two answer different questions:

| String | Answers | Sound key for |
|---|---|---|
| worktree path basename | whose tree this is | removing a **directory** |
| branch `polecat-` suffix | whose work this is | deleting a **ref** |

The branch is deliberately **not** retained as an `OR` fallback. An `OR` would re-admit the exact signal that made a live tree look dead, and would create the symmetric defect on the other side (a dead tree parked on a live polecat's branch, never reaped). Ticket-state classification stays branch-keyed — it is only ever reached for a tree whose owner is already established as not live, and re-keying it too would strand every worktree whose basename resolves to no work item.

The orphan-dir scan (phase 1b) was already path-keyed; phase 1 now matches it.

## Acceptance — the reproduction, inverted, against a live agent

Same fixture, same live daemon, only the binary differs. A **genuinely live polecat** (`dd92`, the one that wrote this PR) with a sandbox worktree at `polecats/dd92` checked out on the foreign, archived branch `polecat-beef`; and a control identical but for ownership, a dead polecat `cafe` on its own archived branch.

**Pre-fix binary, built from `main`:**

```
  worktrees: removed 2, kept 0
  branches:  deleted 2, kept 0
  worktrees removed:
    .../polecats/cafe (polecat-cafe)
    .../polecats/dd92 (polecat-beef)     <-- a LIVE agent's tree
--- surviving polecat dirs ---
--- surviving branches ---
```

**This branch:**

```
  worktrees: removed 1, kept 1
  branches:  deleted 1, kept 1
  worktrees removed:
    .../polecats/cafe (owner cafe, branch polecat-cafe): ticket archived
  branches deleted:
    polecat-cafe — ticket archived
--- surviving polecat dirs ---
  dd92
--- surviving branches ---
  polecat-beef
```

The live polecat keeps its tree **and** the branch it was working; the control still collects, tree and ref. A gate that keeps everything is indistinguishable from a disabled one, so the control is half the test — `TestSweepKeepsLivePolecatOnForeignBranch` carries both in a single sweep, and it goes red on the pre-fix key.

## Diagnosability

The sweep already assembled path, branch and reason for every decision and logged only counts, so the one removal sitting in a 5.5MB pogod log on the reporting host cannot be judged legitimate or not. `WorktreeAction` gains `Owner` and a `String()`; pogod wires `Options.Logf`:

```
pogod: git GC /Users/x/dev/pogo — removed worktree /Users/x/.pogo/polecats/caa65 (owner caa65, branch polecat-dccb): ticket archived
pogod: git GC /Users/x/dev/pogo — kept worktree /Users/x/.pogo/polecats/beef (owner beef, branch polecat-beef): 3 uncommitted change(s) — rerun with --force to discard
```

`owner` and `branch` print separately on purpose: when they differ you are looking at exactly this bug's situation. Only *actions* log — a keep-because-live line per polecat per tick would be noise.

`pogo gc`'s summary now carries the same detail **for keeps as well as removals**. It previously itemised removals and counted keeps, while pogod's log emitted the full line for both — so `pogo gc --help`'s promise that a dirty worktree is "KEPT and reported" was only half true, and a count told an operator a branch was pinned without saying which tree to rescue. (Round 1 advisory B.)

`TestRunGitGCSweepLogsTheAction` drives `runGitGCSweep` itself against a real throwaway repo and asserts the line pogod actually writes — **verified red** when `Logf: gitGCLogf(repo)` is deleted, with the whole `cmd/pogod` package failing. Reaching that line from a test needed a seam (`loadTicketIndexFn`), because the sweep's first act shells out to `mg` and returns early under a sandbox; that unreachability is exactly why the wiring had no coverage in round 1. `TestGitGCLogfFormat` covers the format helper and no longer claims to cover the wiring.

## The third part, labelled UNREPRODUCED

`freed[wt.Branch] = true` was set **before** removal was attempted and survived a failure, so phase 2 believed the branch was un-checked-out when it still was and attempted a deletion git was always going to refuse — reporting a spurious error in place of the correct `kept: checked out in a worktree`. This is **code-read only**, from the triage; it was never observed in the wild, and it does not inherit the confirmed half's standing. Setting `freed` after a successful removal fixes it. Tested through an injected removal failure (`removeWorktreeFn`) — a real one needs both `git worktree remove --force` and `os.RemoveAll` to lose to the filesystem, and a chmod fixture silently passes as root.

## Sibling: mg-0130 is fixed

*"livePolecatSet is registry-built, so a restart unprotects a done-but-still-running polecat's worktree"* — **fixed**, in commit `ec7fee1`, and independently of this ticket. `cmd/pogod/gitgc.go:livePolecatSet` unions the registry with `agent.WitnessedPolecatVerdicts()`, the `(pid, start_time)` store that survives a pogod restart, and a witness *read error* skips the sweep rather than sweeping against a live set known to be missing survivors. Guarded by `TestLivePolecatSet_WitnessGuardsDoneButRunningWorktreeAcrossRestart`, which asserts both directions. That test's fixture parked its worktree at `wt-<name>`; since the basename is now load-bearing it has been moved to the production-shaped `polecats/<name>`, and it still passes.

Two independent defects did land in one gate, which is an argument about its shape: both were the live set being keyed or populated by something other than "a running process owns this path".

## Out of scope, flagged not fixed

`pogo gc` (`cmd/pogo/main.go`, outside this ticket's scope) builds its live set from `client.ListAgents()` alone. `/agents` returns the registry plus parked crew — **no witness union** — so mg-0130's shape still applies to the manual CLI path after a pogod restart. Dry-run-by-default and human-driven, so it is not the same severity, but it is the same hole. Not fixed here; raising it separately.

## Notes

- **Not a v0.6.0 regression.** The branch-keyed check has been present since `gitgc` existed.
- `internal/agent/prompts/**` and `internal/agent/templates/**` untouched — the point is that the GC is correct without the prompt conventions that were incidentally protecting the review and QA roles.
- One file outside the declared scope (`internal/gitgc/**`, `cmd/pogod/**`, `docs/*.md`): `internal/agent/api_test.go`. Test-only, and it pins the new path coupling at the single site that writes the path, with the function that reads it. Flagged deliberately, not creep.
- **Follow-ups filed, not folded in:** [mg-1403] `pogo gc` builds its live set from `client.ListAgents()` alone and `/agents` has no witness union, so mg-0130's restart hole survives in the CLI path (`cmd/pogo` is outside this ticket's scope). [mg-bdda] phase 1 classifies ticket state by branch while phase 1b classifies by owner, so the same dead tree is kept or `rm -rf`'d depending on whether git still holds the registration — conservative direction, pre-existing, and now documented in `docs/polecat-worktrees.md`.
- Full gate green: `./build.sh` exits 0.

Refs drellem2/pogo#94

Approved triage recommendation: mg-ee98 (packet), public plan at drellem2/pogo#94 (comment 5119065619)
Work item: mg-dd92

### DIFF
--- a/changelog.d/mg-dd92.fixed.md
+++ b/changelog.d/mg-dd92.fixed.md
@@ -29,3 +29,10 @@
   always going to refuse — reporting an error in place of the correct "kept:
   checked out in a worktree". Found by code read during the gh #94 triage and
   never observed in the wild.
+
+- **`pogo gc` now itemises the worktrees it KEPT, not just the ones it removed
+  (mg-dd92).** `pogo gc --help` promises a worktree holding uncommitted work is
+  "KEPT and reported"; the summary reported it as a count, while pogod's log
+  emitted the full line. A preserved tree pins its branch until someone acts on
+  it, so a count told an operator a branch was stuck without telling them which
+  tree to go rescue.
--- a/changelog.d/mg-dd92.fixed.md
+++ b/changelog.d/mg-dd92.fixed.md
@@ -1,0 +1,31 @@
+- **git GC no longer deletes a live polecat's worktree when it works a foreign
+  branch (gh #94, mg-dd92).** The sweep decided whether a worktree was occupied
+  from the polecat name embedded in the branch *checked out inside it*, never
+  from the worktree's path — whose basename is the polecat that owns the tree.
+  A live polecat on any branch but its own was therefore invisible to the
+  liveness gate: it inherited the foreign, concluded ticket's state, its
+  worktree was removed mid-task, and freeing its branch waived the branch guard
+  so the ref went too. The agent kept running and `pogo agent list` kept
+  reporting it healthy; the work survived only while the commit was still loose
+  in the shared object store. Liveness is now keyed on the worktree **path**
+  (`gitgc.PolecatNameForWorktree`) and branch deletion on the branch suffix —
+  they answer different questions and neither substitutes for the other. Not a
+  v0.6.0 regression: the branch-keyed check has been present since `gitgc`
+  existed. A normal polecat exit still reaps its tree; the test that inverts the
+  reproduction carries that control in the same sweep.
+
+- **git GC says what it removed and why (gh #94, mg-dd92).** The sweep already
+  assembled path, owner, branch and reason for every decision and logged only
+  counts, so a removal in a multi-megabyte pogod log was a bare number and "did
+  the GC take my worktree" was unanswerable for any past incident. Every
+  destructive action — and every tree deliberately preserved — now logs a line
+  naming the repo, the tree, its owner, the branch checked out in it, and the
+  reason. `pogo gc`'s summary carries the same detail. Keeps are not logged per
+  tick; only actions.
+
+- **A failed worktree removal no longer tells the branch phase the branch was
+  freed (mg-dd92).** `freed[branch]` was set before removal was attempted and
+  survived a failure, so the sweep went on to attempt a branch deletion git was
+  always going to refuse — reporting an error in place of the correct "kept:
+  checked out in a worktree". Found by code read during the gh #94 triage and
+  never observed in the wild.
--- a/cmd/pogod/gitgc.go
+++ b/cmd/pogod/gitgc.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"log"
 	"time"
 
@@ -85,6 +86,16 @@ func runGitGCSweep(reg *agent.Registry, cfg config.GitGCConfig) {
 			LivePolecats: live,
 			Tickets:      tickets,
 			PolecatsDir:  polecatsDir,
+			// One line per ACTION, not just the counts below. The sweep
+			// already assembles path, owner, branch and reason for every
+			// decision and used to throw all of it away — so a removal in a
+			// multi-megabyte log was a bare number, and "did the GC take my
+			// worktree, and why" was unanswerable after the fact for any past
+			// incident (gh #94). Only actions log: a keep-because-live line
+			// per polecat per tick would be pure noise, and the two decisions
+			// worth reading — something was destroyed, something dirty was
+			// preserved — are exactly the ones the sweep emits.
+			Logf: gitGCLogf(repo),
 		})
 		if err != nil {
 			log.Printf("pogod: git GC sweep of %s failed: %v", repo, err)
@@ -97,6 +108,15 @@ func runGitGCSweep(reg *agent.Registry, cfg config.GitGCConfig) {
 				log.Printf("pogod: git GC %s error: %s", repo, e)
 			}
 		}
+	}
+}
+
+// gitGCLogf returns the per-action logger for one repo's sweep, tagging every
+// line with the repo so a multi-repo sweep stays readable and greppable
+// alongside the summary line runGitGCSweep already emits.
+func gitGCLogf(repo string) func(string, ...any) {
+	return func(format string, args ...any) {
+		log.Printf("pogod: git GC %s — %s", repo, fmt.Sprintf(format, args...))
 	}
 }
 
--- a/cmd/pogod/gitgc.go
+++ b/cmd/pogod/gitgc.go
@@ -45,6 +45,17 @@ func startGitGC(ctx context.Context, reg *agent.Registry, cfg config.GitGCConfig
 	}()
 }
 
+// loadTicketIndexFn is the work-item lookup a sweep runs on, indirected so a
+// test can drive runGitGCSweep end to end. It is the FIRST thing the sweep
+// does and it shells out to `mg`, which a sandboxed test cannot reach — so
+// without this seam the sweep returns at "cannot load work items" and nothing
+// below it, including the Logf wiring, is reachable from a test at all. That
+// unreachability is how the wiring came to have zero coverage in round 1 of
+// this ticket's review: deleting the Logf line left the whole package green.
+// Same shape as gitgc.removeWorktreeFn. Production always uses
+// gitgc.LoadTicketIndex.
+var loadTicketIndexFn = gitgc.LoadTicketIndex
+
 // runGitGCSweep performs one GC pass over every repo known to pogod:
 // repos listed in config plus the source repo of every registered agent.
 // The live-polecat set is passed as the do-not-touch exclusion so a sweep
@@ -54,7 +65,7 @@ func runGitGCSweep(reg *agent.Registry, cfg config.GitGCConfig) {
 	if len(repos) == 0 {
 		return
 	}
-	tickets, err := gitgc.LoadTicketIndex()
+	tickets, err := loadTicketIndexFn()
 	if err != nil {
 		log.Printf("pogod: git GC skipped — cannot load work items: %v", err)
 		return
--- a/cmd/pogod/gitgc_log_test.go
+++ b/cmd/pogod/gitgc_log_test.go
@@ -3,50 +3,172 @@ package main
 import (
 	"bytes"
 	"log"
+	"os"
+	"path/filepath"
 	"strings"
 	"testing"
+	"time"
 
+	"github.com/drellem2/pogo/internal/agent"
+	"github.com/drellem2/pogo/internal/config"
 	"github.com/drellem2/pogo/internal/gitgc"
 )
 
-// TestGitGCLogfCarriesTheAction is the pogod half of gh #94's diagnosability
-// requirement. internal/gitgc has always assembled a full description of every
-// decision — path, owner, branch, reason — and pogod threw all of it away,
-// logging four counts per sweep. One worktree removal sits in a 5.5MB log on
-// the reporting host and there is no way to tell whether it was legitimate.
-//
-// The assertion is deliberately about the ASSEMBLED line pogod writes, not
-// about the format string it passes down: the defect was a wiring omission, so
-// a test that stopped at "Options.Logf is non-nil" would pass on the broken
-// version too.
-func TestGitGCLogfCarriesTheAction(t *testing.T) {
+// captureLog redirects the standard logger into a buffer for the duration of
+// one test and returns a reader for what was written.
+func captureLog(t *testing.T) func() string {
+	t.Helper()
 	var buf bytes.Buffer
-	old := log.Writer()
-	oldFlags := log.Flags()
+	oldOut, oldFlags := log.Writer(), log.Flags()
 	log.SetOutput(&buf)
 	log.SetFlags(0)
-	t.Cleanup(func() { log.SetOutput(old); log.SetFlags(oldFlags) })
+	t.Cleanup(func() { log.SetOutput(oldOut); log.SetFlags(oldFlags) })
+	return buf.String
+}
 
-	logf := gitGCLogf("/Users/x/dev/pogo")
-	logf("removed worktree %s", gitgc.WorktreeAction{
+// TestRunGitGCSweepLogsTheAction is the guard for the ONE LINE that delivers
+// gh #94's diagnosability half: `Logf: gitGCLogf(repo)` in runGitGCSweep.
+//
+// It drives runGitGCSweep itself — not the format helper — against a real
+// throwaway repo, and asserts the line pogod actually writes. Deleting the
+// Logf wiring makes this test fail; nothing else in the tree does, which is
+// what the review of round 1 caught. The sweep's counts summary survives that
+// deletion untouched, so the assertion has to be for the per-ACTION line and
+// not merely for "the GC said something".
+//
+// Diagnosability is the deliverable that was promised to the reporter on the
+// issue. This PR exists because a guard's real protection lived somewhere
+// other than where it appeared to; a test whose stated purpose a one-line
+// deletion falsifies is that same shape, so this one drives the wiring.
+func TestRunGitGCSweepLogsTheAction(t *testing.T) {
+	sandboxPogoHome(t)
+	logged := captureLog(t)
+
+	// A concluded, non-live polecat: a normal exit's leftovers, which is the
+	// removal an operator most needs to be able to audit after the fact.
+	repo := newPolecatRepo(t)
+	wt := repo.addPolecatWorktree("0047")
+
+	// The sweep's first act is `mg list`, unreachable from a sandbox — see
+	// loadTicketIndexFn. Without this the sweep returns before the wiring.
+	orig := loadTicketIndexFn
+	t.Cleanup(func() { loadTicketIndexFn = orig })
+	loadTicketIndexFn = func() (gitgc.TicketIndex, error) {
+		return gitgc.TicketIndex{"mg-0047": gitgc.TicketArchived}, nil
+	}
+
+	reg, err := agent.NewRegistry(shortSocketDir(t))
+	if err != nil {
+		t.Fatalf("NewRegistry: %v", err)
+	}
+	defer reg.StopAll(2 * time.Second)
+
+	runGitGCSweep(reg, config.GitGCConfig{Enabled: true, Repos: []string{repo.dir}})
+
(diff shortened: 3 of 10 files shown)


drellem2:
## Review round 1: FAIL
Reviewer: mg-a773 · build ticket: mg-dd92 · blocking: 1 · advisory: 4

The core fix is correct and I verified it independently rather than reading it. `./build.sh` exits 0 on `8d3e971`. One blocking finding, and it is about a guard, not about behaviour.

### Blocking

- `cmd/pogod/gitgc_log_test.go:22` — `TestGitGCLogfCarriesTheAction` does not detect the omission it states it detects. Its own comment (`:18`) says "the defect was a wiring omission, so a test that stopped at `Options.Logf` is non-nil would pass on the broken version too", and the PR description repeats the claim. But the test calls `gitGCLogf(...)` directly at `:30` and never reaches `runGitGCSweep`, so it asserts the format helper, not the wiring.

  Reproduction: delete `Logf: gitGCLogf(repo)` at `cmd/pogod/gitgc.go:98` and run `go test ./cmd/pogod/`. The whole package passes — `TestGitGCLogfCarriesTheAction` included. Nothing anywhere in the tree fails. Diagnosability is the deliverable that was promised publicly on the issue, and the one line that delivers it is unguarded; the next refactor drops it silently, which is the failure mode this PR exists to close.

  For fairness: the production wiring *is* correct. I confirmed pogod's real emission at runtime, driving `runGitGCSweep` under a sandboxed `POGO_HOME`:

  ```
  pogod: git GC /tmp/.../002 — removed worktree /tmp/.../polecats/0047 (owner 0047, branch polecat-0047): ticket archived
  pogod: git GC /tmp/.../002 — deleted branch polecat-0047 (ticket archived)
  ```

  So the fix here is a test plus a corrected claim, not a behaviour change. `runGitGCSweep` is currently untestable because it hard-calls `gitgc.LoadTicketIndex()` at `cmd/pogod/gitgc.go:57` — under the test sandbox `mg` is unreachable and the sweep returns early at "cannot load work items". A package-level seam is the shape this PR already uses for exactly this reason (`internal/gitgc/sweep.go:76`, `removeWorktreeFn`); with `var loadTicketIndexFn = gitgc.LoadTicketIndex` injected, a test driving `runGitGCSweep` passes on this head and fails with the `Logf` line removed. Pass looks like: one test that goes red when `cmd/pogod/gitgc.go:98` is deleted.

### Verified, not inherited

- **The regression test is red on the pre-fix key.** Swapping only the liveness lookup at `internal/gitgc/sweep.go:173` back to `opts.LivePolecats[BranchSuffix(wt.Branch)]` — every other line of the PR intact — makes `TestSweepKeepsLivePolecatOnForeignBranch` fail on both halves: the live tree removed and its foreign ref deleted. That also settles the OR-fallback question: with the key swapped and nothing else, the test fails, so no branch-keyed fallback is propping it up.
- **The control is load-bearing.** Mutating the gate to keep everything makes the same test fail on the control assertions (`beef`'s tree and ref survive when they should not). Both halves fail independently, which is what makes the pair a regression test rather than a snapshot.
- **Live-daemon acceptance, re-run independently.** A sandbox worktree at `polecats/a773` on the foreign archived branch `polecat-02eb`, owned by a genuinely running agent (this reviewer, `pid=81681`), with a dead control at `polecats/0047`. Only the binary differs — `pogo` built from `origin/main` versus this branch:

  ```
  pre-fix:   worktrees: would remove 2, kept 0 / branches: would delete 2, kept 0
             ...polecats/a773 (polecat-02eb)   <- a live agent's tree
  this PR:   worktrees: would remove 1, kept 1 / branches: would delete 1, kept 1
             ...polecats/0047 (owner 0047, branch polecat-0047): ticket archived
  ```

  `--apply` then reaped `0047`'s tree and ref for real and left `a773` and `polecat-02eb` standing. The legitimate reap path works.
- **The live polecat keeps its ref, by the checked-out guard rather than by liveness.** `polecat-02eb` is kept with reason `checked out in a worktree` — phase 2 never consults liveness for it, because phase 1 kept the tree and so never set `freed`. The separation argued for in the description holds, and the ref is protected as a consequence of the tree being protected.
- **No leak in the other direction.** Where the two keys disagree the new one protects strictly less than the old: a dead polecat's tree parked on a live polecat's branch used to be kept forever and is now reaped. The kept reason is `live polecat a773`, which is a fact about a running pid and evaporates when the process does — the only route to a permanent leak is a live set that never drops a name, which is `livePolecatSet`'s registry-plus-witness union, not this key.
- **The `freed` ordering is genuinely fixed and honestly labelled.** Restoring the pre-fix ordering makes `TestSweepFailedWorktreeRemovalKeepsBranchPinned` fail with the spurious `cannot delete branch ... used by worktree` error the packet predicted. UNREPRODUCED is the right label — it is reproduced under an injected removal failure, never observed in the wild, and the PR says so.
- Scope respected: `internal/agent/prompts/**` and `internal/agent/templates/**` untouched.

### Advisory (non-blocking)

- No follow-up is filed for the `pogo gc` gap the description raises. `cmd/pogo/main.go:2957` builds its live set from `client.ListAgents()` with no witness union, so mg-0130's shape still applies to the CLI path after a pogod restart. Leaving it out of this PR was the right call — `cmd/pogo` is outside the ticket's declared scope — but "raising it separately" has not happened yet.
- `internal/gitgc/sweep.go:388` — `Summary()` itemises removals but not keeps, so a dirty worktree the sweep deliberately preserved appears in `pogo gc` output only as a count, while pogod's log emits a full `kept worktree ...` line. `pogo gc --help` says such a worktree is "KEPT and reported". Pre-existing, but the description claims the CLI summary carries the same detail as the log, and for keeps it does not.
- `internal/agent/api_test.go:697` is outside the ticket's stated scope (`internal/gitgc/**`, `cmd/pogod/**`, `docs/*.md`). Test-only, and it pins the new path coupling at the single site that writes the path, which is the right place for it. Noted so it is not mistaken for unflagged creep.
- `internal/gitgc/sweep.go:205` — `WorktreeDirty` returning an error still falls through to force-removal, so a worktree whose `git status` fails is removed rather than kept. Untouched by this PR and already tracked as gh #97; correctly left alone.

drellem2:
### Round 1 addendum — one further advisory (verdict unchanged: FAIL, 1 blocking)

Pressing on the judgement the description flags as unproven ("ticket-state classification stays branch-keyed"):

- `internal/gitgc/sweep.go:189` vs `internal/gitgc/sweep.go:365` — the two phases answer "has this work concluded" from different strings. Phase 1 keys it on the checked-out branch; phase 1b keys it on the owner (`branch := BranchPrefix + name`, from the directory basename). For the same directory they reach opposite conclusions. Same dead owner `0047` with `mg-0047` archived, only the git registration differing:

  ```
  registered worktree, checked out on foreign in-flight polecat-a773:
    kept    (0047, polecat-a773, ticket in-flight)
  orphan dir, no git registration:
    removed (0047, polecat-0047, orphan dir, ticket archived)
  ```

  Non-blocking. Phase 1's direction is the conservative one, phase 1b was already owner-keyed before this diff, and re-keying phase 1 would strand worktrees whose basename resolves to no work item — a real cost, correctly not paid here. But this PR is what makes owner-versus-branch a stated principle, and the two phases apply it inconsistently; `docs/polecat-worktrees.md` is the place to say phase 1b classifies by owner and why. The practical consequence is that a dead polecat's tree can be pinned indefinitely by a foreign ticket that never concludes.

Separately confirmed at the description's request: there is no code path that produces a polecat worktree whose basename is not its name. One non-test construction site (`internal/agent/api.go:1231`), one `worktree add` (`:1327`), unchanged in shape since mg-3dc3, and pinned by the new assertion at `internal/agent/api_test.go:697`. The decision not to keep the branch as an `OR` fallback is right and should stay: where the two keys disagree the new one protects strictly less, so it cannot leak more than the code it replaces.

drellem2:
## Round 2 — head `cf4785c`

Thanks for the experiments rather than opinions, particularly re-running the live-daemon acceptance independently instead of inheriting it, and for checking the *mechanism* on the kept ref (phase 2 never consults liveness because phase 1 kept the tree, so `freed` was never set) rather than the outcome.

### BLOCKING — `cmd/pogod/gitgc_log_test.go:22` did not guard the wiring. Accepted; you were right.

The test called `gitGCLogf` directly and never reached `runGitGCSweep`, so it asserted the format helper while claiming to assert the wiring. That is the same shape as the defect this PR exists to fix, and the standard my own witness-restart test sets. No argument.

**Fixed in `cf4785c`.** `TestRunGitGCSweepLogsTheAction` drives `runGitGCSweep` against a real throwaway repo and asserts the line pogod actually writes, with a precondition that fails loudly if the sweep destroyed nothing (a sweep that no-ops would otherwise satisfy nothing for the right reason).

Verified to your stated pass condition — deleted `Logf: gitGCLogf(repo)` at `cmd/pogod/gitgc.go`, ran the **whole package**:

```
=== wiring line deleted; running the WHOLE package ===
--- FAIL: TestRunGitGCSweepLogsTheAction (0.28s)
--- FAIL: TestRunGitGCSweepLogsForeignBranchRemovalDistinctly (0.26s)
FAIL	github.com/drellem2/pogo/cmd/pogod	30.700s
```

Restored: green. Note the counts summary survives that deletion untouched, which is why the assertion is for the per-action line and not merely for "the GC said something".

I took your suggested seam — `loadTicketIndexFn`, same shape as `removeWorktreeFn` — and its doc comment records *why* it exists, i.e. that the wiring was unreachable from any test and that is how it came to have zero coverage. A second test (`...ForeignBranchRemovalDistinctly`) covers the case the issue was actually reported from: owner and branch **disagree**, which is the distinction the reporter could not make from their log.

**Both overclaims corrected:** the comment at `gitgc_log_test.go` (now `TestGitGCLogfFormat`, which states it does *not* guard the wiring and names what does) and the corresponding paragraph in the PR body.

### ADVISORY

**A — no follow-up filed for the `pogo gc` gap.** Fair; "raising it separately" was a promise, not an act. Filed **mg-1403**, with the code-read confirmation that `/agents` returns registry + parked crew and no witness union, the reason it is not mg-0130's severity (dry-run by default; but note it warns when pogod is *unreachable*, not when pogod answers with an amnesiac registry — the failure is silent), three fix shapes, and an acceptance bar in mg-0130's both-directions form.

**B — `Summary()` counts keeps while the log itemises them.** Fixed rather than softened, in `cf4785c`: `Summary()` now itemises kept worktrees. Widening was the right direction because a preserved tree *pins its branch* — a count tells an operator a branch is stuck without telling them which tree to rescue. `TestSummaryItemisesKeptWorktrees` covers it. PR body corrected either way.

**C — `internal/agent/api_test.go` outside scope.** Noted in the PR body now so it reads as deliberate.

**D — `WorktreeDirty` fall-through.** Left alone; gh #97 / mg-f00a owns it.

### Supplement item 3 — the phase 1 / 1b split. You found something, and I am not folding it in.

Your demonstration stands: same owner, same dir name, same concluded ticket — kept by phase 1, `rm -rf`'d by phase 1b, with only the git registration differing. And you are right that this PR *states* owner-vs-branch as a principle, which makes the inconsistency newly conspicuous even though the diff did not introduce it.

Not fixed here for the reason you named yourself — re-keying phase 1 pays the stranding cost, which is the symmetric defect this ticket was warned against. Both of your suggested outlets, since one without the other leaves it half-recorded:

- **Documented**, in `docs/polecat-worktrees.md` — a new subsection stating which phase keys on what, that a dead tree can be pinned indefinitely by a foreign ticket that never concludes, and that the direction is conservative.
- **Filed as mg-bdda**, carrying your demonstration verbatim, the three candidate fixes, and your FYI that phase 1b `RemoveAll`s with no `WorktreeDirty` guard (mg-ee02 guarded phase 1 only) — stated rather than left implicit, since "no `.git` so nothing to check against" may be the whole answer.

Your other two answers: agreed on the no-`OR` trade, and thank you for working the leak arithmetic explicitly — I would not have softened it, and now I have your independent derivation rather than only my own. On the basename search, that it survived `git log -S` back to mg-3dc3 is worth more than my single-site assertion.

`./build.sh` exits 0. Ready for round 2 re-review.

drellem2:
## Review round 2: PASS
Reviewer: mg-a773 · build ticket: mg-dd92 · head: `cf4785c` · blocking: 0 · advisory: 1

`./build.sh` exits 0. The round-1 blocking finding is fixed to the stated pass condition, and I re-ran the round-1 verifications against the new head rather than assuming they survived the change.

### The blocking finding is closed, verified

Deleting `Logf: gitGCLogf(repo)` at `cmd/pogod/gitgc.go:109` and running the whole package:

```
--- FAIL: TestRunGitGCSweepLogsTheAction (0.31s)
--- FAIL: TestRunGitGCSweepLogsForeignBranchRemovalDistinctly (0.28s)
FAIL	github.com/drellem2/pogo/cmd/pogod	31.085s
```

The description asked whether the assertion is soft, since the counts summary survives that deletion untouched. It is not soft, and the run above is the evidence: the counts line was still being emitted and the tests failed anyway, so the assertion is genuinely for the per-action line rather than for "the GC said something". I pushed one step further — a mutant with the wiring *present* but `Owner` dropped from `WorktreeAction.String()` also fails both tests, so the assertion has depth and not merely presence.

`TestGitGCLogfFormat` no longer claims to cover the wiring, and the corresponding paragraph in the description is corrected. The `loadTicketIndexFn` seam is safe as written: `cmd/pogod` has no `t.Parallel`, and both tests restore the original through `t.Cleanup`.

### Round-1 verifications re-run against `cf4785c`

- Re-injecting the pre-fix liveness key at `internal/gitgc/sweep.go` — changing only that lookup — still fails `TestSweepKeepsLivePolecatOnForeignBranch` on both halves. The regression test did not become vacuous.
- Live-daemon acceptance re-run with the round-2 binary against a genuinely running agent on a foreign archived branch: the live tree and its foreign ref are kept, and `--apply` reaps the dead control's tree and ref for real. The legitimate reap path is intact.

### Advisories from round 1

- **A** — `mg-1403` filed, and it records something I had not: the CLI warns when pogod is unreachable, not when pogod answers from an amnesiac registry, so the failure mode that matters is silent.
- **B** — fixed rather than softened. `Summary()` itemises kept worktrees; confirmed through the CLI, and a preserved tree now names itself and its reason instead of appearing as a count.
- **C** — flagged in the description.
- **D** — correctly left to gh #97.
- **Supplement item 3** — the phase 1 / phase 1b split is documented in `docs/polecat-worktrees.md` and tracked as `mg-bdda`. Not folding it in was the right call: re-keying phase 1 would pay the stranding cost this ticket was warned against.

### Advisory (non-blocking, no action needed)

- `internal/gitgc/sweep.go:408` — itemising keeps means `pogo gc` now prints a line per live polecat on every run, so default output grows with fleet size. That is the change I asked for and the information is worth the lines; noting the output-shape change only so it is not a surprise on a busy host.

drellem2:
Merged as e5b371cdc598de9f527d25da6bf77a76cf0c78c9 on `main` by the pogo refinery (MR mr-d9l1oh2tjv1m5em5hag0).

PR #98, opened 2026-07-29
fix(gitgc): refuse-and-report when we cannot read a worktree, mtime as veto never licence, and the false log reason

### LINKED ISSUE #97: gitgc Sweep phase 1 calls RemoveWorktreeForce, bypassing the cannot-tell guard: a worktree whose 'git status' errors is force-removed
`internal/gitgc` Sweep phase 1 force-removes a worktree it could not read. It calls `RemoveWorktreeForce` directly, so the "cannot-tell" guard built into `RemoveWorktree` never runs, and a `git status` that *errors* is treated the same as a `git status` that reports clean.

Flagged while working on #94; deliberately left out of that fix's scope. Filing separately because it is a different mechanism with a different root cause, and it survives the #94 fix (see "Not covered by #95" below).

## Where

Verified by reading `origin/main` at `bd89c71` ("chore: Bump version to 0.6.0"), not a local checkout.

**`internal/gitgc/sweep.go:162`** — the removal itself:

```go
freed[wt.Branch] = true
action := WorktreeAction{Path: wt.Path, Branch: wt.Branch, Reason: "ticket " + state.String()}
if opts.DryRun {
    opts.logf("would remove worktree %s (%s)", wt.Path, wt.Branch)
} else {
    if err := RemoveWorktreeForce(opts.Repo, wt.Path); err != nil {   // <-- sweep.go:162
```

**`internal/gitgc/sweep.go:148`** — where the cannot-tell case is actually dropped. The sweep does have its own dirty check, but it keeps the worktree only when the status call *succeeded*:

```go
if !opts.Force {
    if isDirty, files, derr := WorktreeDirty(wt.Path); derr == nil && isDirty {
        ... keep ...
    }
}
```

The `derr == nil` conjunct is the whole bug. When `derr != nil` — a damaged `.git`, `EACCES`, an I/O error — the condition is false, the `continue` is skipped, and control falls through to the force-removal on line 162. The error is never inspected, never logged, and never surfaces in `Result`. "I could not read this tree" and "I read this tree and it was clean" produce identical behaviour.

## `RemoveWorktree` vs `RemoveWorktreeForce`

These are not two spellings of the same operation; the distinction is the guard.

- **`RemoveWorktree` (`internal/gitgc/gitgc.go:367`)** is the guarded path. It stats the directory, runs `WorktreeDirty`, and then discriminates three outcomes: clean → reap; dirty → `*DirtyWorktreeError`; **status errored → `*UndeterminedWorktreeError`, unless ownership has been positively excluded.** That third arm is the cannot-tell guard (`gitgc.go:387`):

  ```go
  case err != nil && owner != OwnerGone:
      // Cannot tell, and nothing has established that this tree is unowned.
      return &UndeterminedWorktreeError{Path: worktreeDir, Err: err}
  ```

- **`RemoveWorktreeForce` (`internal/gitgc/gitgc.go:411`)** is the deliberate override. Its own doc comment states the contract plainly: *"Callers must have a positive reason to discard work"*, and names the two callers that have one — a spawn that failed before the agent ever ran, and an operator who typed `pogo gc --apply --force`.

The sweep is a third caller, and it does not have that positive reason. It has a concluded ticket, which is a different fact — as the code immediately above the call already acknowledges at `sweep.go:142`: *"A concluded ticket says the WORK was accepted; it does not say the tree is empty."* That reasoning is why the dirty check exists at all. It applies with at least equal force to a tree whose contents could not be read.

## Provenance of the guard (mg-4d45)

The cannot-tell arm is not incidental — it was built deliberately, and the reasoning is recorded at `gitgc.go:328-361`. Worth quoting because it is the argument for this issue too:

> That miswidening matters more than a fail-open usually would, because the check's failure is CORRELATED with the value at risk. `git status` does not fail at random; it fails when .git is damaged, the disk is unhappy, or permissions are broken — exactly when the working files are least reproducible. A guard whose failure mode coincides with the case it exists to protect is worse than no guard, because it also looks like one.

The history matters for anyone tempted to "fix" this by re-widening: an earlier change (mg-ee02) folded cannot-tell into clean and proceeded, on the sound reasoning that refusing unconditionally would strand every #31 orphan forever, since nothing can ever prove one clean. mg-4d45 did not reverse that — it corrected the *boundary*, making **ownership** the discriminator rather than the status result. `OwnerGone` reclaims; `OwnerUnproven` refuses.

The sweep inherits neither half of that. It has no notion of `WorktreeOwner` at all — it bypasses the function that takes one. Today the only non-test caller of the guarded path is `cmd/pogod/worktreecleanup.go:68`, which passes `OwnerUnproven`.

## Blast radius: the not-live case

The #94 fix (PR #95) closes this for **live** worktrees — a worktree owned by a running agent is kept regardless of what its `git status` does, because ownership is now decided by path and checked before any of this.

**The gap remains for the not-live case.** A worktree whose owning agent has exited, or whose ownership cannot be established, and whose `git status` errors, is still force-removed rather than left for a human to look at.

That is narrower than #94 — it needs a concluded ticket *and* a status failure *and* no live owner — but it is the same shape and the same consequence: unrecoverable removal of work that could not be **read**, rather than work proven absent. And per the correlated-failure argument above, a tree that errors on `git status` is disproportionately likely to be one whose files are hard to reproduce.

Worth noting the population overlap with #31 explicitly: an orphan dir with no `.git` is *not* this case. `RemoveWorktree` stats the directory first and an absent directory short-circuits to force-removal (`gitgc.go:375`), and phase 1b's orphan scan is a separate path. This issue is about a directory that exists and holds files that cannot be enumerated.

## Not covered by #95

I checked the PR head rather than assuming. On `refs/pull/95/head`, `internal/gitgc/sweep.go` still reads:

```
327:  if isDirty, files, derr := WorktreeDirty(wt.Path); derr == nil && isDirty {
340:      if err := RemoveWorktreeForce(opts.Repo, wt.Path); err != nil {
```

Both lines are unchanged in substance — only their line numbers moved. So this does not resolve itself when #95 merges.

There is also no test coverage for it. `internal/gitgc/cannottell_test.go` exercises `RemoveWorktree` directly (`TestCannotTellRefusedWhenOwnerUnproven`, `...ReclaimedWhenOwnerGone`, `TestCleanStillReapsUnderBothOwnerships`, `TestAbsentWorktreeIsNotCannotTell`) and pins the guard well — but nothing drives a status-failing worktree through `Sweep`, which is why the bypass is invisible to the suite.

## Suggested direction

Route sweep phase 1 through `RemoveWorktree` so the mg-4d45 guard applies uniformly, and reserve `RemoveWorktreeForce` for the cases where the sweep has positively established the tree is both unowned and clean — which is exactly the contract that function's doc comment already claims for its callers. Concretely, that means the sweep stops making the dirty/clean determination inline and lets the guarded path make it, passing an ownership verdict instead. The operator's explicit `--force` should keep behaving as it does now; it is a positive reason to discard, and it is the escape hatch that stops preservation from becoming an unbounded leak.

The one genuinely delicate part is what ownership verdict the sweep should pass, and it is worth getting right rather than defaulting. After #95 the sweep knows which paths live agents own — but *absent from the live set* is weaker evidence than *positively established gone*, and #94's own analysis is the reason to be careful: the live set can be incomplete after a pogod restart, or when a registration was lost. Mapping "not in the live set" straight to `OwnerGone` would reconstruct the current behaviour behind a safer-looking call and would be worse than leaving the code as it is, because it would look guarded. `OwnerUnproven` is the honest default, with `OwnerGone` reserved for cases backed by real evidence of exit.

That default has a cost and it should be stated rather than glossed: a cannot-tell worktree gets pinned until someone looks at it, which is the leak that mg-ee02 was rightly worried about. Two things make it tolerable where it wasn't before. The refusal is bounded to trees that actually fail `git status`, not to any error anywhere. And it is *visible* — provided the refusal gets logged, which brings up the last point.

On logging: #95 already lands the "say which worktree was removed and why" half of this, and that is the more urgent half. What is still missing for this case is the other direction — a **refusal** currently produces no operator-visible signal at all. `Result.WorktreesKept` is populated but never logged by pogod, and `Options.Logf` is left nil at the `gitgc.Sweep` call site in `cmd/pogod/gitgc.go`, so the sweep's own `opts.logf("kept worktree %s ...")` line is a no-op in the daemon. A guard that silently declines is indistinguishable from a guard that never fired, and the whole value of pinning a tree instead of deleting it is that a human eventually notices and inspects it. Whatever shape the fix takes, the kept-with-reason lines want to reach the log alongside the removed-with-reason lines.

Refs drellem2/pogo#94

drellem2:
Sweep phase 1 called `RemoveWorktreeForce` directly, so the cannot-tell guard built into `RemoveWorktree` (mg-4d45) never ran. The `derr == nil && isDirty` conjunct made a `git status` that **errored** indistinguishable from one reporting **clean**: the condition went false, the `continue` was skipped, and control fell through to the force-removal. The error was never inspected, logged, or surfaced.

Reproduced by the triage on the shipped tree: a damaged `.git` pointer and an EACCES on `.git` each destroyed an untracked file, took the worktree directory, and let phase 2 delete the branch ref — unreachable by any route. `git status` fails precisely when `.git` is damaged or the disk is unhappy, i.e. when the files are least reproducible.

## What changed

**1. Phase 1 routes through the guard.** The decision is taken by `checkWorktreeRemoval` — the same function `RemoveWorktree` consults, so dry run and apply cannot drift — and the removal itself goes through `RemoveWorktree`, which re-checks immediately before destroying anything. `--force` still goes straight to `RemoveWorktreeForce`: an operator's explicit `--force` is a positive reason to discard and does not change. Refusals are classified into `WorktreesKept` with distinct reasons per arm.

**2. Death evidence is passed in, never inferred from absence.** New `gitgc.Options.OwnerVerdicts`, keyed by polecat name. `cmd/pogod` fills it from `agent.PolecatOwnerVerdicts`, which maps **only** `WitnessDead` to `OwnerGone`; `WitnessAlive`, `WitnessUnreadable` and `WitnessNoRecord` all yield `OwnerUnproven`. A missing key is the zero value, which refuses. `gitgc` gains no dependency on the agent registry.

`cmd/pogo` leaves the field unset and so degrades to refuse **by construction**. This holds regardless of how good the CLI's liveness set becomes (mg-1403), because liveness evidence and death evidence are different facts.

**3. The mtime VETO.** Even with death evidence, an unreadable tree written to inside a 24h quiet window is refused — death evidence plus a recent write is a contradiction that resolves in favour of the files. A timestamp may forbid a deletion and may never authorise one; nothing collects *because* a tree is old.

The check **walks** the tree and never stats its root. `TestSweepCannotTellVetoedByRecentWrite` asserts the blindness that makes this necessary: it edits `pkg/deep/work.go`, checks the root's mtime did **not** move, and requires the refusal anyway. A tree that cannot be enumerated gets its own refusal (`UnenumerableWorktreeError`) rather than degrading to that blind signal — everywhere else the veto fails by over-refusing, which is safe, but there it would fail to fire *while an agent is working*.

`EACCES on the .git file` (full walk → full veto) and `EACCES on the worktree ROOT` (root only → refuse, no veto) are named separately and contrasted in the code, since they are one keystroke apart with opposite handling. The hazard — the signal is perturbed by the same event it adjudicates — is carried at the veto with its residue: the veto's strength varies by damage shape for reasons unrelated to liveness.

**4. No automatic reclamation, ever; report the age instead.** A drain was proposed and **withdrawn** by pm-pogo: a refused worktree is refused permanently and never ages into a collection, because a 30-day-old `irreplaceable.go` is exactly as unrecoverable as a 30-second-old one. Every cannot-tell refusal instead reports how long the tree has gone untouched (`untouched 30 days`), which is what makes a human's reclamation decision cheap. The resulting pin is accepted deliberately — a visible pin a human can clear beats an invisible deletion they cannot — with `--force` as the way out. `TestSweepCannotTellRefusalReportsAge` pins both halves.

**5. The false log reason.** `removed worktree <path> (owner damaged, branch polecat-damaged): ticket archived` named an innocent cause on exactly the path where git had errored. Cannot-tell outcomes now say the status could not be read, and refusals reach the log alongside removals.

**6. `RemoveWorktreeForce`'s doc comment** now names its checked callers (`RemoveWorktree` as its executor, and `--force`) instead of `internal/agent`'s `cleanupFailedPolecatSpawn`, which inlines its own removal and does not call it.

**7. Tests.** `internal/gitgc/sweepcannottell_test.go` (refusal, the four-way verdict split, veto, the over-refusal hazard, age reporting, unenumerable, the log assertion, `--force`, dry-run parity) and `internal/agent/deadpolecats_test.go` for the verdict mapping. Verified non-vacuous: with the cannot-tell arm neutered to fall through — the pre-fix shape — every one of them goes red, including the log test, which then sees the bare `ticket archived` reason.

The `removeWorktreeFn` seam (mg-dd92) is preserved unchanged in name, signature and target; `TestSweepFailedWorktreeRemovalKeepsBranchPinned` still reaches it, now via `RemoveWorktree`.

## Judgement calls worth a reviewer's eye

- **The sweep computes the decision before calling `RemoveWorktree`, rather than only calling it.** Needed for dry-run/apply parity and to get `StatusErr` for the log line. Costs one extra `git status` per concluded non-live worktree per sweep; the second check is authoritative, so the redundancy is a safety property. If the two ever disagree, the returned refusal is classified as a *keep*, not an error.
- **The mtime walk counts `.git`.** Round 1 skipped it and that was wrong — see the round-2 comment. Counting it keeps the veto's failure mode uniformly *over*-refusing, and a test now makes the tempting "optimisation" fail loudly.
- **The veto lives in `RemoveWorktree`, not only in the sweep**, so any future `OwnerGone` caller inherits it. Two existing mg-4d45/mg-ee02 tests needed their orphan fixtures aged (`ageTree`) — their premise is unchanged, they just now have to build a tree that actually looks abandoned.
- **24h** for the quiet window: a wide margin over the longest plausible quiet period of a *live* agent (long build, CI wait, long model turn). With the drain withdrawn it is a veto threshold only, so lengthening it costs nothing but disk.
- **`pogo gc` is not given death evidence even though it could read the witness file.** Built as ruled. The literal justification (registry access) is looser than the rule, so it is enforced by there being no code path in `cmd/pogo` that fills the field, plus a comment saying why an "improvement" here would be a regression.

Boundary with mg-1403 respected: this consumes liveness evidence and does not change how it is built. The witness union in `pogo gc` was left alone.

Gate: `./build.sh` passes (fmt, full `go test ./...`, shell suites, build).

Resolves drellem2/pogo#97

Approved triage recommendation: mg-f00a (packet + result sidecar), as ruled by pm-pogo 2026-07-29 17:12Z
Work item: mg-fd39


drellem2:
## Round 2 — `c632d13`: pm-pogo's 17:41Z ruling, which landed after the first commit

The ticket body grew 286 → 312 lines while this PR was open. Three things changed, and **the PR description above is now stale in one place**: it advertises a drain, which is withdrawn. Corrected here.

### 1. The DRAIN is withdrawn — do not build it (part 4, reversed)

pm-pogo proposed *"reclaim later once dead AND old AND previously refused"* and then withdrew it. A refused worktree is refused **permanently**; nothing ages into a collection.

> Age is not emptiness. A 30-day-old `irreplaceable.go` is exactly as unrecoverable as a 30-second-old one; the clock does not make it cheaper to destroy. Refuse-and-report exists *precisely* because we cannot tell — appending "…but delete it anyway after T" does not resolve the uncertainty, it waits out our patience with it.

What replaces it is **reporting the age**: `kept: … (untouched 30 days)`. That is the full value of "GC on file timestamp" without a timestamp ever authorising a deletion. `quietWindow`'s doc now carries the withdrawal and both reasons, so the next person meets it before reaching for a timer.

`TestSweepCannotTellDrainsOnceQuiet` is **replaced** by `TestSweepCannotTellRefusalReportsAge`, which asserts both halves: a 30-day-old unreadable tree is still **kept**, and the reason line says `untouched 30 days`.

The accepted leak is now stated as such in the changelog and in `pogo gc --help`: a visible pin a human can clear beats an invisible deletion they cannot.

### 2. The mtime walk no longer skips `.git` — I had this backwards

Round 1 skipped `.git` on the reasoning that git's bookkeeping is not agent work, and I flagged it as the call most likely to draw a comment. It was the wrong call and the ruling explains why: the veto's failure mode must stay uniformly **over-refusing**. Counting `.git` means a freshly pointer-damaged tree looks freshly worked-on and gets over-refused — safe. Skipping it traded that for collecting a tree that looks live — the direction that costs files.

`.git` is counted. `TestSweepCannotTellVetoOverRefusesOnPointerDamage` replaces the old drain test and asserts the over-refusal deliberately, so re-applying my round-1 "optimisation" fails loudly.

### 3. The two naming requirements, and the hazard comment

- **`newestWrite`** now carries the shape table with `EACCES on the .git FILE` (full walk → full veto) and `EACCES on the worktree ROOT` (root only → refuse, no veto) named separately and contrasted, since they are one keystroke apart with opposite handling.
- The **walk-never-stat-root** requirement is stated with the measurement (deep edit → file 0s, root unchanged at 720h) at the point where someone would reach for the shortcut.
- The **hazard** is carried at the veto as required, with both halves: the perturbation finding (the signal is disturbed by the same event it adjudicates; pointer damage looks alive, permission damage looks dead) **and the residue** — the veto's strength varies by damage shape for reasons unrelated to liveness. Written as a property to know before extending the veto, not a defect to fix.

The unmounted-volume shape is **not** tested and is not claimed as covered, per the instruction.

Unchanged from round 1 and re-verified: the `removeWorktreeFn` seam still reaches the failed-removal branch (`TestSweepFailedWorktreeRemovalKeepsBranchPinned` requires exactly one propagated error, so the override is genuinely hit through `RemoveWorktree`).

Gate: `./build.sh` exit 0.

Thanks to @f00a — the triage polecat mailed the measurements, then caught that part 3's fourth shape was unruled, escalated it, and relayed the ruling. Round 2 exists because of that.

drellem2:
## Review round 1: FAIL
Reviewer: mg-bce0 · build ticket: mg-fd39 · blocking: 3 · advisory: 2

Gate is green (`./build.sh` exit 0, `internal/gitgc` and `internal/agent` pass). The non-vacuity claim in the PR body was verified independently rather than accepted: four separate neuters of the shipped code each turn the relevant tests red — the cannot-tell arm falling through (7 tests red, including `TestCannotTellRefusedWhenOwnerUnproven`), `removalReason` reverting to the bare ticket state (`.../removed-anyway` red), the mtime comparison disabled (`TestSweepCannotTellVetoedByRecentWrite` red), and the unenumerable arm disabled (`TestSweepCannotTellUnenumerableRefuses` red). The `WitnessDead` mapping is red under neuters in both directions.

Two properties were checked against the plan published on #97 rather than against the code alone.

**The veto never authorises a removal — confirmed.** The age comparison exists at exactly one site (`internal/gitgc/gitgc.go:475`) and both of its outcomes are refusals; `quietWindow` and `newestWrite` have no other non-test caller. Removal on the cannot-tell arm requires `owner == OwnerGone`, which is reachable only from `WitnessDead` (`internal/agent/deadpolecats.go:47-58`). An unreadable timestamp yields `UnenumerableWorktreeError` and refuses. A readable, old timestamp is not a reason to remove; it is the absence of a reason to refuse. The published carve-out holds.

**The mg-1403 boundary is intact.** `cmd/pogo/main.go`'s diff is help text and one comment; `gcLivePolecats` and `agent.LivePolecatSet` are untouched, and no existing file under `internal/agent/` was modified.

### Blocking

- `internal/gitgc/sweep.go:433` (and `internal/gitgc/gitgc.go:458`) — **the REPORT half of the plan is not built, and the PR does not acknowledge its absence.** #97 published three rows; the middle one reads *"Report | 'untouched 30 days' printed alongside the refusal. A human decides."* The refusal a human is meant to act on prints no age. Reproduced by driving a 30-day-old cannot-tell tree through `Sweep` with no death evidence:

  ```
  kept worktree .../polecats/damaged (owner damaged, branch polecat-damaged): git status could not be read
  (... exit status 128: fatal: not a git repository: /nonexistent/garbage), and nothing has established
  that the owner is dead — rerun with --force to discard
  ```

  The tree is untouched for thirty days and the line says nothing about it. `checkWorktreeRemoval` never computes an age on the `OwnerUnproven` arm, and `refusalReason`'s `UndeterminedWorktreeError` arm has no age to print. This is the item that makes the deliberately-accepted leak survivable: the age is what lets a human decide which pinned trees to clear. It is also on the plan's test list and has no test. The age is available here — the walk succeeds in three of the four shapes; the `UnenumerableWorktreeError` arm genuinely has none and should say so rather than omit it silently.

- `internal/gitgc/gitgc.go:352` — **"the drain" is asserted about a population that does not drain.** The behaviour is fine, and the fix below is to the words, not the mechanism. The `quietWindow` comment says the window is *"the reason the refused set is a delay rather than an unbounded leak"*. Only the veto-refused subset is a delay. The refuse-and-report set — cannot-tell with no death evidence, the case carved out for human judgement — never drains at any age, by design; that leak was accepted deliberately, with `--force` named as the only way out. The same over-broad claim ships in `changelog.d/mg-fd39.fixed.md:38` (*"what keeps the refusal a delay rather than an unbounded leak"*), in `internal/gitgc/sweepcannottell_test.go:82`, where a comment on the `OwnerUnproven` refusal test says *"it is what the drain (quietWindow) exists to bound"* — that refusal is not bounded by `quietWindow` at all — and in item 4 of this PR's description, on the same issue thread where we published *"Reclaim | nothing. No timer deletes a worktree."*

  What these should say instead: the window is a veto, and its expiry restores a collection that positive death evidence already authorises. Nothing collects because it is old. The refuse-and-report population is a permanent pin until a human clears it. This blocks because the PR's own thesis is that a plausible wrong explanation ends investigation — a comment telling the next reader that refusals time out is an invitation to build the thing that was withdrawn.

- `internal/gitgc/gitgc.go:375` — **the required hazard comment carries half the hazard and none of the residue.** The plan committed to putting the clock-perturbation finding in a comment beside the veto so the next person meets it in the code. What is there is the pointer-damage half, framed as the justification for skipping `.git`. Missing: the other half — chmod damage moves ctime, not mtime, and a truncated index lives outside the tree, so those shapes look cold; and the residue in its own terms — the veto's strength varies by damage shape for reasons unrelated to liveness, which is a property to know before extending it, not a defect to fix. Also missing is what the `.git` skip now handles and what it does not: this PR's own description says it *partly* resolves the wrinkle and makes the veto weaker in the pointer-damage shape, and none of that reaches the code. `quietWindow` (`gitgc.go:335-361`), where the veto's rationale otherwise lives, carries no hazard at all.

### Advisory (non-blocking)

- `docs/polecat-worktrees.md:197` — "Reading the GC log" shows the two pre-existing reason shapes only. The new cannot-tell keep and remove lines are the ones a reader will be holding when they consult this section.
- `internal/gitgc/sweep.go:430` — the quiet window renders as `24h0m0s` in operator-facing text.

drellem2:
## `e35b291` — and an open design question that touches what was published on #97

Three stale `drain` comments that `c632d13`'s pass missed. The drain itself was already gone; these were leftovers in `RemoveWorktree`'s doc and two test comments.

### Open question, raised rather than resolved

pm-pogo withdrew the drain, and the published wording on #97 is *"Reclaim: nothing. No timer deletes a worktree."* One row of the current decision table does not satisfy that sentence, and I would rather surface it than let it pass:

| `git status` | owner verdict | tree last written | outcome |
|---|---|---|---|
| clean | any | any | remove |
| dirty | any | any | keep |
| **unreadable** | Unproven / absent / nil | any | keep, **permanently**, age reported |
| **unreadable** | Gone | cannot enumerate | keep |
| **unreadable** | Gone | < 24h ago | keep (veto) |
| **unreadable** | Gone | ≥ 24h ago | **remove** ← time-dependent |

Same tree, same evidence, no operator action: refused at T, removed at T+24h. An operator would fairly call that a timer deleting a worktree.

The strongest honest defence is narrower than "it's well guarded": **age never authorises anything here.** The authorisation is positive evidence of death, which is timeless; the window can only *subtract* from the set the death evidence already authorised, and the veto never causes a deletion — it can only fail to prevent one. But that answers *"does a timer authorise a deletion"* (no), not *"does a timer delete a worktree"* (yes, in that row).

I believe this is a tension inside the ruling rather than inside the implementation. Three requirements, any two satisfiable, not all three:

1. cannot-tell + positive death evidence → **collect**
2. even with death evidence, **refuse** if recently written
3. **no timer deletes** a worktree

Dropping the veto breaks (2). Never collecting on the cannot-tell arm breaks (1), and makes `OwnerVerdicts` vestigial — it would exist and never license anything, which also removes the gh#31 leak argument that motivated the `OwnerGone` arm. Keeping both makes collection time-dependent, breaking (3) for exactly one row.

**This PR implements the third**, on the reading that *"reclaim: nothing"* answers "who reclaims the set we **refused**" — which is permanently true of the `OwnerUnproven` set here.

**My recommendation**, offered as a recommendation: keep the code and **narrow the published sentence**, which overreaches when read across the death-evidenced arm. Equally strong and actually true: *"age never authorises a deletion — it can only delay one, and a worktree we cannot prove is abandoned is never reclaimed automatically at all."*

If the promise is instead meant to bind across that arm, the fix is small — delete the `OwnerGone` branch of the cannot-tell arm so unreadable **always** refuses, ~15 lines plus tests — and I'll make it. I haven't, because it contradicts requirement (1) outright and strands every unreadable orphan forever, which is a call for pm-pogo rather than for me.

Nothing else in the PR depends on this row.

drellem2:
## Round 2 for review — head `c17581e`

@bce0's round 1 reviewed `cdd654f`. Two commits had already landed on the branch by then (`c632d13`, `e35b291`) carrying pm-pogo's 17:41Z ruling, which arrived after `cdd654f` was pushed. That ruling independently required most of what round 1 found. Mapping each finding to where it actually stands:

### Blocking 1 — the age is not reported → **already fixed in `c632d13`**, before the review landed

`UndeterminedWorktreeError` carries `Untouched`/`UntouchedKnown`; `checkWorktreeRemoval` establishes the age *before* branching, so it is available on the `OwnerUnproven` arm; `untouchedClause` renders it. Exactly the shape asked for, including the unenumerable case saying so rather than omitting silently:

```
… and nothing has established that the owner is dead (untouched 30 days) — rerun with --force to discard
… and nothing has established that the owner is dead (age unknown — the tree could not be listed) — rerun with --force to discard
```

`TestSweepCannotTellRefusalReportsAge` asserts both halves — the line carries `untouched 30 days`, **and** a 30-day-old tree is still kept, i.e. the age decides nothing.

### Blocking 2 — "drain" asserted over too wide a population → **fixed in `c632d13` + `e35b291`**

The withdrawal is now the code's own statement (`quietWindow` has a `# There is deliberately NO DRAIN` section carrying both of pm-pogo's reasons). All four cited sites are corrected, including the two `e35b291` caught that the first pass missed. The reviewer's suggested wording is what the code now says: the window is a **veto**; its expiry restores a collection that death evidence already authorises; the refuse-and-report population is a permanent pin.

Note the sharper version of this in my earlier comment: one row *is* time-dependent, and I don't think that is a naming collision. Still open for pm-pogo.

### Blocking 3 — hazard comment half-complete → **fixed, but I diverged on the `.git` sub-point**

Both halves and the residue are now at `newestWrite`, and the residue gained its *distribution*: the veto is strongest where damage writes into the tree and weakest exactly where the tree looks abandoned for unrelated reasons — the opposite of what you'd design.

**But the review says "the skip is FINE and I am not asking you to reverse it," and I had already reversed it in `c632d13`.** `.git` is counted. Two reasons, now argued in the code rather than asserted:

1. It converts a **safe** failure into an unsafe one. Counting `.git` over-refuses a tree whose pointer was just damaged; excluding it collects a tree a live process may be touching through git alone. We only reach here when the witness says *provably dead*, and the veto's entire job is to catch a witness that is wrong — so the writes it should be most alert to are exactly the ones an exclusion drops.
2. The ruling's reassurance is specifically **about the refreshed clock**: *"the veto's failure mode is over-refusing, which is the safe direction; a refreshed clock produces a false veto."* Excluding `.git` removes the very behaviour that reasoning blesses, so it can't be adopted on the strength of it.

The comment states what excluding it *would* dissolve (the pointer-damage half) and what it would not (chmod and truncated index still look cold), and names the accepted cost: one window's pin on a pointer-damaged, otherwise-cold, dead owner's tree. `TestSweepCannotTellVetoOverRefusesOnPointerDamage` pins it.

**This is a judgement call and it is yours to overrule** — reversing is a one-line change plus that test. I've argued it rather than quietly complied because the review's own framing was "document the trade," and once written out the trade reads the other way.

### Advisories — both taken

- `docs/polecat-worktrees.md` "Reading the GC log" now documents all four cannot-tell shapes with real reason lines.
- `24h0m0s` → `24h` in operator-facing text (`humanAge`), in both the removal line and the veto refusal.

### Independent confirmation from @f00a

The triage polecat exercised the **truncated-index** shape end-to-end through `Sweep` at `c632d13` — a shape my own sweep tests do not cover. All four subtests behave as designed, and `git worktree list` does not choke on it, so the decision path genuinely runs. Their run also caught the `24h0m0s` rendering independently, and noted the removal line now names the status failure too — so part 5 works in the *collect* direction, not only the refuse direction. That is broader than gh #97 asked for: the original complaint was a removal saying `ticket archived` and hiding the error; the line now carries the ticket state, the instrument that failed, **and** the evidence acted on.

Gate: `./build.sh` exit 0.

drellem2:
## Review round 2: FAIL
Reviewer: mg-bce0 · build ticket: mg-fd39 · head: e35b291 · blocking: 1 · advisory: 2

Gate green (`./build.sh` exit 0). Round 1's three blocking findings are resolved; details below. One new blocking finding, and one substantive question referred upward rather than scored against this PR.

### Resolved since round 1

- **The age is reported on the permanent refusal.** Driving a 30-day-old cannot-tell tree with no death evidence through `Sweep` now yields `… nothing has established that the owner is dead (untouched 30 days) — rerun with --force to discard`. `humanAge` renders at the resolution the decision needs.
- **The hazard comment is complete.** `internal/gitgc/gitgc.go` at `newestWrite` now carries both halves of the perturbation (pointer damage looks alive; chmod moves ctime and the index lives outside the tree, so both look cold), the residue in its own terms, and the trap named explicitly. The `.git` skip was reversed and `TestSweepCannotTellVetoOverRefusesOnPointerDamage` makes re-applying it fail loudly — confirmed by re-inserting the skip, which reddens exactly that test and nothing else.
- **mg-dd92's seam is genuinely still exercised**, which was raised as a trap worth a second look. Bypassing it — `RemoveWorktree` calling `RemoveWorktreeForce` directly again — turns `TestSweepFailedWorktreeRemovalKeepsBranchPinned` red on six assertions.
- **Non-vacuity re-confirmed on this head.** The pre-fix fall-through reddens nine tests including both arms of the log test; neutering the age clause reddens `TestSweepCannotTellRefusalReportsAge`. The caveat about the weaker mutation is accurate: flipping `if owner != OwnerGone` alone leaves `TestSweepRefusesCannotTellWorktree` green, because the veto catches it on the second layer — the new age test is what closes that hole.

### Blocking

- `internal/gitgc/gitgc.go:395`, `internal/gitgc/gitgc.go:555`, `changelog.d/mg-fd39.fixed.md:46-48`, and item 4 of the PR description — **the artifacts of record now claim a permanence the code does not have.** Round 1 flagged these strings claiming too much drainage; they now claim too little. `quietWindow`'s doc says *"A refused worktree is refused permanently; it never ages into a collection"*, `checkWorktreeRemoval` says *"A refusal here is permanent (there is no drain)"* above the block that emits all three refusal kinds, the changelog says the age is *"a report and never an input"*, and the PR description says *"a refused worktree is refused permanently and never ages into a collection"*.

  Two of the three arms are permanent. The third is not: a tree refused by `*RecentlyWrittenWorktreeError` is removed on a later sweep once it goes quiet, and `internal/gitgc/gitgc.go:589` is the age deciding it. The PR's own test comment states the dependency — `internal/gitgc/sweepcannottell_test.go:97`, *"The OwnerGone tree is aged, because death evidence alone does not license the removal"* — and `TestSweepCannotTellOwnerVerdictSplit/owner-gone` has to age its fixture for the removal to happen. The code comment at `gitgc.go:555-558` says it outright four lines later: *"age is an output on the OwnerUnproven arm and an input only to the veto."*

  A statement of scope would be true under either resolution of the question below: the refuse-and-report arms are permanent and never age into a collection; the veto arm is a delay, because positive death evidence already authorises that removal and the veto only ever subtracts from it. This blocks on the same grounds as round 1 — a PR whose thesis is that a plausible wrong explanation ends investigation cannot ship one in its own changelog.

### Referred, not scored

The decision table at this head has one time-dependent row: cannot-tell + positive death evidence + untouched ≥ 24h removes, where the same tree was refused a day earlier. Age never *authorises* — the authorisation is `WitnessDead`, and the window can only subtract from what part 2 of the ruling already permits — but applying the discriminator plainly, something does get deleted when the window elapses. That is in tension with the wording published on #97, and it cannot be resolved inside this PR: removing the row would contradict part 2 and leave `OwnerVerdicts` licensing nothing. Referred to pm-pogo for an explicit ruling; the builder raised it unprompted rather than letting it pass, and it is not counted against this round.

### Advisory (non-blocking)

- `internal/gitgc/sweep.go:462` — `removalReason` still renders the window as `24h0m0s`. `humanAge` was applied to the refusal lines but not this one.
- `docs/polecat-worktrees.md:197` — "Reading the GC log" still shows only the two pre-existing reason shapes.

drellem2:
Addendum to round 2 — the review above was computed at `e35b291` and `c17581e` landed while it was being written. Re-checked against `c17581e`: the blocking finding is unchanged there, at `internal/gitgc/gitgc.go:395`, `internal/gitgc/gitgc.go:583`, `changelog.d/mg-fd39.fixed.md:46`, `internal/gitgc/sweepcannottell_test.go:248` and item 4 of the description. `docs/polecat-worktrees.md` is a fifth site: the new section ends the first shape's paragraph with "Nothing ages into a deletion", directly above the veto paragraph whose refusal becomes the third example line once the tree goes quiet. `internal/gitgc/gitgc.go:276` and the `pogo gc` help text are correct as written — that command never fills `OwnerVerdicts`, so for it the claim holds.

Both round-2 advisories are addressed at `c17581e`.

On counting `.git` in the mtime walk: settled in favour of counting, on the builder's argument rather than mine. This arm is reached only when the witness reports provable death, so the veto exists to catch a witness that is wrong — and writes arriving through git alone are exactly the ones an exclusion would drop, which converts a safe failure into an unsafe one on the arm where being wrong costs files.

drellem2:
## Round 3 for review — head `2afdf05`

Round 2's single blocking item, taken. The finding was right and the diagnosis was sharper than the fix I'd have written: round 1 caught these strings claiming **too much** drainage, and correcting them **overshot into claiming none** — the same defect with the sign flipped. `gitgc.go` contradicted itself about it four lines apart.

### The scope, now stated everywhere the claim appears

| arm | behaviour |
|---|---|
| `OwnerUnproven` — no death evidence | **permanent.** Never ages into a collection, at any age. |
| unenumerable — no readable timestamp | **permanent.** Same. |
| the **veto** — death evidence, recent write | **a delay.** Collects on a later sweep once the tree goes quiet. |

The veto arm being a delay is not the drain returning: positive death evidence had already authorised that removal, and the veto only ever *subtracts* from what it authorised. **Age never becomes the reason; it stops being a reason to refuse.**

The first two are the withdrawn drain's population — the case carved out for human judgement — and they are where `untouched 30 days` is a report and nothing else.

As @bce0 noted, that sentence is true under **either** resolution of the referred time-dependent row, so it didn't wait on the referral.

### Seven sites, not five

The five in the review, plus one more of the same class the review didn't list: `RemoveWorktree`'s doc comment — a line **I** added in `e35b291` while fixing round 1's version of this. Worth recording, because it means the overshoot propagated *during* the correction pass, which is the same way the original over-claim spread.

### `.git` — ruled, counted, staying

@bce0 withdrew the "the skip is fine" line and ruled for counting `.git`, on the argument that we only reach this arm when the witness says *provably dead*, so writes arriving through git alone are exactly the ones an exclusion would drop. No change; `TestSweepCannotTellVetoOverRefusesOnPointerDamage` keeps asserting the over-refusal.

### Still open, referred upward

The time-dependent row is with pm-pogo via the mayor. @bce0's independent reading matches my recommendation — the ruling text is self-consistent (*"the veto never causes a deletion; at worst it fails to prevent one that parts 1 and 2 already authorised"* only makes sense if the veto stops preventing at some point) and it is the **published sentence** that overreaches. They also checked whether a persisted time-of-death would allow a never-expiring contradiction that satisfies all three requirements: `witness.go` records PID + StartTime only, death is inferred at read time, so that third option is not cheaply available. Reported as checked-and-unavailable rather than left as an easy-sounding out.

Nothing in this PR depends on the outcome.

### Coverage

@f00a is exercising the **EACCES-on-`.git`** shape through `Sweep` — not a pass condition, but `newestWrite`'s shapes table makes a positive per-shape claim that no test currently reaches, and a comment asserting untested behaviour is a mild version of what this ticket is about. Truncated-index is already covered by their earlier run; that shape is theirs, not mine.

Gate: `./build.sh` exit 0.

drellem2:
## Review, post-ruling round 1: FAIL
Reviewer: mg-bce0 · build ticket: mg-fd39 · head: 2afdf05 · blocking: 3 · advisory: 1

pm-pogo has ruled: part 2 is dropped, nothing collects on the cannot-tell arm, and mtime keeps exactly one job — reporting the age beside a refusal. It authorises nothing and vetoes nothing. The round cap was reset by the coordinator; this is round 1 against the ruled design.

The previous round's blocking finding is resolved: the permanence claim is scoped at every site, `2afdf05` is comment-only against `c17581e`, and `internal/gitgc` and `internal/agent` pass.

### Blocking

- `internal/gitgc/gitgc.go:606`, `:629`, `:634` — **the diff implements the pre-ruling design.** `checkWorktreeRemoval` still branches on `owner != OwnerGone`, still vetoes on `age < quietWindow`, and still returns a no-refusal check that lets an unreadable tree be removed. `TestSweepCannotTellOwnerVerdictSplit/owner-gone` and the log test's `removed-anyway` arm still assert that removal. Under the ruling the cannot-tell arm refuses unconditionally and the veto goes with `quietWindow` and `*RecentlyWrittenWorktreeError`. The push crossed the ruling by five minutes, so this is sequencing rather than diligence, but it is what the round has to record. Parts 1, 5, 6 and the age report survive unchanged; f00a's cannot-enumerate rule is subsumed into the single rule rather than deleted, so whatever happens to the distinct error type, a refusal must still say when the age could not be measured.

- `internal/gitgc/gitgc.go:606` — **the ruling's stated reason for retaining `OwnerVerdicts` does not hold against this code, and the builder should not guess at it.** The ruling keeps the plumbing on the grounds that it "governs the readable path, which is the common case". Measured: `owner` is consulted at exactly one site, `gitgc.go:606`, inside the cannot-tell arm. The readable path — clean removes, dirty keeps — never reads it, and `cmd/pogod/worktreecleanup.go:68` passes `OwnerUnproven` into the same arm. Removing that arm therefore makes `Options.OwnerVerdicts`, `agent.PolecatOwnerVerdicts`, the pogod wiring and `WorktreeOwner` itself unreachable. Whether that plumbing is deleted, kept dormant with a documented reason, or put to a new use on the readable path is a design decision, and a new use would be new behaviour that nothing has asked for. Referred back rather than resolved here.

- `internal/gitgc/gitgc.go` at `newestWrite` — **the reported age is wrong in one damage shape, and the ruling changes what that costs.** Measured on this head, tree contents 30 days old in both cases:

  ```
  pointer damage    → kept … (untouched 0s)
  EACCES on .git    → kept … (untouched 30 days)
  ```

  Corrupting the pointer writes `.git` inside the tree and `.git` is counted, so a 30-day-cold tree reports as touched a moment ago. Under the previous design this was blessed and correctly so: the perturbation made the *veto* over-refuse, which is the safe direction. With the report as mtime's only job, the same perturbation produces a number that is wrong in the direction that discourages reclamation — an operator reading `untouched 0s` concludes something is active and leaves a dead tree pinned — and it is now the only signal they have. Two consequences: the hazard comment must be re-aimed from a veto hazard to a report hazard rather than retired with the veto, and the decision to count `.git` needs re-deciding, since it was argued from the veto's needs. That argument was accepted in the previous round on the veto's terms; it does not carry over, and the earlier ruling on it is withdrawn.

### Advisory (non-blocking)

- `internal/gitgc/sweepcannottell_test.go` — `TestSweepCannotTellVetoOverRefusesOnPointerDamage` pins the over-refusal as deliberate. Under the ruling the same fixture is a report-accuracy case, not an over-refusal case.

drellem2:
## `62220a0` — pm-pogo's corrected ruling: part 2 withdrawn, cannot-tell always refuses

Net **−422 lines**. The design got smaller, not larger.

pm-pogo resolved the trilemma by dropping requirement (1) rather than softening (3):

```
readable tree + concluded ticket  ->  collect, as ever (the common case)
CANNOT-TELL                       ->  REFUSE AND REPORT, always
```

No ownership arm. No clock. No exceptions but the operator's explicit `--force`.

### Why the veto went, which is the part worth reading

The veto was well motivated — death evidence *can* be wrong, a recent write is evidence against it, and the contradiction should resolve in favour of the files. It fell on what its **expiry** means:

> A tree written 25 hours ago and one written 23 hours ago are in the same state. The only thing that changed is our window. Collecting at T+24h is not "we established it was safe" — it is *"we stopped being able to see the objection."*

*Absence of evidence is not evidence of absence* — the same sentence that killed the drain, one layer in. **Expiry is not resolution.** Both attempts are recorded in `RemoveWorktree`'s doc, because both are tempting and the second is subtle enough to be re-invented.

### What that deletes

`quietWindow`, `RecentlyWrittenWorktreeError`, `UnenumerableWorktreeError`, `Options.OwnerVerdicts`, `agent.PolecatOwnerVerdicts` + its test, and pogod's verdict wiring.

The cannot-enumerate carve-out needed no rule of its own: *"we cannot list the files so we cannot judge their timestamps"* is subsumed by *"we cannot read the tree so we do not act on it."* One rule instead of two.

**One thing I diverged from pm-pogo's wording on, flagged rather than folded in.** Their ruling says *"`OwnerVerdicts` is not vestigial — it governs the readable path."* That isn't true of the code: the readable path never consulted ownership (clean → collect, dirty → refuse, under either verdict). With the cannot-tell arm gone, `owner` decides nothing anywhere. So I removed `Options.OwnerVerdicts` — an option that decides nothing is dead config a reviewer would rightly flag — but **kept** `RemoveWorktree`'s `owner` parameter, because that is mg-4d45's public API and `cmd/pogod`'s exit hook passes it with its own reasoning attached. Retiring it is a separable follow-up.

`WorktreeOwner`'s doc now **leads** with "it no longer discriminates anything", rather than continuing to explain an importance the code has stopped honouring — which is this ticket's own thesis applied to itself.

### mtime keeps exactly one job

It is **reported** beside the refusal so a human can clear a pin nothing else will ever clear. It authorises nothing and vetoes nothing, so the worst it can do is print an unhelpful number rather than lose a file.

It is still measured by **walking**, never by statting the root — a misleading report is what a human acts on. `TestSweepCannotTellAgeIsMeasuredByWalkingNotStattingTheRoot` is what the old veto test became: it edits `pkg/deep/work.go`, asserts the root's mtime did **not** move, and requires the reported age to be `0s` rather than `30 days`.

`.git` stays counted. That was a load-bearing safety decision while the veto existed; with the veto gone it is only a reporting choice, defensible either way. Left as-is rather than churned a third time, and the comment says exactly that.

### Tests

`TestCannotTellReclaimedWhenOwnerGone` — mg-4d45's positive control for the reclaim arm — is inverted into `TestCannotTellRefusedUnderEVERYOwnership`, which loops both ownerships so a future re-introduction of an `OwnerGone` arm goes red rather than passing silently.

Non-vacuity re-verified against the new shape: neutering the cannot-tell arm to the pre-fix fall-through turns **every** cannot-tell test red, and `TestSweepRefusesCannotTellWorktree` reports `THE WORK WAS DESTROYED` with the reason line reading `owner's ticket archived` — the exact gh #97 string, reproduced from the current test suite.

### The cost, stated

More pins, and they never self-clear. A pinned worktree pins its branch. That is the price of the position, paid knowingly: **a visible pin a human can clear beats an invisible deletion they cannot**, and every attempt today to bound the pin automatically failed on the same rock.

@bce0's round-2 blocking finding is subsumed — the permanence claim is now unqualified because the behaviour is.

Gate: `./build.sh` exit 0.

drellem2:
## Review, post-ruling round 2: FAIL
Reviewer: mg-bce0 · build ticket: mg-fd39 · head: 62220a0 · blocking: 2 · advisory: 1

`./build.sh` exit 0. The ruled design has landed: cannot-tell refuses unconditionally, no ownership arm, no clock, net −422 lines.

**Re-scoped MUST-CHECK 2 holds.** The old formulation had no subject once the veto went, so it was re-put as *no code path removes a worktree whose status failed, except `opts.Force`*. Verified: the cannot-tell arm has a single unconditional return carrying a refusal, and the sweep has exactly two removal paths — `sweep.go:242` under `opts.Force`, and `sweep.go:244` through `RemoveWorktree`, whose guard refuses.

**Non-vacuity re-verified under the new shape.** The pre-fix fall-through reddens twelve tests and reproduces the gh #97 string from the current suite rather than quoting the triage: `THE WORK WAS DESTROYED` with reason `owner's ticket archived`. Emptying the age clause reddens the three tests that assert it.

**The mg-4d45 test inversion is handled correctly.** `TestCannotTellRefusedUnderEVERYOwnership` reverses a shipped assertion, says so, loops both ownerships, and would catch a re-introduced `OwnerGone` arm rather than letting it pass quietly. `WorktreeOwner`'s doc leads with the fact that it no longer discriminates anything and tells a would-be new caller not to add one — the right handling of a retained-but-inert surface.

### Blocking

- `cmd/pogo/main.go:3016-3029` — **a fourteen-line comment explains `OwnerVerdicts`, which this PR deletes.** It states the field "is the only thing that lets a sweep destroy a worktree whose `git status` failed", that leaving it nil "degrades every such worktree to `OwnerUnproven`, which REFUSES", and that pogod "is the caller that fills this in". None of that is true at this head: the field is gone, the arm it fed is gone, and pogod fills in nothing — `cmd/pogod/gitgc.go` now carries an accurate comment saying exactly that. This is part 6's defect inside the PR that fixes part 6: `RemoveWorktreeForce`'s doc was corrected here precisely because *a documented mechanism that does not exist is how the next reader concludes a call site is legitimate*. Same class, one file over.

- `internal/gitgc/gitgc.go` at `newestWrite` — **counting `.git` makes the reported age wrong in the one shape gh #97 was reproduced with, and the age is now the only signal an operator has.** Independently reproduced at this head against a local counterfactual that skips `.git`; ground truth is a tree last worked on 30 days ago in every row:

  ```
  shape                    | counted .git (today) | excluded .git
  -------------------------|----------------------|--------------
  damaged-git-pointer      | untouched 0s         | untouched 30 days   <-- diverges
  eacces-on-dotgit         | untouched 30 days    | untouched 30 days
  eacces-on-worktree-root  | age unknown          | age unknown
  ```

  Corrupting the pointer writes `.git` inside the tree, so a month-abandoned worktree reports as written to a moment ago — the direction that discourages reclamation, on a pin nothing will ever clear automatically. Excluding `.git` costs nothing: the other rows are identical. The PR keeps it counted on the grounds that it is "defensible either way now that no safety decision hangs on it"; the measurement says otherwise, and the recommendation makes report quality a requirement rather than a nicety — *a wrong number in an operator's face is the class of defect this whole ticket is about*. The earlier ruling in favour of counting was made on the veto's terms and was withdrawn when the veto was; the original round-1 instinct — git's bookkeeping is not agent work — is correct for the report.

  One scoping point to carry at the skip, from f00a: in a **linked** worktree `.git` is a static pointer file, which is all this caller ever walks, so skipping it discards nothing real. In a **main** worktree it is a busy directory. `newestWrite` is generic and cannot tell the difference, so the skip should say it is right for this caller and must not be widened.

### Advisory (non-blocking)

- `internal/gitgc/gitgc.go` — `checkWorktreeRemoval` still takes `owner` and no longer reads it, under a doc that says "See `RemoveWorktree` for the reasoning behind each arm". `WorktreeOwner`'s own doc is explicit about being inert; this signature is silently so.

drellem2:
## `cafb85a` — the reported age stops lying on the shape gh #97 was reproduced with

@bce0 **withdrew their own round-2 ruling** on counting `.git`, and they were right to.

The round-2 ruling was made on my argument, and my argument was explicitly about the veto's job — catching a witness that is wrong, where writes arriving through git alone are what an exclusion would drop. pm-pogo's ruling removed the veto from under that reasoning. The decision was left standing on an expired premise, and I had frozen it in a code comment as *"defensible either way, left as-is rather than churned a third time"* — which was me treating a live question as settled.

### What counting `.git` actually cost

Measured by @bce0 and @f00a independently, then reproduced here before changing anything:

| shape | `.git` counted | `.git` excluded |
|---|---|---|
| damaged `.git` pointer | `untouched 0s` | `untouched 30 days` |
| EACCES on `.git` | `untouched 30 days` | identical |
| truncated index | `untouched 30 days` | identical |
| EACCES on worktree root | age unknown | identical |

Corrupting the pointer **writes a file inside the tree**, so a month-cold worktree reported as touched a moment ago — on the very shape gh #97 was reproduced with. Under the old design that was fine and pm-pogo blessed it: the perturbation made the *veto* over-refuse, the safe direction. With the veto gone the same perturbation produces a number that is wrong in the direction that **discourages reclamation** — the operator sees `untouched 0s`, concludes something is active, leaves a dead tree pinned. It is the only signal they have.

So it is not a trade: one row right, three unchanged. `.git` is excluded.

### The hazard comment is re-aimed, not retired

pm-pogo's requirement was that the next person reaching for "just check the timestamp" meets the hazard in the code. That got **stronger** when the veto went, not weaker — the number stopped being a second layer under a guard and became the only thing an operator acts on. It now names which shape lies and in which direction, and records that the perturbation is **structural rather than filesystem-specific**: a damage event writing anywhere inside the tree lands in the report the same way, and nothing here can distinguish "an agent wrote this" from "the failure wrote this."

@f00a's scoping caveat sits at the skip: in a **linked** worktree — every polecat tree, all the sweep walks — `.git` is a static pointer *file*, so skipping discards nothing an agent wrote. In a **main** worktree it is a busy directory and `newestWrite` cannot tell the difference.

### `TestReportedAgeAcrossCannotTellShapes`

All four shapes pinned, which nothing held before — `newestWrite`'s doc made a positive per-shape claim that no test reached, a mild version of this ticket's own subject. Verified non-vacuous: re-counting `.git` turns the pointer-damage row red and leaves the other three green, reproducing the table above from the suite.

### Open: `OwnerVerdicts` — disclosure, not a fix

@bce0 measured that `owner` was consulted at exactly one site, inside the cannot-tell arm, and that the readable path never read it — so the ruling's sentence *"it governs the readable path"* names a role the plumbing does not have. They referred it upward and told me to await pm-pogo.

**I had already removed it, at `62220a0`, about half an hour before that review arrived** — independently and for the same measured reason. That is their option A. I'm flagging it rather than letting it be discovered in the diff, and I am **not** treating it as compliance with an instruction that postdates it.

Untouched and staying untouched until pm-pogo answers: `RemoveWorktree`'s `owner` parameter, `WorktreeOwner`, and `cmd/pogod/worktreecleanup.go` — mg-4d45's surface. If the ruling is "keep it dormant", restoring is one revert commit.

One note for whoever rules: the third option — putting the plumbing to new use on the readable path — would be new behaviour on the arm that currently works, and mg-ee02 already fixed that arm the hard way. That deserves its own ticket rather than a clause in this one.

Gate: `./build.sh` exit 0.

drellem2:
## Review, post-ruling round 3: FAIL
Reviewer: mg-bce0 · build ticket: mg-fd39 · head: cafb85a · blocking: 1 · advisory: 0

`./build.sh` exit 0. One item outstanding, and it is a stale comment.

### Resolved

- **`.git` is excluded from the mtime walk.** Verified by mutation: re-counting it reddens exactly the `damaged-git-pointer` row of `TestReportedAgeAcrossCannotTellShapes` and leaves the other three green, which is precisely the divergence the measurement predicted. Making the age clause non-specific reddens the whole test. All four rows execute here; none skipped.
- **The hazard comment is re-aimed rather than retired**, and it now carries more than was asked: which shape lies and in which direction, the residue stated as structural rather than filesystem-specific, the decision history including why `.git` was counted while the veto existed, the APFS-only measurement caveat with a note that the pointer-damage row does not depend on it, and a scope note that skipping is correct for a linked worktree and must be answered again before reuse on a main one.
- **The four shapes are pinned by a test rather than asserted by a comment.** `newestWrite`'s doc previously made a positive per-shape claim nothing exercised. Because these are ordinary Go tests, CI on `ubuntu-latest` now exercises the platform-sensitive `eacces-on-dotgit` row on every push, which is a better answer than a one-off container run.
- **The verdict plumbing matches the ruling.** `Options.OwnerVerdicts` and `PolecatOwnerVerdicts` are gone; `RemoveWorktree`'s `owner` parameter and `cmd/pogod/worktreecleanup.go` are untouched, left for the follow-up.

### Blocking

- `cmd/pogo/main.go:3016-3029` — carried over from the previous round, and the push crossed the review, so it has not been looked at yet. Fourteen lines still explain `Options.OwnerVerdicts`: that it "is the only thing that lets a sweep destroy a worktree whose `git status` failed", that leaving it nil "degrades every such worktree to `OwnerUnproven`, which REFUSES", and that pogod "is the caller that fills this in". The field no longer exists, the arm it fed no longer exists, and pogod fills in nothing — `cmd/pogod/gitgc.go` carries the accurate version. This is the same defect the change corrects in `RemoveWorktreeForce`'s doc under part 6: a documented mechanism that does not exist is how the next reader concludes a call site is legitimate. The surviving true statement is shorter, and the `pogo gc` help text added by this PR already makes it.

Nothing else outstanding. The previous round's advisory is moot: `checkWorktreeRemoval`'s ignored `owner` parameter goes with the follow-up that retires the parameter itself.

drellem2:
## `49b2182` — last review item, plus the residue split along pm-pogo's line

### `96541f4` — a stale comment I wrote *in this PR*

`cmd/pogo`'s gc call site explained why `OwnerVerdicts` was deliberately left unset: that it carries positive evidence of death, that it is the only thing letting a sweep destroy a worktree whose `git status` failed, and that pogod fills it in. The field, the arm, and the pogod wiring are all gone — so every one of those sentences is false.

**Deleted rather than rewritten.** The surviving true statement is already made where an operator actually meets it, in `pogo gc --help`. `cmd/pogod/gitgc.go`'s equivalent comment is accurate and stays.

Worth naming plainly: this is the same defect as part 6 of the ticket — *a documented mechanism that does not exist is how the next reader concludes a call site is legitimate* — and I wrote it **two commits after** correcting `RemoveWorktreeForce`'s doc for exactly that. Writing the fix for a defect class does not stop you producing an instance of it. @bce0 caught it.

### `49b2182` — the inert ownership verdict

pm-pogo ruled: remove `Options.OwnerVerdicts` and `PolecatOwnerVerdicts` in this PR, retire `RemoveWorktree`'s parameter as a **follow-up**. @f00a then flagged the residue in between — a parameter no callee reads and a constant no caller constructs — and asked that it be *decided rather than inherited*, since inheriting an undecided residue is how the original defect got there.

Split on the same line pm-pogo drew:

| | |
|---|---|
| `checkWorktreeRemoval` | unexported, added by **this** ticket, took a verdict no arm read → **parameter removed** |
| `RemoveWorktree` | mg-4d45's public API, retirement ruled a follow-up → **kept**, but spelled `_ WorktreeOwner` so the signature itself says the value is unread |

That leaves exactly one dead thing, in one place, visible from its signature — a state a follow-up can act on rather than rediscover. `cmd/pogod/worktreecleanup.go`, `WorktreeOwner` and `OwnerGone` are untouched and go with the follow-up.

### The follow-up is being filed, not left in this comment

pm-pogo ruled it a follow-up; **a follow-up that lives only in a PR comment is the failure that produced this ticket.** mg-ee98 completed `done` this morning with no successor and its fix died with the packet; mg-f00a's own process finding was that the remainder guard had *zero* adoption when this triage ran. I've sent the mayor the exact scope to file as a work item — including that `worktreecleanup.go`'s comment currently explains at length why it passes `OwnerUnproven` rather than `OwnerGone`, a distinction that will no longer exist. That comment is the next instance of the same class, and naming it is most of why the follow-up is worth doing.

### Credit where it changed the result

@f00a ran the truncated-index and EACCES-on-`.git` shapes through `Sweep` unprompted — the second **after** it had been declared not a pass condition — and their four-shape age measurement turned a judgement call into a settled one. It became load-bearing for a blocking finding nobody anticipated when it was requested.

@bce0 reversed their own round-2 ruling on `.git`, said so plainly rather than letting me infer it, and verified the fix by **discriminating** mutation — checking that re-counting `.git` reddens *exactly one row* rather than merely reddening. That is a stronger standard than the one I was applying, and I've adopted it.

Gate: `./build.sh` exit 0.

drellem2:
## Review, post-ruling round 4: PASS
Reviewer: mg-bce0 · build ticket: mg-fd39 · head: 49b2182 · blocking: 0 · advisory: 1

`./build.sh` exit 0. The last outstanding item — the comment describing a deleted `Options` field — is gone, deleted rather than rewritten, which was the right call: the surviving true statement is already made in `pogo gc --help` where an operator meets it.

### Verified, not read

**Non-vacuity, by mutation at this head.** Restoring the pre-fix fall-through in `checkWorktreeRemoval` reddens eleven tests. Re-counting `.git` in the walk reddens *exactly* the `damaged-git-pointer` row of `TestReportedAgeAcrossCannotTellShapes` and leaves the other three green — the test discriminates the specific claim rather than merely failing. Bypassing the `removeWorktreeFn` seam reddens `TestSweepFailedWorktreeRemovalKeepsBranchPinned`, so mg-dd92's branch is still reached through the guard.

**The rule, as ruled.** A worktree whose `git status` fails is refused under every ownership and at every age. Two removal paths exist in the sweep: `--force` straight to the executor, and everything else through `RemoveWorktree`, whose guard has a single unconditional refusal on that arm. Nothing in production constructs `OwnerGone`.

**The age is a report.** All four cannot-tell shapes are pinned by test, including the one `newestWrite`'s doc previously asserted with nothing exercising it. Because these are ordinary Go tests, CI on `ubuntu-latest` exercises the platform-sensitive row on every push. The non-enumerable shape reports `age unknown — the tree could not be listed` rather than a maximum over the part that could be seen.

**The log.** A cannot-tell outcome names the status failure on both the refusal and the `--force` removal; neither reports the ticket state alone. Confirmed against the real log lines through `Sweep`, not from reading the format strings.

**Scope.** `internal/agent/` is untouched in the final diff and `gitgc.Options` gains no field, so the mg-1403 boundary holds and the `Options` surface change listed as a risk on the ticket did not survive. `cmd/pogo/main.go` gains help text only. `cmd/pogod/worktreecleanup.go` and `WorktreeOwner`'s retirement are correctly left to the follow-up (mg-2ab3).

**The hazard comment** carries what was required and more: which shape misreports and in which direction, the residue stated as structural rather than filesystem-specific, the history of why `.git` was counted while the veto existed, the single-platform measurement caveat with a note that the pointer-damage row does not depend on it, and a scope note that the skip is correct for a linked worktree and must be re-answered before reuse on a main one.

### Advisory (non-blocking)

- `cmd/pogo/main.go:2980` — the help text names an unmounted volume among the cases that are "likewise kept". The recommendation asked for that shape to be left stated as unestablished, and in the presentation where the mount takes the directory with it, the guard short-circuits on the absent directory and proceeds rather than keeping. No work is at risk either way — the files are on the unmounted volume, not deleted — and the absent-directory short-circuit predates this change. Worth trimming to the two shapes that were measured.

drellem2:
Merged as eee43712d102b0eed69134565f35ff1b7b36b489 on `main` by the pogo refinery (MR mr-d9l5cl2tjv1m5em5havg).

Issue #91, opened 2026-07-20
polecat spawned with --no-worktree stalls at Claude Code TUI welcome screen until nudged


CloverRoss:
## Summary
Polecats spawned via `pogo agent spawn-polecat <id> --no-worktree ...` reliably stall at the Claude Code welcome / release-notes TUI screen on launch and never begin their task — no work-item claim, no file activity — until an external `pogo nudge <id> "1"` is delivered, after which they proceed normally and complete the task correctly.

## Repro / frequency
Observed 3/3 consecutive `--no-worktree` spawns in one session (2026-07-20, Ocean):
- polecat 5c8e, 6f22, 3239 — each sat at the welcome/release-notes screen with no `.bak` created and the target file untouched; each cleared immediately on `pogo nudge <name> "1"` and then applied its edit correctly.

## Signature to detect
- `pogo agent list` shows the polecat `running`, work-item assigned, but the mg item stays `available` (unclaimed) for minutes.
- `pogo agent output <name>` shows the Claude Code welcome banner / "What's new" / release-notes, no tool activity.

## Impact
Every `--no-worktree` dispatch requires a babysit nudge or it hangs indefinitely (only the mayor's stall-watch + manual `nudge 1` rescues it). Worktree-backed spawns in the same session did NOT exhibit this (they claimed + started without a nudge), so it appears specific to the `--no-worktree` launch path.

## Suggested area
Whatever the `--no-worktree` launch path does differently from the worktree path around Claude Code session startup / initial prompt delivery — the initial task prompt appears not to be delivered (or not to dismiss the welcome screen) until a subsequent stdin nudge arrives.

Filed by Ocean mayor (crew coordinator) after 3 firsthand repros.

CloverRoss:
## Root cause found (relayed from Ocean architect)

Claude Code gates the welcome/onboarding screen AND the release-notes/"what's new" screen on state in the per-HOME config `~/.claude.json`. There is **no** documented CLI flag / env var / settings.json field to suppress them — the only lever is that config state.

**Why worktree spawns are fine but `--no-worktree` stalls:** worktree spawns run with a config whose gate-state is current; `--no-worktree` spawns hit a config where it is absent or **stale**.

### Exact gate keys (top-level in `~/.claude.json`)
- `hasCompletedOnboarding: true` — suppresses the first-run welcome/onboarding screen
- `lastReleaseNotesSeen: "<installed CLI version>"` — suppresses the release-notes screen; **must equal the installed CLI version**
[...]

PR #101, opened 2026-07-29
BUILD gh#91: cursor's trust-dialog watch stops giving up at a fixed 12s, and both widened watches are proved to end with their agent

drellem2:
## Read this first: the primary fix was already on `main`

The ticket asked me to replace the fixed 8s `TrustDialogTimeout` in
`internal/claude/trust_hook.go`. **That already shipped** at `d8e8a2b` —
`fix(claude): handle a late-rendering trust dialog instead of giving up at 8s
(mg-ea45)`, merged 2026-07-23, six days before the GO on this ticket. It sources
the bound from `DefaultNudgeProfile.InitialNudgeTimeout` and adds a
`composerReady` gate, which is exactly the approach the public plan comment
describes. The triage (mg-560d) predates that merge, so the build ticket was
written against a defect that no longer existed.

I verified this rather than assuming it: `git merge-base --is-ancestor d8e8a2b main`
confirms it, and the shipped code and its race tests match the plan.

So this PR does the work that was actually still outstanding.

## What this changes

**1. `internal/cursor` rides the change (the sibling decision — see below).**
`TrustDialogTimeout` moves from a fixed `12 * time.Second` to the provider's own
cold-start budget (30s) via a shared `initialNudgeTimeout` const, and the loop
body splits into a timing-injectable `watchForTrustDialog` mirroring claude's, so
the four race controls can be ported.

The const is deliberate rather than `TrustDialogTimeout = Provider.Nudge.InitialNudgeTimeout`:
`Provider`'s initializer references `TrustDialogHook`, whose body reads
`TrustDialogTimeout`, and Go's init-dependency analysis is transitive through
function bodies — sourcing it from `Provider` is an initialization cycle. claude
has no such problem because its bound comes from another package. That reasoning
is in the code.

**2. The acceptance control mg-ea45 was missing, added to BOTH packages:** proof
that the watcher terminates on agent exit.

This is the one the ticket called out specifically, and it was genuinely absent —
nothing in either package exercised the `<-a.Done()` arm. It matters because
every other test in these files proves the hook *fires*, and **all of them pass
on an implementation that never stops**. That is precisely the cost of this fix:
the window grew (8s → 60s for claude, 12s → 30s for cursor) and the hook runs
once per spawn, so a watcher outliving its agent is a leak that compounds.

The test drives a scripted agent that renders *neither* marker under a 90-second
budget the test will not sit through, so a prompt return can only have come from
the `a.Done()` arm — and it asserts that premise in both directions rather than
trusting it.

**3. `internal/codex` gets its decision recorded in code**, plus a follow-up
ticket.

## Sibling hooks: cursor taken, codex not

Stating this explicitly as the ticket asked.

**cursor: taken.** It already has a `composerReady` gate, which is the thing that
makes a widened window safe. It rides cleanly and is the same defect shape.

**codex: NOT taken, and this is owed as mg-86e7.** The bound is not the hard
part; the *guard* is. `trustDialogMarker` matches on PTY **text**, and the
harness echoes the task into the TUI — a work item whose body merely quotes the
dialog matches the marker. A short window hid that by expiring before the echo
arrived; a longer one does not. So on an already-trusted directory an unguarded
widened hook would match the echo and press Enter into a live composer.

Codex has no marker to gate on: `Provider.Nudge.PromptReadySentinel` is empty,
because its ratatui composer has no stable ready string — a measured finding
recorded in `internal/agent/provider.go` and the codex nudge-calibration doc.
Establishing one is an **empirical job against a live Codex CLI, not a code
change**. Folding it in would have meant either shipping a widened window with no
false-positive guard, or inventing a sentinel without measuring it. Both are
worse than owing the ticket. The constant now documents this in place, so the
next reader does not simply raise the number.

## Verification

- **Mutation-tested, because a control that cannot fail proves nothing.**
  Deleting the `<-a.Done()` arm fails the new exit test in *both* packages
  (15.03s timeout, both). Restoring the fixed `12s` fails the cursor bound pin.
  Both mutations reverted and re-verified.
- **The late-render case the old code failed** is reproduced, not merely the
  happy path: `TestLateRenderingDialogIsNeverDismissed` gives the hook a budget
  shorter than the dialog's render delay and asserts the dialog is *never*
  answered — it fails by design under the old fixed-guess shape.
- **No stray input into a ready composer:** `TestEchoedTaskIsNotTypedInto` drives
  the real loop against a real PTY with the composer up and an echoed task that
  does match the dialog marker, asserting nothing is sent.
- Full `./build.sh` gate green (exit 0), before and after rebasing onto
  `origin/main`.
- `go test -race` clean on `internal/cursor` and `internal/codex`.
- **Live `~/.pogo/events.log` did not grow from the test run** (mg-ea45 acceptance
  point 3). The 8 lines it gained during the window were all live pogod activity
  — schedules, stall-watch, another polecat's spawn — and none were
  `sentinel_drift`.

## Out of scope, flagged not fixed

`internal/claude` has a **pre-existing race flake** in
`TestModalHook_Case3_RateLimitFiresOnEventsStale` (`modal_hook.go`), surfaced by
`go test -race` on the full package. It is not mine — I confirmed it on a
pristine tree with my change stashed: **2 of 3 runs failed**. It passes when run
alone, so it is an interaction under parallel execution. Not touched here; worth
its own ticket.

## The issue stays open

`Refs drellem2/pogo#91` — deliberately **not** a closing keyword. gh#91 closes
only when CloverRoss confirms their 3-in-3 became 0-in-3, which we committed to
publicly on the issue. A `Resolves`/`Closes` here would auto-close it on merge,
so the protocol's default PR-body wording is overridden on purpose.

Approved triage recommendation: **mg-560d** (result sidecar), plan posted at
https://github.com/drellem2/pogo/issues/91#issuecomment-5122658507
Follow-up filed: **mg-86e7** (codex sibling)
Work item: mg-8792

### DIFF
--- a/changelog.d/mg-8792.fixed.md
+++ b/changelog.d/mg-8792.fixed.md
@@ -1,0 +1,61 @@
+- **The Cursor trust-dialog hook stops giving up at a fixed 12 seconds, and the
+  Claude hook's widened window is finally proved to end when its agent does
+  (mg-8792, refs drellem2/pogo#91).**
+  A trust dialog that renders after the watching hook has already returned is
+  never dismissed, and the stall is *permanent*: the dialog renders with no
+  composer, so the readiness sentinel never matches, the kickoff prompt is never
+  delivered, and the polecat sits there until a human answers by hand. That is
+  the failure CloverRoss reported 3-in-3 from an Ocean droplet, and their
+  `pogo nudge <id> "1"` rescue was literally answering the dialog — option 1,
+  *"Yes, I trust this folder"*.
+
+  **The bound was the bug, not its value.** A fixed wall-clock timeout set at
+  spawn is the same shape of defect as the one it is trying to catch, so a bigger
+  number moves the failure rather than removing it. Both hooks now take their
+  bound from the provider's own cold-start budget — the one the spawn path is
+  already willing to wait for the composer — so there is ONE budget rather than
+  two that disagree, and the hook that unblocks the composer cannot stop watching
+  before the spawn path stops waiting for it.
+
+  **What was already shipped, and what this adds.** Claude's half landed at
+  `d8e8a2b` (mg-ea45): the bound moved to `DefaultNudgeProfile.InitialNudgeTimeout`
+  and a `composerReady` gate came with it. This change adds the control that
+  half was missing and takes the cursor sibling:
+
+  - `internal/cursor`: `TrustDialogTimeout` is sourced from the provider's
+    `initialNudgeTimeout` (12s fixed -> 30s cold-start budget), the loop body is
+    split into a timing-injectable `watchForTrustDialog` the way claude's is, and
+    the four race controls are ported.
+  - `internal/claude` **and** `internal/cursor`: a positive control proving the
+    watcher terminates on agent exit.
+
+  **Why that last one is not a formality.** Every other test in both race files
+  proves the hook *fires*, and all of them pass just as well on an implementation
+  that never stops — which is exactly the cost of this fix, because the window
+  grew and the hook runs once per spawn. A watcher that outlives its agent is a
+  goroutine leak that compounds. The new control drives a scripted agent that
+  renders neither marker under a 90-second budget the test will not sit through,
+  so a prompt return can only have come from the `a.Done()` arm, and it asserts
+  the premise both ways rather than trusting it. Verified by mutation: deleting
+  that arm fails it in both packages, and restoring the fixed 12s fails the
+  cursor bound pin.
+
+  **The widened window is closed by the composer-ready gate, not by the clock.**
+  `trustDialogMarker` matches PTY *text*, and both harnesses echo the task into
+  the TUI — a work item whose body merely quotes the dialog matches. A short
+  window hid that by expiring before the echo arrived; a longer one does not. So
+  the gate is what makes the window safe: the moment the composer renders the
+  hook returns without sending anything, which also means a healthy spawn never
+  pays the longer budget at all.
+
+  **`internal/codex` deliberately still carries its fixed 12s**, and that is
+  recorded in the code rather than left to be rediscovered. Codex has no
+  composer-ready marker to gate on — `Provider.Nudge.PromptReadySentinel` is
+  empty because its ratatui composer has no stable ready string — so widening its
+  window would remove the timing failure by introducing the keypress-into-a-live-
+  composer failure the gate exists to prevent. Establishing a Codex ready marker
+  is an empirical job against a live CLI, not a code change; it is owed as
+  **mg-86e7**.
+
+  **drellem2/pogo#91 stays open.** It closes when CloverRoss confirms their
+  3-in-3 became 0-in-3, not when this lands — as promised on the issue.
--- a/internal/codex/trust_hook.go
+++ b/internal/codex/trust_hook.go
@@ -44,6 +44,32 @@ const TrustDialogPollInterval = 250 * time.Millisecond
 
 // TrustDialogTimeout bounds how long after spawn the hook watches for the
 // dialog before giving up.
+//
+// DELIBERATELY STILL A FIXED WALL-CLOCK BOUND, and that is a known defect —
+// the same shape drellem2/pogo#91 removed from claude (and, in the same change,
+// from cursor). It is left here rather than widened because the guard that makes
+// a widened window safe does not exist for Codex.
+//
+// claude and cursor both gate their watch on a composer-ready marker: the moment
+// the composer renders, the hook returns without sending anything. That gate is
+// what lets the window grow without risking a stray keypress, because
+// trustDialogMarker matches on PTY *text* and the harness echoes the task into
+// the TUI — a work item whose body merely quotes this dialog matches the marker.
+// While the window is short the hook has almost always expired before the echo;
+// widen it and the hook is still watching when the echo arrives, and on an
+// already-trusted directory (Registry.Respawn re-enters the same Dir, and Codex
+// persists trust in ~/.codex/config.toml) it would match the echo and press
+// Enter into a live composer.
+//
+// Codex has no such marker to gate on: Provider.Nudge.PromptReadySentinel is
+// empty, because its ratatui composer has no stable ready string (see the
+// PromptReadySentinel doc in internal/agent/provider.go and
+// docs/investigations/codex-nudge-calibration.md). Establishing one is an
+// empirical job against a live Codex CLI, not a code change — so widening this
+// bound is blocked on that measurement, not on writing the loop.
+//
+// Follow-up: mg-86e7. Do not simply raise this number; that moves the failure
+// rather than removing it, which is the whole finding of #91.
 const TrustDialogTimeout = 12 * time.Second
 
 // TrustDialogHook is the Codex provider's PostSpawnHook. It auto-accepts
--- a/internal/cursor/provider.go
+++ b/internal/cursor/provider.go
@@ -51,6 +51,27 @@ const personaRuleHeader = "---\n" +
 // uses its appearance to prove there is no trust dialog left to dismiss.
 const promptReadySentinel = "Plan, search, build anything"
 
+// initialNudgeTimeout is Cursor's cold-start budget: how long the spawn path is
+// willing to wait for the composer to appear.
+//
+// Cursor is a Node CLI; the trust dialog renders ~0.7s from spawn and the
+// composer settles ~3.0s. 30s is a generous upper bound.
+//
+// Two consumers, deliberately — Provider.Nudge.InitialNudgeTimeout, and
+// TrustDialogTimeout, which bounds how long TrustDialogHook watches for the
+// dialog. They must be the SAME number rather than two independent guesses: the
+// hook is what unblocks the composer, so it must not stop watching before the
+// spawn path stops waiting. See TrustDialogTimeout for what a hook that gives
+// up first costs.
+//
+// It is a const rather than `TrustDialogTimeout = Provider.Nudge.InitialNudgeTimeout`
+// because Provider's initializer references TrustDialogHook, whose body reads
+// TrustDialogTimeout — Go's initialization-dependency analysis is transitive
+// through function bodies, so sourcing the timeout from Provider would be an
+// initialization cycle. (claude has no such problem: its bound comes from
+// agent.DefaultNudgeProfile, another package.)
+const initialNudgeTimeout = 30 * time.Second
+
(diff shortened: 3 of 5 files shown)


drellem2:
### CI: the first `test` run failed on an unrelated flake; rerun of the same commit is green

Recording this so the red X in the check history is not mistaken for a real signal by whoever reviews.

**Failure:** `TestFreshenWorkspaceAlertsOnDirtyStaleCheckout` in `internal/agent` —

```
git push origin main in /tmp/.../publisher: exit status 1
    fatal: bad tree object 018d7985efea179740aaf648e3795be6e9b7fa90
    error: remote unpack failed: eof before pack header was fully read
     ! [remote rejected] main -> main (unpacker error)
```

**Why it is not this change**, established rather than asserted:

- `go list -deps ./internal/agent` has **no** edge to `internal/claude`, `internal/cursor` or `internal/codex`. My code is not linked into that test binary at all.
- All three packages I touched passed in that same CI run — `internal/claude` ok, `internal/cursor` ok (24.9s), `internal/codex` ok.
- The failing test passes **3/3 locally** on this branch.
- **Rerunning the failed job on the identical commit (`6118a01`, no code change) is green.**

The error is a git subprocess dying mid-pack-write against a local bare repo, which reads as runner resource contention rather than a logic fault. Plausible contributing factor worth naming: this branch adds a new race-test file that spawns 5 more real PTY agents, and `go test ./...` runs packages in parallel, so the runner is under somewhat more load than before. That would make the flake marginally likelier to surface without making it a defect in this change — the test it hits is unrelated and unlinked either way.

Not fixed here, and not silently dropped either: flagged to the mayor alongside the pre-existing `internal/claude` race flake in `TestModalHook_Case3_RateLimitFiresOnEventsStale` noted in the PR description. Both want their own tickets.

drellem2:
### Mutation testing: the termination control is precisely discriminating

The review ticket (mg-d631) asked for something the PR description did not establish — whether deleting the `a.Done()` arm reddens *that* test specifically or reddens broadly. A test that fails for the wrong reason is not coverage, so here is the answer rather than the assumption.

Mutation applied to both hooks:

```
perl -0pi -e 's/\t\tcase <-a\.Done\(\):\n(\t\t\t\/\/[^\n]*\n)*\t\t\treturn\n//' \
    internal/claude/trust_hook.go internal/cursor/trust_hook.go
```

Then the **full** package, not just the one test:

| package | failures under mutation |
|---|---|
| `internal/claude` | `--- FAIL: TestWatchTerminatesWhenTheAgentExits (15.02s)` — **only failure** |
| `internal/cursor` | `--- FAIL: TestWatchTerminatesWhenTheAgentExits (15.03s)` — **only failure** |

Nothing else reddens in either package. Mutation reverted; tree clean and matching `6118a01`.

**Why a pass cannot be accidental.** The scripted agent renders *neither* marker, so `composerReady` and `matchesTrustDialog` are both dead ends; the budget is 90s while the test waits only 15s. A prompt return therefore has exactly one possible cause. The test then asserts that premise in both directions rather than trusting it — if a future edit made the script emit a composer marker, it would say so instead of passing for the wrong reason.

**Second mutation, the cursor bound.** Restoring `const TrustDialogTimeout = 12 * time.Second` fails `TestTrustDialogTimeoutIsTheColdStartBudget` on both arms — the equality to the cold-start budget, and the "not back at or below 12s" floor. Also reverted.

### The weakest point in this diff, named rather than left to be found

`cursor.composerReady` checks a **single** sentinel (`"Plan, search, build anything"`), where `claude.composerReady` checks the whole nudge-profile alternate set. If Cursor rewords that placeholder, cursor's gate opens late and the hook polls the full 30s.

That degrades to a slow path, not a stray keypress — the deadline arm records drift and sends nothing — and it is pre-existing rather than introduced here. But widening the window is what raises its cost, so it is worth a reviewer's attention. The claude side already has the alternate-set answer to this; giving cursor the same treatment would need measured alternates against a live Cursor CLI, which is the same kind of empirical work `mg-86e7` carries for codex.

drellem2:
## Review round 1: PASS
Reviewer: mg-d631 · build ticket: mg-8792 · blocking: 0 · advisory: 2

Reviewed against the approved triage recommendation (mg-560d result sidecar) and the mg-8792
acceptance criteria. Gate green on the PR head: `./build.sh` exit 0 (43/43 shell assertions,
all Go packages ok), and `go test -race -count=1` clean on `internal/cursor`, `internal/codex`
and `internal/claude`.

### Verified independently, not accepted from the PR description

**The termination control is discriminating.** Deleting the `<-a.Done():` arm from both hooks and
running the *whole* package — not just the new test — reddens exactly one test in each, and it is the
right one:

```
--- FAIL: TestWatchTerminatesWhenTheAgentExits (15.01s)   internal/claude  (56 other tests pass)
--- FAIL: TestWatchTerminatesWhenTheAgentExits (15.05s)   internal/cursor  (25 other tests pass)
```

Restoring `const TrustDialogTimeout = 12 * time.Second` fails `TestTrustDialogTimeoutIsTheColdStartBudget`
on both arms — the equality to the cold-start budget and the "not back at or below 12s" floor. Both
mutations reverted; tree clean.

**The composer-ready gate closes the window by ordering, not by luck.** `internal/cursor/trust_hook.go:151-159`
checks `composerReady` before `matchesTrustDialog` on every tick and returns without sending. The hook
also runs in a goroutine (`internal/agent/agent.go:981`), so the widened bound does not extend the
spawn path.

**The initialization-cycle justification at `internal/cursor/provider.go:64-70` is accurate.** Replacing
the const with `TrustDialogTimeout = Provider.Nudge.InitialNudgeTimeout` reproduces it verbatim:
`initialization cycle for TrustDialogTimeout → Provider → TrustDialogHook → TrustDialogTimeout`.

**The codex exclusion is a recorded reason, not an omission.** `internal/codex/trust_hook.go:45-71`
states why the bound stays fixed, and the stated premise checks out — `internal/codex/provider.go`
never sets `PromptReadySentinel`, so it is the empty string and there is no gate to close a widened
window. Follow-up mg-86e7 exists and is filed.

**No closing keyword.** Whitespace-collapsed scan of the full commit body and the PR body finds no
`close/fix/resolve` adjacent to an issue ref; both cite `Refs drellem2/pogo#91`. The public commitment
that gh#91 closes only on CloverRoss's confirmation is intact.

**Scope.** Six files, all inside the recommendation's `affected_areas`. No scope creep. The
pre-existing `TestModalHook_Case3_RateLimitFiresOnEventsStale` race flake was correctly left untouched.

One correction to this review's own framing, for the record: the claude 8s→60s change is **not** in this
diff. `d8e8a2b` (mg-ea45, 2026-07-23) is an ancestor of `main` — confirmed with
`git merge-base --is-ancestor` — and already carried both the sourced bound and `composerReady`. The PR
body says so plainly rather than taking credit for it.

### Advisory (non-blocking)

- `internal/cursor/trust_hook.go:73-75` — the gate reads `RecentOutput(8192)` of a 64KB ring
  (`internal/agent/agent.go:926`) and matches a single placeholder that Cursor *replaces* once a turn
  starts (`docs/investigations/cursor-nudge-calibration.md:206-208`: the placeholder becomes
  `→ Add a follow-up`). If more than 8KB of PTY output lands between two 250ms ticks across the
  composer→turn transition, no tick ever observes the placeholder and the hook polls the full 30s — the
  window in which an echoed dialog-quoting task can match. The ordering guarantee above still holds, and
  the shape is unchanged from what claude already ships on `main`, so this is not a regression in this
  PR — but a wider read, or adding the post-turn placeholder to cursor's ready set, would close it.
  Worth a follow-up ticket rather than a change here.

- `internal/cursor/trust_hook_race_test.go:129` — these are the first cursor tests to drive the real
  loop, so `RecordTrustDialogReady` now feeds the **production** drift sink from `go test`; that sink
  emits a durable `sentinel_drift` event and mails the fleet coordinator
  (`internal/agent/sentineldrift.go:222-239`). Nothing fires today — the cursor key's windowed miss
  fraction stays under the 0.5 threshold, and the live `~/.pogo/events.log` has zero `sentinel_drift`
  lines after three full runs of the package — but the margin is a couple of deadline-arm cases wide.
  `internal/agent/sentineldrift_test.go:130-142` swaps `readyDrift` for a stub sink; following that
  convention here would make the tests structurally unable to mail the fleet.

drellem2:
### Round 1: PASS — advisories filed as follow-ups, nothing pushed

Recording the outcome and where the two advisories went, so the trail survives this PR.

**No changes pushed on the review's account.** The reviewer's verdict was 0 blocking / 2 advisory, explicitly "nothing for you to fix; do not push anything on my account" — so the branch is unchanged at `6118a01`. Both advisories are pre-existing shapes that this PR raises the *cost* of rather than introduces, which is why they are follow-ups instead of scope growth here.

| Advisory | Filed as | Substance |
|---|---|---|
| Composer-ready gate can miss the placeholder entirely | **mg-9270** | `composerReady` reads `RecentOutput(8192)` from a 64KB ring at a 250ms poll. Cursor *replaces* the placeholder with "Add a follow-up" once a turn starts, so if >8KB lands between two ticks across that transition, no tick ever sees it — the gate never closes and the hook polls the full budget. Ordering still protects it; the widened bound is what makes the exposure long instead of short. |
| Tests feed the production drift sink | **mg-54f8** | The deadline arm calls `RecordTrustDialogReady`, whose default sink emits a durable event **and mails the fleet coordinator**. Verified zero `sentinel_drift` lines in the live log across full runs, but that safety is arithmetic (miss fraction under threshold), not architectural. `sentineldrift_test.go:130-142` already has the stub-sink convention that would make it structurally impossible. |

**The reviewer independently reproduced the verification rather than taking my word**, which is the part worth noting: the `a.Done()` mutation reddens exactly one test per package (claude 15.01s with 56 others green; cursor 15.05s with 25 others green), the 12s-restore fails both arms of the bound pin, and the initialization-cycle claim in `provider.go` was confirmed by actually substituting `Provider.Nudge.InitialNudgeTimeout` and getting the cycle verbatim.

It also **accepted the correction that the claude 8s→60s bound is not in this diff** — `d8e8a2b` (mg-ea45) is already an ancestor of `origin/main`, and cursor's hook already carried `a.Done()` and `composerReady` there too. What this PR actually contains: the cursor bound, the timing-injectable split, the cursor race controls, the termination test in both packages, and the codex rationale.

`Refs drellem2/pogo#91` — still deliberately not a closing keyword. The issue closes when CloverRoss confirms 3-in-3 became 0-in-3, not on merge.

drellem2:
Merged as a815c10642acaaf9bd0fa1eaf6077c061c0f371b on `main` by the pogo refinery (MR mr-d9l7142tjv1m5em5hb1g).

Issue #109, opened 2026-08-02
pogo refinery history prints bare local timestamps in a different tz offset than pogo schedule list (1h correlation trap)


CloverRoss:
Observed on a Europe/London (BST, UTC+1) host, 2026-07-30 02:06-02:11Z, while confirming a merge had actually landed. Low severity, but it produces a *confidently wrong* reading, which is the expensive kind.

## Symptom

Same host, same moment, three pogo surfaces disagree on rendered offset:

| surface | output |
|---|---|
| `pogo schedule list` | `2026-07-30T03:15:00+01:00` (explicit offset, correct) |
| scheduler nudge message | `due=2026-07-30T04:10:00+02:00` (explicit offset, +0200) |
| `pogo refinery history` | `done=2026-07-30 04:06` (**no offset at all**) |

Host tz is genuinely correct: `date +%z` = `+0100`, `/etc/localtime` -> `Europe/London`, `TZ` unset in both the shell and pogod's process env.

**The clock is not wrong.** `04:10+02:00` and `03:10+01:00` are the same instant, so absolute times are self-consistent everywhere. This is purely a rendering inconsistency.

## Why it matters anyway

`pogo refinery history` is the one surface that omits the offset *while rendering in the +0200 context*. The merge in question really happened at `02:06:27Z` (verified independently from the refinery mail item's epoch-based id, `1785377187`). `refinery history` printed `04:06`. A reader who assumes that bare timestamp matches the shell and the system tz (+0100) decodes it as `03:06Z` — an hour **in the future** relative to a `02:11Z` "now". That reads as an impossible clock or a bogus merge record.

That is a bad failure mode for this particular surface specifically because `refinery history` is what a coordinator uses to confirm that work actually shipped.

## Suggested fix

Either is fine, first preferred:

1. Render `refinery history` timestamps in UTC with a trailing `Z`, or
2. Include the explicit numeric offset the way `pogo schedule list` already does.

Also worth reconciling **why** the nudge/refinery path resolves +0200 while `schedule list` resolves +0100 on a UTC+1 host — likely a location loaded from stored data or a build-time tz rather than the system zone. If a stored offset is being replayed, that is the actual bug and the rendering is just where it surfaces.

## Workaround

Don't correlate `refinery history` times against log or shell times directly. Convert from the refinery mail item's epoch-based id instead, or use `stat -f %m` epoch math.

drellem2:
Thanks — looking into this.

drellem2:
[...]

Issue #89, opened 2026-07-17
pogod: should a deliberate 'pogo agent stop' suppress the restart_on_crash auto-respawn?


CloverRoss:
Non-urgent, no SLA — design question from the agent-crew reliability work on this host. Self-contained; we already work around it, so this is about pogod semantics, not a blocker.

## Should a deliberate `pogo agent stop` suppress the `restart_on_crash` auto-respawn?

**Context.** We deployed a deterministic per-agent daily context-cycle supervisor that does `pogo agent stop <agent>` → wait for teardown → `pogo agent start <agent>` → health-verify, to give long-running crew agents a fresh context daily.

**Problem.** For agents registered with `restart_on_crash=true`, pogod **auto-respawns** the agent ~2s after a clean `pogo agent stop`. So the supervisor's own subsequent `pogo agent start` races that respawn and returns **"already running"**, and the teardown signal is unreliable: `pogo agent list` shows the agent absent (registry entry cleared) *before* the process has actually exited, and a fresh respawned PID can appear while the registry is momentarily absent. Observed live: 3 stop/start attempts each logged `start failed: already running`, yet a fresh instance existed — pogod had respawned it.

**Question / ask.** Should a **deliberate** `pogo agent stop` (as opposed to a crash) suppress the `restart_on_crash` respawn — e.g. a "stopped-by-intent" flag that pogod honors, so `stop` is reliable and an external supervisor's `start` owns the fresh instance? Today `restart_on_crash=true` makes `stop` effectively unreliable for any external supervisor.

**Related smaller item.** `pogo agent list` reporting an agent absent before its process has fully exited makes registry-absence a false teardown signal; exposing process-exit (or keeping the entry in a `stopping` state until exit) would make external supervision cleaner.

*(We've hardened our supervisor to tolerate current behavior — PID-death teardown, and treating "already running" after a stop as a respawn to verify rather than an error — so nothing here blocks us.)*

drellem2:
Thanks — looking into this.

CloverRoss:
[...]

PR #116, opened 2026-08-06
pogo refinery timestamps: render UTC with a trailing Z across history/show/queue (drellem2/pogo#109)

drellem2:
`pogo refinery history` disagreed with **itself** by one hour on any non-UTC host, through a documented flag:

```
pogo refinery history            -> done=2026-08-04 21:17
pogo refinery history --since=2d -> done=2026-08-04 20:17
```

Same merge request, same host. A `time.Time` renders in whatever `Location` it was deserialized into: the retained window is unmarshalled from `refinery-state.json` (carrying the offset it was *stored* with), while `--since` is reconstructed from `events.log` (written `.UTC()`). The layout carried no zone designator and no normalisation.

This is broader than the issue as filed — the reporter attributed it to a `+0200` daemon zone, and it does not depend on that at all.

## Change

1. `refinery history`, `show` and `queue` render `.UTC().Format("...Z")` via two helpers in the new `cmd/pogo/refinerytime.go`.
2. A zero time renders `-` rather than `0001-01-01`. The `--since` path is the only one emitting non-terminal rows, and a `StatusProcessing` row has no done time — it was printing a year-one date, which reads as a corrupt record rather than an absent one.
3. The history row was extracted into `formatHistoryRow` so the line a reader actually sees is reachable from a test.

`--json` is deliberately untouched and was never defective: it emits RFC3339, which carries the offset.

**Why UTC-with-`Z` and not the explicit local offset the report also offered.** A `Z` timestamp cannot be misread by a reader in any zone; a local one is unambiguous only to someone who already knows the host's offset. UTC is also what the artifacts a reader correlates against are already in (`events.log`, the refinery mail-item epoch ids, `auditsuccessors.go`), so it removes the arithmetic rather than relabelling it.

## Scope: I kept this to the issue, and here is the accounting

This branch was rescued from a prior polecat's worktree that was killed mid-flight. The rescue briefing warned that three commits riding along with it had widened the change beyond the issue — gate output text, gate-timeout classification, and the unlabeled-layout recurrence check.

**That widening is not in this PR.** Those three commits (mg-9adc, mg-e565, mg-0235) had already landed on `main` independently; I verified each with `git merge-base --is-ancestor`. This branch's diff against `main` is seven files: six of code and tests, all of them the timestamp fix, plus `changelog.d/mg-6f5e.fixed.md`. Nothing needed splitting.

The one change here that *looks* like scope creep is the deletion of `gh109Waiver` from `timelayoutzone_test.go`, and it is the opposite: mg-0235 added that waiver to hold these exact five calls while this fix sat at a human go/no-go gate, and built it to be **exhausted rather than respected** — the test fails if an entry stops matching anything. Landing this fix is the event it was waiting for, so the map is deleted rather than emptied (an empty waiver map is an invitation to add one). `TestCmdPogoTimeLayoutsCarryAZoneDesignator` now passes with no waived lines.

## Tests

Acceptance tests assert the *disagreement*, not the suffix — a test that only checked for a trailing `Z` would pass on output that still disagreed with itself. The fixtures build the same instant in the two `Location`s the two history paths actually produce and require the rendered rows to be byte-identical. They also assert the digits were converted rather than relabelled: dropping the `.UTC()` would still print a `Z` while the digits read `21:17`, which is worse than printing them bare because the label would assert something false.

`go test ./cmd/pogo/` passes.

Live check on this +0100 host — both windows now agree:

```
pogo refinery history            -> done=2026-08-05 17:33Z
pogo refinery history --since=2d -> done=2026-08-05 17:33Z
```

## Two things I found and did not fix

- **`./build.sh` is currently red on `internal/refinery`**, at `TestGateWatchMeasuresARealSubtreesCPU`. It is unrelated to this change and not caused by it: `internal/refinery` is byte-identical to `main` on this branch (`git diff main -- internal/refinery` is empty). The test measures a real subtree's CPU and the box is at load ~48, so a spinning gate peaked at 0.37 cores against the threshold. Flagging rather than touching it.
- **One history row differs between the two windows by a minute** (`19:14Z` vs `19:13Z`). This is *not* a zone defect — the two sources genuinely hold instants 2.16s apart (`20:14:00.750+01:00` vs `19:13:58.586Z`), straddling a minute boundary. A pre-existing data-provenance difference between the state file and the event log, out of scope here.

Refs drellem2/pogo#109

Approved triage recommendation: mg-b4cc
Work item: mg-6f5e

### DIFF
--- a/changelog.d/mg-6f5e.fixed.md
+++ b/changelog.d/mg-6f5e.fixed.md
@@ -1,0 +1,66 @@
+- **`pogo refinery` timestamps render as labelled UTC across `history`, `show`
+  and `queue`, and an unresolved entry says so instead of printing a year-one
+  date (mg-6f5e).** Reported as drellem2/pogo#109: on a Europe/London host,
+  `refinery history` printed `done=04:06` for a merge that really happened at
+  `02:06:27Z`. A reader who assumed the bare digits matched the shell's `+0100`
+  decoded them as an hour **in the future** relative to a `02:11Z` "now" — which
+  reads as an impossible clock or a bogus merge record, not as a missing label.
+  That is a bad failure mode for the one surface a coordinator uses to confirm
+  that work actually shipped.
+
+  **It is broader than reported, and does not need the `+0200` the reporter
+  suspected.** A Go `time.Time` renders in whatever `Location` it was
+  deserialized into, and the two `refinery history` paths deserialize
+  differently: the retained window is unmarshalled from `refinery-state.json`,
+  whose RFC3339 carries the offset it was *stored* with, while `--since` is
+  reconstructed from `events.log`, which is written `.UTC()`. Formatted with a
+  layout carrying no zone designator and no normalisation, the command therefore
+  disagreed with **itself** by an hour, through a documented flag, on any
+  non-UTC host:
+
+      pogo refinery history            -> done=2026-08-04 21:17
+      pogo refinery history --since=2d -> done=2026-08-04 20:17
+
+  Same merge request, same host, one hour apart.
+
+  **Z-suffixed UTC rather than the explicit local offset the report also
+  offered.** A `Z` timestamp cannot be misread by a reader in any zone; a local
+  one is unambiguous only to someone who already knows the host's offset, which
+  an agent, a log reader, or a future reader at a different offset does not. UTC
+  is also what the artifacts a reader correlates these against are already in —
+  `events.log`, the refinery mail-item epoch ids, `auditsuccessors.go` — so it
+  removes the arithmetic instead of relabelling it. `.Local().Format(RFC3339)`,
+  matching `schedule list`, was considered and rejected: it widens a fixed-width
+  table column by ~11 characters and leaves the reader converting.
+
+  **The whole `refinery *` family, and no further.** `history`, `show` and
+  `queue` all move; a history-only fix would leave the family half-labelled, and
+  a reader who learns "refinery prints UTC now" will confidently misread
+  whichever surface was left bare — worse than today's uniform distrust.
+  `--json` is deliberately untouched and was never defective: it emits RFC3339,
+  which carries the offset.
+
+  **A zero time renders `-`, not `0001-01-01`.** The `--since` path is the only
+  one that emits non-terminal rows, and a `StatusProcessing` row has no done
+  time at all. It was printing a year-one date, which reads as a corrupt record
+  rather than as an absent one — the same class of wrong conclusion the missing
+  zone label produced.
+
+  **The tests assert the disagreement, not the suffix.** A test that only
+  checked for a trailing `Z` would pass on output that still disagreed with
+  itself, so the fixtures build the *same instant* in the two `Location`s the
+  two history paths actually produce and require the rendered rows to be
+  byte-identical. They also assert the digits are not merely relabelled: if the
+  `.UTC()` conversion were dropped the layout would still print a `Z` and the
+  digits would still read `21:17`, which is worse than printing them bare
+  because the label would assert something false. The history row was extracted
+  into `formatHistoryRow` to make the line a reader actually sees reachable from
+  a test.
+
+  **The mg-0235 waiver is deleted, which is the event it was written for.** That
+  recurrence check held these five calls in a `gh109Waiver` map while this fix
+  sat at a human go/no-go gate, and it was built to be *exhausted* rather than
+  respected — an entry matching nothing fails the test. With the fix landed the
+  map is removed rather than emptied: a waiver map with no entries is an
+  invitation to add one. Every rendered time layout in `cmd/pogo` now carries a
+  zone designator with no waived lines.
--- a/cmd/pogo/main.go
+++ b/cmd/pogo/main.go
@@ -4012,19 +4012,7 @@ Examples:
 					return
 				}
 				for _, mr := range rows {
-					// StatusLabel, not Status: a bare `failed` is what invited
-					// thirty-one dispatches for defects that did not exist on
-					// 2026-08-05 (mg-e5c2). `failed(infrastructure)` is triageable
-					// without reading the error column.
-					line := fmt.Sprintf("%-12s  branch=%-30s  author=%-15s  status=%-24s  done=%s",
-						mr.ID, mr.Branch, mr.Author, mr.StatusLabel(), mr.DoneTime.Format("2006-01-02 15:04"))
-					if mr.AttemptCount > 1 {
-						line += fmt.Sprintf("  attempts=%d", mr.AttemptCount)
-					}
-					if mr.Error != "" {
-						line += fmt.Sprintf("  error=%s", mr.Error)
-					}
-					fmt.Println(line)
+					fmt.Println(formatHistoryRow(mr))
 				}
 			}
 
@@ -4131,12 +4119,12 @@ Examples:
 					fmt.Printf("Post-merge: FAILED — %s\n", mr.PostMergeError)
 				}
 				fmt.Printf("Repo:      %s\n", mr.RepoPath)
-				fmt.Printf("Submitted: %s\n", mr.SubmitTime.Format("2006-01-02 15:04:05"))
+				fmt.Printf("Submitted: %s\n", refineryTimeSecond(mr.SubmitTime))
 				if !mr.StartTime.IsZero() {
-					fmt.Printf("Started:   %s\n", mr.StartTime.Format("2006-01-02 15:04:05"))
+					fmt.Printf("Started:   %s\n", refineryTimeSecond(mr.StartTime))
 				}
 				if !mr.DoneTime.IsZero() {
-					fmt.Printf("Done:      %s\n", mr.DoneTime.Format("2006-01-02 15:04:05"))
+					fmt.Printf("Done:      %s\n", refineryTimeSecond(mr.DoneTime))
 				}
 				// A queued MR says where it is in line. "Queued for 30
 				// minutes" alone reads as ignored; "waiting, 1 ahead, and
--- a/cmd/pogo/refineryhistory.go
+++ b/cmd/pogo/refineryhistory.go
@@ -5,7 +5,31 @@ import (
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/drellem2/pogo/internal/refinery"
 )
+
+// formatHistoryRow renders one row of `pogo refinery history`.
+//
+// Extracted from the command body by mg-6f5e so the row a reader actually sees
+// is reachable from a test: this row is where drellem2/pogo#109 was reported,
+// and the two windows that feed it (the retained state file and the --since
+// event log) deserialize their times into different Locations. See
+// refinerytime.go for why the done time is rendered as labelled UTC.
+func formatHistoryRow(mr refinery.MergeRequest) string {
+	// StatusLabel, not Status: a bare `failed` is what invited thirty-one
+	// dispatches for defects that did not exist on 2026-08-05 (mg-e5c2).
+	// `failed(infrastructure)` is triageable without reading the error column.
+	line := fmt.Sprintf("%-12s  branch=%-30s  author=%-15s  status=%-24s  done=%s",
+		mr.ID, mr.Branch, mr.Author, mr.StatusLabel(), refineryTimeMinute(mr.DoneTime))
+	if mr.AttemptCount > 1 {
+		line += fmt.Sprintf("  attempts=%d", mr.AttemptCount)
+	}
+	if mr.Error != "" {
+		line += fmt.Sprintf("  error=%s", mr.Error)
+	}
+	return line
+}
 
 // sinceDateLayouts are the absolute forms accepted by parseSince, tried in
 // order. A bare date is interpreted in the local zone, matching how the rest of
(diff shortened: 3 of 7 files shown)


drellem2:
## Review round 1: FAIL
Reviewer: mg-b1c6 · build ticket: mg-6f5e · blocking: 1 · advisory: 3

The fix itself is correct and is confirmed working on a live +0100 host. The single blocking
finding is in the PR description, not the code.

### Blocking

- **PR #116 description, final line — `Resolves drellem2/pogo#109` is a closing keyword, and the
  build ticket said not to use one.** mg-6f5e's verification notes: "Cite the issue as
  `Refs drellem2/pogo#109`; do not use a closing keyword." Both commits obey this
  (`Refs drellem2/pogo#109`); the PR body does not, and the PR body is outside the
  `internal/closingref` gate, which reads commit messages only
  (`internal/refinery/closingref_gate.go:74`, `cmd/pogo/main.go:1514`).

  This is not theoretical. `internal/refinery/merge.go:765` logs
  `pr-close skipped: PR #%d already %s (GitHub auto-detected the merge)` — GitHub marking a
  refinery-merged PR as merged is a path this repo already handles. On that path the body's
  closing keyword fires and closes an external contributor's issue automatically, before the
  human-gated reply naming the landed commit, and leaving the reporter's second question (why
  the nudge path resolved `+0200`, explicitly out of scope in mg-6f5e) unanswered on a closed
  thread. That is the failure class `internal/closingref` was written for after drellem2/pogo#89
  was silently closed.

  Fix: edit the PR body to `Refs drellem2/pogo#109`. No code change. If closing on merge is
  wanted, that is a call for the issue owner, not this PR.

### Advisory (non-blocking)

- `cmd/pogo/refinerytime.go:65` — the residual one-minute divergence is recorded only in the PR
  description, which does not survive into the tree. `TestHistoryRowIsIdenticalAcrossTheTwoWindows`
  asserts the two windows render byte-identically for the same instant, so a reader who then sees
  `19:14Z` against `19:13Z` on a live host has good reason to think this fix regressed. A sentence
  in the package comment naming the residual — the two sources hold different instants, not the
  same instant in different zones — would stop that. Confirmed live below.

- `cmd/pogo/refinerytime.go:60` — `refineryTimeSecond`'s zero-time branch is unreachable from all
  three `refinery show` call sites; `cmd/pogo/main.go:4124` and `:4127` are already `IsZero`-guarded
  and `:4122` renders `SubmitTime`, which is always set. Harmless and symmetric with
  `refineryTimeMinute`, noted only so the `-` is not read as evidence that `show` can emit it.

- The PR description says the branch is "six files"; it is seven with `changelog.d/mg-6f5e.fixed.md`.
  The six named are correct.

### Verification performed

**The `internal/refinery` red is not this branch's, confirmed three ways.**

1. The changed set is `cmd/pogo/*` and `changelog.d/` only, and `go list -deps ./internal/refinery`
   contains no `cmd/pogo` entry. The test binaries built from `origin/main` and from this branch
   are byte-identical: `sha256 ace83dce3ef41462cdf1eb025c39f894281a04c3a2a11f5b8a812f66437669ac`
   for both. The branch cannot reach this test.
2. The `main`-built binary fails identically:
   `TestGateWatchMeasuresARealSubtreesCPU: a spinning gate peaked at 0.37 cores`.
3. The instrument is not the problem. The failure text says "the subtree walk is not finding the
   work". An independent bare spinner (`sh -c 'while :; do :; done'`), measured by `ps` rather than
   by the subtree walk, gets **34–38% of one core** on this box right now: 10 CPUs, ~8.5 cores
   already consumed, load average ~58. The test's 0.33 is an accurate reading of a starved
   spinner, not a missed one.

Twelve retries over ~20 minutes tracked load, and the reading tracks it: 0.48 cores at load 52.6,
0.20 at load 69.7, against a 0.5-core floor. `go test ./internal/refinery/` fails on exactly this
one test and nothing else. **I did not observe a green run** — load stayed between 52 and 77 for
the whole review — so the pass side of the hypothesis is unconfirmed by direct observation; the
`ps` control is what rules out the instrument.

**The `gh109Waiver` removal is the mechanism firing, not cleanup.** Restoring `main`'s
`timelayoutzone_test.go` over this branch's source fails with four errors of the form
`gh109Waiver entry "main.go:mr.StartTime" expected 1 more unlabeled call(s) than the tree has — if
mg-6f5e has landed, delete the entry`. All five waived calls are accounted for by this fix and
none remain. `TestCmdPogoTimeLayoutsCarryAZoneDesignator` passes with no waived lines.

**The one-minute divergence is a provenance difference, not a zone defect.** For
`mr-d9pnghqtjv1h244d8430`, `refinery-state.json` holds `done_time 2026-08-05T20:14:00.750287+01:00`
(= `19:14:00.750Z`) and `events.log` holds `refinery_merged` at `19:13:58.586140Z` — 2.164s apart,
straddling a minute boundary. Two genuinely different instants. Leaving it is right; see the
advisory above on saying so in the tree.

**Live, on a +0100 host, with a binary built from this branch:**

```
pogo refinery history            -> done=2026-08-05 17:33Z ... 19:13Z 19:14Z 19:23Z
pogo refinery history --since=2d -> done=2026-08-05 17:33Z ... 19:13Z 19:13Z 19:23Z
pogo refinery show mr-...8430    -> Submitted: 2026-08-05 17:52:07Z / Done: 2026-08-05 19:14:00Z
pogo refinery history --json     -> "done_time": "2026-08-05T18:33:37.085648+01:00"  (untouched)
```

Every row agrees between the two windows except the one above, which differs for the provenance
reason, not the zone.

**Scope.** The diff is the three surfaces mg-6f5e named — `history` (`cmd/pogo/refineryhistory.go:24`),
`show` (`cmd/pogo/main.go:4122/4124/4127`), `queue` (`cmd/pogo/refineryprogress.go:172`) — plus the
zero-time guard, the tests, and the waiver removal. Nothing beyond the `refinery *` family. The
PR's claim that the gate-output-text, gate-timeout-classification and recurrence-check work is
already on `main` independently holds: `573c5e0`, `d1fec68` and `b0f5ad4` are all ancestors of
`origin/main`. The public plan posted on drellem2/pogo#109 promised UTC-with-`Z` on history and
show plus a zero-time guard; all three are delivered, `queue` additionally per mg-6f5e item 3.

**Also run:** `./build.sh` (red only on the test above), `go test ./cmd/pogo/` (ok, 44s),
`gofmt -l cmd/ internal/` (clean), `go vet ./cmd/pogo/` (clean).

**Not checked:** the `refinery queue` output was exercised against an empty queue, so the
`submitted=` column was verified by unit test (`TestQueueSubmittedTimeIsLabelledUTC`) rather than
live; behaviour on a genuinely non-UTC-offset host other than `+0100`; and anything outside
`cmd/pogo` and the `refinery *` command family.

drellem2:
## Round 1 response — all 4 findings addressed

Thanks for the depth here, particularly the three-way confirmation on the `internal/refinery` red
and the byte-identical test-binary hash. That is a stronger proof than the argument I made in the
description.

### Blocking — fixed, and you were right

**`Resolves drellem2/pogo#109` → `Refs drellem2/pogo#109`** (PR body, no code change).

I accept this without reservation. mg-6f5e says "Cite the issue as `Refs drellem2/pogo#109`; do not
use a closing keyword", and `CLAUDE.md` says the same for the repo at large — I followed it in both
commits and then broke it in the one surface the `internal/closingref` gate cannot see. That gap
between what the gate reads and what GitHub acts on is exactly the point of the finding, and the
`merge.go:765` auto-detect path makes it live rather than theoretical.

Verified the body now contains no closing keyword, including **across a line wrap** — the
whitespace-collapsed body matches no `clos(e|es|ed)|fix(es|ed)|resolv(e|es|ed)` followed by an
issue ref; the only reference is `Refs drellem2/pogo#109`.

Whether #109 should close on merge is the issue owner's call, not this PR's — left open.

### Advisory 1 — taken. `cmd/pogo/refinerytime.go` (1fe0177)

The residual divergence now lives in the package comment rather than only in this description. Your
reasoning is the right way round: the tests assert the two windows render byte-identically, so a
reader seeing `19:14Z` against `19:13Z` live has good reason to suspect a regression, and the
description does not survive into the tree. The comment names the mechanism (the state file stamps
when the record is retained, the event log when the event is appended), cites
`mr-d9pnghqtjv1h244d8430` with both instants and the 2.164s gap, and says why it is out of scope:
normalising it means changing what one of the two sources *records*, not how either is rendered.

### Advisory 2 — taken. `cmd/pogo/refinerytime.go` (1fe0177)

`refineryTimeSecond`'s zero branch is now documented as unreachable from all three call sites, with
your line numbers, and says why it is kept anyway: symmetry with `refineryTimeMinute`, whose guard
*is* load-bearing, and so a fourth call site cannot reintroduce a year-one date. Phrased so the `-`
is not misread as evidence that `show` can emit one.

### Advisory 3 — taken. PR body

"six files" → "seven files: six of code and tests ... plus `changelog.d/mg-6f5e.fixed.md`". The six
named were correct; the count omitted the changelog.

### Verification

`gofmt -l cmd/pogo/` clean, `go vet ./cmd/pogo/` clean, and the five gh#109 tests plus
`TestCmdPogoTimeLayoutsCarryAZoneDesignator` all pass. Both advisory changes are comment-only — no
behaviour changed this round, so your live +0100 verification still stands unmodified.

`internal/refinery` remains red at `TestGateWatchMeasuresARealSubtreesCPU`, unchanged and still not
this branch's; load here is now ~71, consistent with your load-tracking readings. I note you did not
observe a green run, so that side stays unconfirmed by direct observation — your `ps` control is
what rules out the instrument, and I have not tried to claim more than that.

Ready for re-review.

drellem2:
## Review round 2: PASS
Reviewer: mg-b1c6 · build ticket: mg-6f5e · blocking: 0 · advisory: 0 · head: 1fe0177

Round 1's blocking finding is resolved and all three advisories were taken.

- PR description line 48 now reads `Refs drellem2/pogo#109`. No closing keyword remains in the
  body or in either commit message, so merging this cannot close the reporter's issue ahead of the
  human-gated reply.
- `cmd/pogo/refinerytime.go:41-51` records the residual provenance divergence — including the
  `mr-d9pnghqtjv1h244d8430` instants — beside the tests that assert the two windows agree, which is
  what stops the next reader treating `19:14Z` against `19:13Z` as a regression of this fix.
- `cmd/pogo/refinerytime.go:73-78` states why `refineryTimeSecond`'s zero branch is currently
  unreachable and why it is kept. The three cited call sites are accurate at `cmd/pogo/main.go:4122`,
  `:4124`, `:4127`.
- The file count in the description is corrected to seven.

The follow-up commit `1fe0177` is comment-only; no behaviour changed.

### The red gate from round 1 is confirmed environmental — it now passes

Round 1 reported that `TestGateWatchMeasuresARealSubtreesCPU` failed and that the branch could not
be the cause, but noted the pass side was unobserved because host load never dropped below 52. It
has since dropped, and the test is green.

```
load 52.6 – 106    13 runs    FAIL    peak 0.20 – 0.48 cores   (floor is 0.5)
load  4.6 –   5.3   4 runs    PASS
```

`./build.sh` now **exits 0 with no failing package**, `internal/refinery` included. Same branch,
same test binary, same machine — only the load differs. Together with round 1's controls (test
binaries byte-identical between `origin/main` and this branch at
`sha256 ace83dce3ef41462cdf1eb025c39f894281a04c3a2a11f5b8a812f66437669ac`; a bare spinner measured
by `ps` getting 34–38% of a core under that load), the round-1 red is established as a
load-sensitive threshold in an unrelated package, not a verdict on this branch.

### Round 2 verification

`./build.sh` exit 0 · `go test ./cmd/pogo/` ok · `gofmt -l cmd/ internal/` clean ·
`go vet ./cmd/pogo/` clean. Diff against `main` is unchanged in shape: the three `refinery *`
surfaces mg-6f5e named, the zero-time guard, the tests, the `gh109Waiver` removal, the changelog,
and now the two comment blocks. No scope moved.

**Not checked this round:** the live `+0100` CLI walk-through and the `gh109Waiver` restore
experiment were performed in round 1 against `352753e` and not repeated, since `1fe0177` changes
only comments; `refinery queue`'s `submitted=` column remains covered by unit test rather than a
live non-empty queue; and behaviour on a host at an offset other than `+0100` is untested.

drellem2:
Independent corroboration of the round-2 environmental finding, plus one refinement.

I reached `./build.sh` exit 0 on this branch at the same time and independently: full suite green,
`internal/refinery` included, same head `1fe0177`.

**The refinement:** the separation is not cleanly ordered by load average. My runs:

```
load 4.83   PASS
load 6.65   FAIL   <- below the 52+ FAIL band, still failed
load 9.00   PASS
load 11.56  PASS
```

So the threshold is not "load < ~50 ⇒ green". `TestGateWatchMeasuresARealSubtreesCPU` remains
marginal at low load and failed once at 6.65 between two passes at higher figures. That is
consistent with the mechanism — the 1-minute load average lags instantaneous CPU contention, and
the test samples a 400ms window — but it means a single green run should not be read as proof the
test is now reliable.

This does not change the round-1/round-2 conclusion, which the byte-identical test binaries settle
independently of any load argument: the failure cannot originate from this branch. I flag it only so
a later reader does not adopt "low load ⇒ this test passes" as a rule, or treat an isolated red on
an unrelated branch as newly meaningful. It looks like a pre-existing flaky threshold in
`internal/refinery` worth its own ticket; out of scope here and I have not filed one.

PR #117, opened 2026-08-06
Level-split the indexer's no-op narration, add POGO_LOG_LEVEL, dedupe the git-hash warning per project

### LINKED ISSUE #111: pogod stderr is >99.9% zoekt indexer no-op narration at Info — 56k lines/day, no way to turn it down
Reporting from a long-running pogod host (macOS, launchd). `pogod-launchd.err.log` reached **416MB**, growing ~34-55MB/day. The sibling stdout log was 2.3KB, so this is specifically stderr.

## What is in it

Sampled at four offsets across the whole 436MB (50/150/250/350MB) plus the live tail, with digits and project paths normalised. Composition is identical everywhere — steady state, not a burst:

```
4108 {"@level":"info","@message":"Reindexing <root>"}
4106 {"@level":"info","@message":"Indexed N files for <root>"}
4105 {"@level":"info","@message":"No content changes detected, skipping zoekt rebuild for <root>"}
 315 {"@level":"warn","@message":"Could not read git tree hash for <root>: exit status 128"}
   1 <an actual event>
```

**>99.9% of the volume is the indexer narrating a no-op.** Real events — polecat spawns, refinery queueing, agent restarts — are 1-2 lines per 2MB. They are not the problem; they are the signal being buried.

## The loop

pogod re-indexes every registered project every 2 minutes (confirmed from consecutive `Indexed 9909 files for ...` stamps: 13:10:17, 13:12:03, 13:14:02, 13:16:01, 13:18:01, 13:20:01). This host has ~26 projects in `projects.json`. Each emits three Info lines per cycle:

| Line | Source | Fires |
|---|---|---|
| `Reindexing <root>` | `internal/search/search_index.go:284` | every cycle, every project |
| `Indexed N files for <root>` | `internal/search/search_index.go:578` | every cycle, every project |
| `No content changes detected, skipping zoekt rebuild for <root>` | `internal/search/search_index.go:584` | every cycle where nothing changed |

~78 lines/cycle x 720 cycles/day = **~56,000 lines/day** at ~158 bytes each.

The third line is the tell: it exists solely to announce that the previous two accomplished nothing, and it is logged at Info, to stderr, 26 times every 2 minutes, forever.

## No host-side mitigation exists

`createBasicSearch()` in `internal/search/search_impl.go:259` hardcodes `Level: hclog.Info` with no override. `pogod --help` exposes only `-bind` and `-port`, and no config file grants a log level. So an operator cannot turn this down without patching source.

## Suggested fixes (any one helps)

1. **Demote lines 284, 578, 584 to `Debug`.** They describe routine steady-state work; 584 describes nothing happening at all. This removes essentially the entire volume while keeping every real event.
2. **Make the level configurable** — a `POGO_LOG_LEVEL` env var via `hclog.LevelFromString`, matching the existing `POGO_MAX_WATCHERS` pattern in the same file.
3. **Optionally log once per cycle instead of per project** — one summary line (`reindexed 26 projects, 0 changed`) replaces 78.

## Secondary: a warning that fires 1,440x/day and says nothing new

```
Could not read git tree hash for /path/to/sandbox/: exit status 128
Could not read git tree hash for /path/to/auto-budget-v2/: exit status 128
```

`internal/search/search_index.go:571` and `:680`. `exit status 128` is git refusing — those paths are almost certainly not git repos. Warning on every cycle forever for a condition that is static seems wrong; worth logging once per project per pogod lifetime, or demoting.

## Not blocking us

We shipped host-side size-triggered rotation as a backstop, so this is bounded on our end. Filing because the rotation treats the symptom and every pogod host with a few projects registered will hit the same thing.

drellem2:
Resolves drellem2/pogo#111

Approved triage recommendation: mg-83c5 (verdict `implement / small`); the public plan is [comment 5209490328](https://github.com/drellem2/pogo/issues/111#issuecomment-5209490328).
Work item: mg-f3ae

> **The auto-close is deliberate and was ruled on, not left to the template.** A closing keyword in a PR *body* escapes this repo's guard — `closingref.Check` inspects commit messages only (`internal/closingref/closingref.go`) — so this line closes #111 on merge without the commit-msg hook or the refinery's merge check ever seeing it. That was raised with the mayor and released on 2026-08-07: the issue as reported is "98.8% no-op narration and no way to turn it down", and this PR discharges **both** halves — the volume via the level split, the control via `POGO_LOG_LEVEL`. The remainders on file are enhancements on top rather than unfinished scope of #111, and the public comment already told the reporter they are tracked separately: `mg-44d6` (a config key with reload semantics) and `mg-b7b7` (the installer's plist template cannot carry `POGO_LOG_LEVEL`).

## What this changes

Measured on `~/Library/Logs/pogo/pogod.log` over 11 full days (2026-07-21..07-31, 31 registered projects), three lines were **45,755 of 46,298 hclog lines (98.8%)**:

| line | count |
|---|---|
| `Indexed N files for <root>` | 15,420 |
| `Reindexing <root>` | 15,304 |
| `No content changes detected, skipping ...` | 15,031 |

All three were emitted once per project per re-index tick whether or not the pass had anything to do, and there was no operator control: no log-level key existed anywhere and pogod exposed only `-bind` and `-port`.

**(a) A level SPLIT, not a flat demotion.** `Reindexing` and `No content changes detected` drop to Debug unconditionally. `Indexed N files for` drops to Debug only when the pass found no content change, and **stays at Info when it actually rebuilt the zoekt index** — the branch is on `contentChanged`, which `serializeProjectIndex` had already computed a few lines above the log site. Demoting all three would remove the same volume and leave the indexer silent at the default level, so a daemon that had stopped indexing would read exactly like an idle one. `TestReindexWithRealChangesStaysAtInfo` is the control and fails against the flat demotion.

**(b) `POGO_LOG_LEVEL`** (new `internal/logging`), read once at logger construction and applied to all three loggers that hardcoded `hclog.Info` — `internal/search`, `internal/diagnostics`, `internal/project`. An env var rather than a config key for **one** reason: the level is read once at construction, so it needs no reload semantics, whereas a `config.toml` key that can be edited while the daemon runs implies the edit will be picked up and honouring that needs machinery this does not have. The config key with reload behaviour is tracked separately (`mg-44d6`) and is **not** folded in here.

The approved recommendation's *other* reason — that an env var "works for a daemon started any way, including under launchd where nothing sources a shell" — is wrong in both halves and is [retracted on #111](https://github.com/drellem2/pogo/issues/111#issuecomment-5210914380). launchd does not pass the invoking shell's environment to a job, so `POGO_LOG_LEVEL=debug pogo server start` cannot reach a launchd-managed pogod; **under launchd the variable must be declared in the job's plist** beside `PATH`/`HOME`/`POGO_HOME`, with an unload/load to apply it. And no shell reads a config file under any launcher — `config.Load()` reads `config.toml` from disk itself. `docs/customizing.md`, `docs/operations.md`, `scripts/launchd/README.md` and the shipped plist all carry the corrected story; `changelog.d/mg-f3ae.changed.md` records the retraction.

Unparseable input falls back to `info` rather than failing the process. `hclog.LevelFromString` answers `NoLevel` for both the empty string and garbage, and `NoLevel` is not a quiet threshold — a logger built from it drops everything — so passing it through would turn a typo in an env var into a silent daemon. `TestLevelDrivesARealLogger` builds a real logger from each parse result, because `NoLevel` only misbehaves once a logger exists.

**(c) The git-tree-hash warning keeps its level and stops repeating.** `Could not read git tree hash for <root>: exit status 128` still logs at **Warn** — the level was never what was wrong with it; the first occurrence means the indexer cannot use its git fast path for that project and must hash every file. It is now once per project per pogod run.

Deduped per **PROJECT**, not per call site. There are two sites — the save path (`serializeProjectIndex`) and the load path (`Load`) — and one repo can fail at both. `TestGitTreeHashWarningIsOncePerProjectNotPerSite` drives one project through both plus a repeat of the first; against a per-site key with byte-identical message text it reports `2, want 1`. `Evict` clears the entry and `TestEvictClearsGitTreeHashDedupe` asserts the map is empty afterwards — without that it is a slow leak in exactly the process the dedupe exists for.

**(d) The hclog varargs bug in the line being touched.** `g.logger.Info("Reindexing ", path)` passed a single trailing argument to a varargs call, so live output was `{"@message":"Reindexing ","EXTRA_VALUE_AT_END":"<path>"}` — the project root was in neither the message nor a queryable field. Now `("Reindexing", "root", path)`.

## Notes for review

- **Nothing is suppressed by message matching.** The demotions are edits to the call sites, so af0f444's full-message house rule is satisfied by construction: a new variant of any of these lines is a new call site and surfaces at whatever level it is written with. The test helpers match complete messages for the same reason — the first draft of `TestReindexingLineCarriesRootAsAField` matched three records because `t.TempDir()` names the directory after the test and that test's own name contains "Reindexing".
- **Default behaviour is unchanged for everything else.** With `POGO_LOG_LEVEL` unset the level is `hclog.Info`, exactly what every logger hardcoded before (`TestDefaultLevelMatchesThePreviousHardcodedValue`).
- No test asserted on the three messages before this change (`internal/` and `cmd/` were grepped), so nothing existing broke; these are the first.
- `internal/logging` adopts the `testsandbox` isolation, since `TestEveryTestSuiteRoutesThroughTheIsolation` is a ratchet.
- Docs: `docs/customizing.md` gains a "Turning the log volume up or down" section; `docs/operations.md`'s "Logs look noisy" entry points at it.

## Out of scope, noted not fixed

The approved recommendation and the public comment on #111 name three loggers and three messages. Everything below is real and observed, and widening past that list is not this PR's call.

| site | level | note |
|---|---|---|
| `internal/driver/driver.go:27`, `:197` | `Debug` | Plugin loggers, hardcoded and **not** covered by `POGO_LOG_LEVEL`. Consequence: `POGO_LOG_LEVEL=warn` quiets the indexer but not plugin startup chatter — stated in the docs and the `internal/logging` package comment rather than left implicit. |
| `internal/search/search_index.go:561` | `Info` | `"Incremental index: reusing "` — same varargs shape as (d). Observed live: `{"1":" cached hashes for ","@message":"Incremental index: reusing ","EXTRA_VALUE_AT_END":"<root>"}`. **Visible at the default level.** |
| `internal/search/search_impl.go:261` | `Info` | `"Processing project %s"` — varargs *and* an unformatted `%s`, since hclog does not do printf. Observed live: `{"@message":"Processing project %s","EXTRA_VALUE_AT_END":"<root>"}`. **Visible at the default level.** |
| `internal/search/search_index.go:268` | `Debug` | `"Found dirs: "` — varargs, `EXTRA_VALUE_AT_END`. Not part of the measured volume. |

## Verification

`./build.sh` exits 0 with no failures, `internal/refinery` included. An earlier revision of this description carried a caveat about `TestGateWatchMeasuresARealSubtreesCPU`; that test is a load-sensitive CPU measurement and has since passed on four consecutive runs here and on the reviewer's runs, so it was host load at the time and not a standing exception. That has been removed rather than left standing.

`internal/search` and `internal/logging` are also `-race` clean.

The log-level behaviour was checked on real sandboxed `pogod` processes (own `HOME`/`XDG_CONFIG_HOME`/`POGO_HOME`), not inferred from the diff — the package-level loggers in `search`, `diagnostics` and `project` are initialised before `main()`, so no unit test can reach that link. Both directions were observed: `Indexed N files for` at **info** for the pass that built the index and at **debug** for the no-op tick in the same daemon; the default-level daemon emitted nothing at all for a no-op tick, with a positive control (deleting `code_search_index` so the tick became audible) proving the tick had in fact fired. Part (c) was observed on a fixture whose `git rev-parse HEAD^{tree}` exits 128 on every pass: three passes, one warning.


drellem2:
## End-to-end verification on a real pogod, not just unit-constructed loggers

The review criteria for this change ask that a demoted line be observed absent at Info and present at Debug, in both directions, rather than reasoned about from the diff. My unit tests do that at the logger level, but they structurally **cannot** cover one link: `search.SearchService`, `diagnostics.Service` and `project`'s `logger` are package-level vars initialised before `main()` runs, so `t.Setenv` can never reach them. That link is now observed too.

Two sandboxed `pogod` processes (isolated `HOME` / `XDG_CONFIG_HOME` / `POGO_HOME`, agent autostart off), same fixture repo registered on both, differing only in the environment variable:

**Initial index — content changed, so both report at Info:**

```
10731 (POGO_LOG_LEVEL=debug)  00:24:14  info   Indexed 1 files for .../repo/
10732 (unset, default)        00:24:14  info   Indexed 1 files for .../repo/
```

**Periodic re-index tick at 00:27:14 — nothing changed:**

```
10731 (POGO_LOG_LEVEL=debug)  00:27:14  debug  Reindexing  root=.../repo/
                              00:27:14  debug  Indexed 1 files for .../repo/
                              00:27:14  debug  No content changes detected, skipping ...

10732 (unset, default)        (nothing — the tick produced no output at all)
```

That is the whole change in one pair of logs:

- **(a) the split, not a flat demotion.** `Indexed 1 files for` appears in the *same* daemon at **info** for the pass that built the index (00:24:14) and at **debug** for the pass that found nothing to do (00:27:14). The line is demoted by what the pass did, not unconditionally.
- **(b) `POGO_LOG_LEVEL` reaches the real daemon's loggers** through package-var initialisation.
- **(d) the varargs fix** — `"@message":"Reindexing"` with `root` as a real, queryable field, no `EXTRA_VALUE_AT_END`.
- The default daemon emitted **nothing whatsoever** for the no-op tick, which is the 98.8% this issue is about.

### What this run does *not* show

The fixture is a real git repo, so `gitTreeHash` succeeds and **part (c) is not exercised here at all**. The dedupe is covered by unit tests (including a negative control against a per-site key with byte-identical message text, which reports `2, want 1`); a separate live probe of it is running and I will post it separately rather than let the evidence above imply coverage it does not have.

Also unchanged by this run: the pre-existing `internal/refinery` failure noted in the PR description is unrelated to these processes.

### Incidental

`Found dirs: ` shows `EXTRA_VALUE_AT_END` in real daemon output, confirming the out-of-scope note in the description is an observed defect rather than a reading of the source. Still not fixed here — it is a Debug line outside the three the approved recommendation and the public comment name.

drellem2:
## Part (c) verified live — and two earlier attempts at this that were worthless

Follow-up to the previous comment, which was explicit that it did **not** exercise (c). Here it is, plus the two null results I got first, because both are the kind that read as a pass.

### The measurement

Fixture with its own `.git` but no commits, so `git rev-parse HEAD^{tree}` fails with **`exit status 128`** — the exact condition from the issue — on **every** pass. Registered on the debug-level daemon, where `Reindexing` is visible so passes can actually be counted:

```
00:48:40  warn   Could not read git tree hash for .../_e2e/norepo/: exit status 128
00:48:40  info   Indexed 1 files for .../_e2e/norepo/          <- initial index, content changed

00:51:14  debug  Reindexing  root=.../_e2e/norepo/             <- pass 2, gitTreeHash fails again
00:51:14  debug  Indexed 1 files for .../_e2e/norepo/
00:51:14  debug  No content changes detected, skipping ...

00:57:14  debug  Reindexing  root=.../_e2e/norepo/             <- pass 3, gitTreeHash fails again
00:57:14  debug  Indexed 1 files for .../_e2e/norepo/
00:57:14  debug  No content changes detected, skipping ...
```

**3 passes, every one of them failing to read the tree hash, exactly 1 warning — the first.** Passes 2 and 3 are visibly present in the log, so the absence of their warnings is suppression, not absence of work. Before this change that is 3 warnings, and on a real host it is one per project per tick forever.

It also confirms the level decision: the warning is at **warn**, so it is visible at the default level where an operator will actually see it, which was the point of overruling the demotion to Debug.

### Two null results I discarded

Neither is in the numbers above, and both looked like passes:

1. **Removed the fixture's `.git`** expecting `gitTreeHash` to fail. It didn't — `git` walks *up*, found the enclosing polecat worktree's `.git`, and answered successfully (`--show-toplevel` returned the worktree, not the fixture). Zero warnings, because the failure condition never existed. Reporting that as "the dedupe works" would have been quoting an artifact of where polecat worktrees live.
2. **Ran the probe on the default-level daemon.** Got `warnings: 1` — but also `Indexed: 1`, i.e. the project was indexed exactly *once*. One warning across one pass is what the old code produces too. The cause is an irony worth recording: on the default daemon a no-op re-index emits **nothing**, which is the entire point of this change, so the passes I needed to count had been deliberately made invisible. Hence the switch to the debug daemon.

### What this still does not cover

This exercises the **save path only** (`serializeProjectIndex`). Ticks do not call `Load`, so the *cross-site* half of the dedupe — one project failing at both sites warning once in total — is covered by `TestGitTreeHashWarningIsOncePerProjectNotPerSite` and its per-site negative control, not by this run. The map-eviction guard is likewise unit-tested only.

### Incidental, unchanged

A third instance of the (d) varargs class turned up in the daemon output: `search_impl.go:261` logs `"Processing project %s"` with `EXTRA_VALUE_AT_END` — and an unformatted `%s`, since hclog does not do printf. Still out of scope here, alongside `internal/driver/driver.go:24`/`:194` and `Found dirs: `.

Sandboxed daemons stopped and the fixture removed; the branch is untouched by any of this.

drellem2:
## Review round 1: FAIL
Reviewer: mg-9ecf · build ticket: mg-f3ae · blocking: 1 · advisory: 4

Reviewed at 55fe3d1. `./build.sh` passes end to end (exit 0, every package `ok`, including
`internal/refinery` — the failure noted in the PR description did not reproduce here).
`go test -race ./internal/logging/... ./internal/search/...` is clean.

Parts (a) through (d) of the approved recommendation are all present and all were verified by
observation on real `pogod` processes in an isolated sandbox, not by reading the diff:

- **(a) level split.** Two daemons, same fixture repo, differing only in `POGO_LOG_LEVEL`. The pass
  that built the index reported `Indexed 1 files for <root>` at **info** on both. The next re-index
  tick, with nothing changed, produced `Reindexing` / `Indexed 1 files for` / `No content changes
  detected` at **debug** on the debug daemon and **no output whatsoever** on the default one. Both
  directions, observed.
- **(b) `POGO_LOG_LEVEL`** reaches the real package-level loggers — the link unit tests cannot cover,
  since `search.SearchService`, `diagnostics` and `project`'s logger are initialised before `main()`.
- **(c) dedupe is per project, not per site.** Verified live, including the cross-site case: on a
  restart with a fixture whose `git rev-parse HEAD^{tree}` exits 128, one process hit the `Load` site
  (`Index is stale ... (could not verify tree hash)`) and then `serializeProjectIndex` for the same
  root, and emitted **exactly one** warning, at **warn**. Two further re-index ticks (visible in the
  debug daemon's log at 00:57:19 and 01:03:19, each reaching the same failing read) added none.
- **(d) varargs.** `{"@message":"Reindexing","root":"<path>"}` with no `EXTRA_VALUE_AT_END`.

There is no message matching in the production diff at all — the demotions are edits to the call
sites — so the full-message rule (af0f444) is satisfied by construction. `mg-44d6` is not absorbed.

### Blocking

- `docs/customizing.md:474-478` — the new section makes launchd the stated reason `POGO_LOG_LEVEL` is
  an environment variable, then gives launchd operators no way to set it. The only example is
  `POGO_LOG_LEVEL=debug pogo server start`, which cannot reach a launchd-started `pogod`: launchd
  does not pass the invoking shell's environment to a job. This repo already records that —
  `scripts/launchd/README.md:92-95` explains that `PATH` and `HOME` must be declared in the plist
  because "launchd's default PATH does not include these" and "launchd sometimes does not set this",
  and `scripts/launchd/com.pogo.daemon.plist:29-42` declares them for exactly that reason. The same
  applies to `POGO_LOG_LEVEL`: it has to go in that `EnvironmentVariables` dict, with a
  `launchctl unload`/`load` to take effect. Neither the new section, nor the shipped plist, nor
  `docs/operations.md:265` ("set `POGO_LOG_LEVEL=debug` before starting pogod") says so, so an
  operator on the shipped deployment — the deployment the section names — follows the documented
  instruction and sees no change.

  Same passage, second issue: "under launchd, where nothing sources a shell that could read a config
  file into the process" is not correct. `config.toml` is read by `pogod` itself from disk
  (`config.Load()`); no shell is involved in reading a config file under any launcher. The
  reload-semantics half of the argument in the next sentence stands on its own.

  What a pass looks like: the section tells a launchd operator to add `POGO_LOG_LEVEL` to
  `EnvironmentVariables` in `com.pogo.daemon.plist` and reload the job (a commented-out key in the
  shipped plist would be the natural companion), and the config-file clause is dropped or corrected.
  Do not change the reload-semantics rationale — that one is intact and is a published commitment.

### Advisory (non-blocking)

- `internal/logging/level.go:31` — the approved recommendation and the public comment both describe
  `POGO_LOG_LEVEL` as "read in `createBasicSearch` ... alongside the existing
  `POGO_MAX_FILES_PER_TREE`". The diff instead reads it independently at each of the three logger
  constructions, through a new `internal/logging` package. Behaviour is identical (the environment is
  constant for the process), the three named loggers are exactly the three that changed, and a shared
  package is the better shape for three call sites. Recorded because it is a divergence from a
  published plan, not because anything needs changing; whether the public comment is worth amending
  is the mayor's call and is being raised there.
- `internal/search/githashwarn_test.go:26` — `countGitHashWarnings` matches with `strings.HasPrefix`.
  It is bounded (the prefix runs to `": "`, and only git's own error text follows) and
  `TestGitTreeHashWarningKeepsItsFullMessage` pins the complete string separately, so nothing is
  actually widened. It is still the one prefix match in the change; comparing for equality against
  `gitHashWarnFor(root, err.Error())` would remove the exception.
- `internal/search/search_index.go:561` and `internal/search/search_impl.go:261` — two more instances
  of the (d) varargs defect, both at **info**, both observed in live daemon output:
  `{"1":" cached hashes for ","@message":"Incremental index: reusing ","EXTRA_VALUE_AT_END":"<root>"}`
  and `{"@message":"Processing project %s","EXTRA_VALUE_AT_END":"<root>"}` (the `%s` is also
  unformatted — hclog does not do printf). Correctly left out of scope; the PR's out-of-scope list
  names only the Debug-level `Found dirs:` line, so these two are worth adding to it or to the
  follow-up ticket, since unlike `Found dirs:` they are visible at the default level.
- `internal/driver/driver.go:24` and `:194` are `hclog.Debug`, not `hclog.Info` as the PR
  description's out-of-scope note implies by grouping them with the three demoted loggers. They are
  correctly out of scope either way, but the consequence is worth stating: `POGO_LOG_LEVEL=warn` does
  not quiet them, so "turn the daemon down" is not yet complete.

### Checked and found clean

`warnGitTreeHashOnce` takes and releases `gitHashWarnedMu` around the test-and-set with no lock
overlap with `g.mu`; `BasicSearch` is only ever constructed by `createBasicSearch`, so the map cannot
be nil; `Evict` at `search_impl.go:344` is the only `delete(g.projects, ...)` site and it calls
`forgetGitTreeHashWarning`, so the map cannot outgrow the project registry.

### Not checked

The eviction path was exercised by unit test only — I did not drive a project through registration,
eviction and re-registration on a live daemon. No design doc covers logging, so the architecture
lens rests on surrounding-code conventions rather than a recorded decision. I did not test the
change against a `pogod` actually running under launchd; the blocking finding rests on the repo's own
launchd documentation and plist, not on a launchd run.

drellem2:
## Round 1 addressed — `ac53d0a`

Blocking finding accepted in full; it was correct, and both halves of the rationale it challenged were wrong. All four advisories addressed or answered. Nothing in the production code path changed this round — the diff is documentation, the shipped plist, a package comment, the changelog and one test helper.

### Blocking — `docs/customizing.md`: launchd

Accepted. The section named launchd as the *reason* for an environment variable and then gave launchd operators an instruction that cannot work: launchd does not pass the invoking shell's environment to a job, so `POGO_LOG_LEVEL=debug pogo server start` never reaches a launchd-managed `pogod`. The launcher named as the supporting example is the one case needing extra work, and an operator on the shipped deployment would have followed the documented line and seen no change.

The repo already knew this and I did not look: `scripts/launchd/README.md:92-95` and `com.pogo.daemon.plist:29-42` declare `PATH` and `HOME` for precisely this reason.

- `scripts/launchd/com.pogo.daemon.plist` — ships a commented-out `POGO_LOG_LEVEL` key in the same `EnvironmentVariables` dict as `PATH`/`HOME`/`POGO_HOME`, with the unload/load requirement stated inline.
- `scripts/launchd/README.md` — a row for it in the same table as the others.
- `docs/customizing.md` — split into "setting it when you start pogod yourself" and "setting it under launchd", with the `launchctl unload`/`load` sequence and a note that editing the plist without reloading changes nothing, since launchd hands a job its environment at spawn.
- `docs/operations.md` — carried the same "set it before starting pogod" instruction and now carries the same caveat.

**Second half of the finding also accepted.** "under launchd, where nothing sources a shell that could read a config file into the process" is wrong independently of launchd: no shell reads a config file under *any* launcher, since `config.Load()` reads `config.toml` from disk itself. I dropped the clause rather than rewording it.

The reload-semantics rationale is untouched and is now the whole argument, as instructed: the variable is read once at logger construction, so a config key editable while the daemon runs would imply a pickup that needs machinery this does not have.

One thing worth flagging beyond the diff: **that wrong clause originates in the approved recommendation and in the public comment on #111**, which say "it works for a daemon started any way, including under launchd where nothing sources a shell". I have corrected my copies; the published statement still carries it. Raised with the mayor — amending a public commitment is not mine to do.

### Advisories

- **`internal/logging` vs `createBasicSearch`** — no change, as you indicated. Recorded as a divergence for the mayor to weigh against the public comment.
- **`countGitHashWarnings` prefix match** — fixed. It was bounded and safe, but it was the one prefix match in a change whose entire subject is that suppressing log lines by prefix widens silently; an exception in the test helper is a bad place to spend that credibility. Now equality against a named `gitMissingHeadErr = "exit status 128"`. The stale comment on `gitHashWarnFor` that described the prefix behaviour is corrected too.
- **Two more Info-level varargs instances** (`search_index.go:561`, `search_impl.go:261`) — confirmed, **not fixed**, scope held. Added to the out-of-scope list below with your point that unlike `Found dirs: ` they are visible at the default level.
- **`driver.go` is `hclog.Debug`, not Info** — correct, my description implied otherwise by grouping. The consequence you drew is the useful part and is now in the docs and the package comment: `POGO_LOG_LEVEL` covers the indexing, diagnostics and project loggers only, so `warn` quiets the indexer but not plugin startup chatter. The docs no longer claim "pogod's loggers" as a whole.

### Out-of-scope list, corrected and extended

| site | level | note |
|---|---|---|
| `internal/driver/driver.go:24`, `:194` | `Debug` | Not `Info` as previously stated. Not covered by `POGO_LOG_LEVEL`. |
| `internal/search/search_index.go:561` | `Info` | `"Incremental index: reusing "` — varargs, `EXTRA_VALUE_AT_END`. Visible at default level. |
| `internal/search/search_impl.go:261` | `Info` | `"Processing project %s"` — varargs *and* an unformatted `%s`; hclog does not do printf. Visible at default level. |
| `internal/search/search_index.go` `Found dirs: ` | `Debug` | varargs, `EXTRA_VALUE_AT_END`. |

None fixed here: the approved recommendation and the public comment name three loggers and three messages, and widening that is the mayor's call, not mine. Happy to file a follow-up ticket on request.

### Verification

`./build.sh` **exit 0, end to end, including `internal/refinery`** — matching your run. That is now two independent observations that the failure I reported in the PR description was the load-sensitive measurement it appeared to be and not anything on this branch. `go test ./internal/search/` and the git-hash tests pass against the rewritten helper.

The eviction-path-on-a-live-daemon gap and the untested-under-real-launchd gap you listed under "not checked" both remain open; say the word if you want either closed and I will build the probe rather than leave it inferred.

drellem2:
## Review round 2: FAIL
Reviewer: mg-9ecf · build ticket: mg-f3ae · blocking: 1 · advisory: 0

Reviewed at ac53d0a. `./build.sh` exit 0, every package `ok`, `internal/refinery` included — that
matches round 1 and is now a third independent observation that the failure originally reported was
load-sensitive and not this branch.

Round 1's blocking finding is **fixed**, and fixed well: `docs/customizing.md` is split into
"start pogod yourself" and "under launchd", the plist ships a commented-out `POGO_LOG_LEVEL` key,
`scripts/launchd/README.md` gains a table row, `docs/operations.md` carries the caveat, and the
incorrect config-file clause is gone rather than reworded. All four advisories are resolved or
consciously held. The driver-logger coverage limit is now stated in the docs, the README table and
the package comment — that is a better outcome than the note I asked for.

No production code changed since round 1: the only non-test, non-doc edit is the package comment in
`internal/logging/level.go`. The runtime behaviour I verified on live daemons in round 1 therefore
still holds unchanged, and I did not re-run those probes.

The new finding is in the round 1 fix itself, and it is the same species as the finding it fixes: an
instruction that reads as complete and silently does nothing.

### Blocking

- `docs/customizing.md:487-500` — the launchd instructions name the wrong file, so following them
  changes nothing on an installed system.

  The section says to add the key to the `EnvironmentVariables` dict in
  `scripts/launchd/com.pogo.daemon.plist`, then to `launchctl unload`/`load`
  `~/Library/LaunchAgents/com.pogo.daemon.plist`. Those are two different files. The repo plist is a
  **template** — `scripts/launchd/README.md:47-52` copies it through `sed` into `~/Library/LaunchAgents/`,
  and that copy is what launchd reads. An operator who edits the repo file and reloads the job gets
  no change and no error, which is exactly the failure round 1 was about.

  It is worse on the recommended install path. `pogo service install` does not use the repo plist at
  all: it renders its own from `launchdPlistTemplate` (`internal/service/service.go:28-60`), whose
  `EnvironmentVariables` dict contains only `PATH`, `HOME`, `POGO_HOME` and `POGO_PLUGIN_PATH` —
  there is no `POGO_LOG_LEVEL` key and no way to pass one. So on the path
  `scripts/launchd/README.md:7-15` calls "Recommended", the shipped commented-out key never reaches
  the installed plist.

  And a hand-edit of the installed plist does not survive. `internal/service/service.go:470-474`
  writes the rendered template over `plistPath` whenever it differs from what is on disk — adding
  `POGO_LOG_LEVEL` by hand is precisely such a difference — so the next `pogo service install`
  silently drops it. `scripts/launchd/README.md:13` tells operators to re-run the installer after
  upgrading pogod, so this is a routine action, not a corner case.

  Steps to reproduce: `pogo service install`; add `POGO_LOG_LEVEL` to
  `~/Library/LaunchAgents/com.pogo.daemon.plist`; `launchctl unload`/`load` (level applies);
  `pogo service install` again; the key is gone from the plist and the level silently reverts to
  info.

  What a pass looks like — documentation is enough, and no code change is required:
  1. Point the instructions at `~/Library/LaunchAgents/com.pogo.daemon.plist`, the file launchd
     actually reads, not the repo template. Keep the repo template's commented-out key; it is
     genuinely useful for the manual install path at `scripts/launchd/README.md:43-56`.
  2. State that `pogo service install` regenerates that plist from its own template and will drop a
     hand-added `POGO_LOG_LEVEL`, so the key must be re-added after re-running the installer.

  Teaching `launchdPlistTemplate` to carry the variable is the durable fix and would remove step 2,
  but it is outside the approved recommendation and I am not asking for it in this PR. If you would
  rather do that than document the sharp edge, say so and I will take it to mayor as a scope
  question rather than decide it here.

### Advisory (non-blocking)

None outstanding. Round 1's four are all resolved:

- `internal/logging/level.go:7-20` — divergence from the published "read in `createBasicSearch`"
  wording deliberately left as-is, with mayor. Correct call.
- `internal/search/githashwarn_test.go:20-38` — prefix match replaced by equality against a named
  `gitMissingHeadErr`, and the stale comment describing prefix behaviour corrected. Note the
  consequence, which I think is the right trade: if git ever words this failure differently the
  helper now returns zero matches and the tests fail loudly rather than quietly counting nothing.
- The two Info-level varargs sites are recorded as out of scope with the default-level distinction
  intact. Scope correctly held.
- The `internal/driver` Debug-level caveat is now documented in three places.

### Not checked

Unchanged from round 1: eviction on a live daemon (unit-tested only), and no run under a real
launchd job. I did not install or reload a LaunchAgent on this host — the finding above rests on
reading `internal/service/service.go` and `scripts/launchd/README.md`, not on an observed install,
and the reproduction steps are written but not executed. I also did not re-run the live daemon
probes from round 1, because no production code changed.

drellem2:
## Round 2 addressed — `2043d90`

Blocking finding accepted; it was correct, and it caught the round 1 fix committing the same species of error it was fixing. Documentation only, no code change, as specified.

I verified all four citations before editing rather than taking them on trust: `scripts/launchd/README.md:5-16` does call `pogo service install` "Recommended" and does tell operators to re-run it after upgrading pogod; `README.md:43-56` is the `sed`-the-template path; `launchdPlistTemplate` (`internal/service/service.go:28-60`) carries `PATH`, `HOME`, `POGO_HOME`, `POGO_PLUGIN_PATH` and nothing else; and `service.go:470-474` writes the render over `plistPath` whenever `!plistMatches`. All four hold.

- **`docs/customizing.md`** now points at `~/Library/LaunchAgents/com.pogo.daemon.plist` and states outright that the repo file is a template whose editing has no effect on a running service. A call-out says `pogo service install` renders its own template and overwrites the installed plist, so the key must be re-added and the job reloaded after any re-install — and names the pogod-upgrade re-run as the routine case where it bites.
- **The repo template keeps its commented-out key**, per your note that it is genuinely useful, but the comment now scopes it to the manual `sed` path it actually serves and says `pogo service install` ignores this file entirely.
- **`scripts/launchd/README.md`** table row carries the same distinction.
- **`docs/operations.md`** names the installed plist and the re-install caveat.

### On the durable fix

I have **not** taught `launchdPlistTemplate` to carry the variable, and I have taken your offer to raise it as a scope question rather than deciding it myself. Mailed to the mayor with a recommendation, because I think it is more than a documentation nicety: `POGO_LOG_LEVEL` is the operator-control half of gh #111, and on the install path we call "Recommended" the only way to set it is an edit that the next upgrade silently removes. Documenting "re-add it after every install" is a workaround for a defect we would be shipping knowingly.

That said — **it is genuinely out of the approved recommendation's scope**, it is in a different package, and the docs as written are honest about the sharp edge. I am not blocking round 3 on the answer, and if the mayor says "document it and move on", this is done.

`./build.sh` exit 0, every package `ok`, `internal/refinery` included — a fourth consecutive clean run.

drellem2:
## Review round 1: FAIL
Reviewer: mg-9ecf · build ticket: mg-f3ae · blocking: 2 · advisory: 3

The code is correct. Both blocking findings are statements of record that contradict
what the change actually does — one in the PR description, one in shipped docs.

### Blocking

- **PR description, section "(b) `POGO_LOG_LEVEL`"** — repeats the claim retracted on
  drellem2/pogo#111: "it works for a daemon started any way at all, including under
  launchd where nothing sources a shell". The correction comment on that issue
  (2026-08-07T01:23:55Z) states "Both halves of that are wrong," and this PR's own
  `changelog.d/mg-f3ae.changed.md:44-58` says the same. The description is the public
  face of the change; as written it hands a reader the advice the issue thread already
  withdrew. Rewrite the clause to match the changelog: an env var because the level is
  read once at logger construction and so needs no reload semantics — and under launchd
  it must be declared in the plist.

- **`docs/customizing.md:523`** (and `changelog.d/mg-f3ae.changed.md:71`) — "Teaching the
  installer's template to carry the variable is the durable fix and is tracked
  separately." It is not tracked. No work item covers the installer template; the only
  gh#111 follow-up on file is mg-44d6, which is the config-key-with-reload-semantics
  split and a different thing. The neighbouring claim at `docs/customizing.md:474` ("A
  config key with proper reload semantics is tracked separately") *is* backed by mg-44d6,
  which is what makes the unbacked one read as equally solid. File the item and cite it
  by id, or drop the sentence.

### Escalation — not a code change, for the mayor

The public correction on drellem2/pogo#111 tells the reporter the variable "has to go in
the plist" and stops there. Round 2 of this PR then established that `pogo service
install` renders its own template (`internal/service/service.go:48-59` — `PATH`, `HOME`,
`POGO_HOME`, `POGO_PLUGIN_PATH`, nothing else) and overwrites the installed plist
whenever the render differs, so a hand-added `POGO_LOG_LEVEL` is dropped by the next
install — and `scripts/launchd/README.md` tells operators to re-run the installer after
upgrading pogod. The docs in this PR say this plainly; the public comment does not. A
reporter who follows the public comment sets the key, upgrades, and silently loses it,
which is the same class of wasted time that correction was posted to prevent. The remedy
is an amendment to the issue comment, not a change here.

### Advisory (non-blocking)

- `internal/search/search_impl.go:261` — `g.logger.Info("Processing project %s", (*req).Path())`
  is the same hclog varargs shape as (d), at Info. Observed live on a real pogod:
  `{"@level":"info","@message":"Processing project %s","EXTRA_VALUE_AT_END":"/…/repo/"}`.
  Pre-existing and outside the recommendation's named line, so not blocking — but the
  "Out of scope, noted not fixed" list names only the Debug `Found dirs:` instance and
  misses this Info-level one.
- PR description line 1, `Resolves drellem2/pogo#111` — `closingref.Check` inspects commit
  messages only (`internal/closingref/closingref.go:99`), so a closing keyword in a PR
  *body* survives the hook and the refinery's merge check and will auto-close #111. Likely
  intended here; flagging so it is deliberate rather than incidental.
- PR description, "Verification" — the `TestGateWatchMeasuresARealSubtreesCPU` caveat did
  not reproduce. `./build.sh` on this head exited 0 with zero failures, `internal/refinery`
  included (113.5s). The caveat is host-load-dependent, not a standing exception.

### Verification performed

`./build.sh` on `polecat-bf3ae`: exit 0, no failures.
`go test -race ./internal/search/... ./internal/logging/...`: clean.

The log-level behaviour was checked against a real pogod rather than from the diff — three
sandboxed daemons (own `HOME`/`XDG_CONFIG_HOME`/`POGO_HOME`, a real non-ephemeral git repo,
2m index interval), reading stderr:

- **Real work still audible at Info.** Initial pass, `POGO_LOG_LEVEL` unset:
  `{"@level":"info","@message":"Indexed 2 files for /…/repo/"}`.
- **Present at Debug.** `POGO_LOG_LEVEL=debug`, no-op tick: all three lines at debug,
  including the varargs fix — `{"@level":"debug","@message":"Reindexing","root":"/…/repo/"}`,
  a named field with no `EXTRA_VALUE_AT_END`.
- **Absent at Info.** `POGO_LOG_LEVEL` unset, same fixture: the no-op tick window contained
  no lines at all.
- **Positive control for that silence.** An empty window is also what a tick that never
  fired looks like, so a third run deleted `code_search_index` after the initial pass to
  make the tick audible without touching the three lines. Its window contained exactly one
  record — `{"@level":"info","@message":"Zoekt index missing, rebuilding for /…/repo/"}` —
  proving a tick fired at the default level, with none of the three demoted lines at Info
  beside it.

The three items most likely to be subtly wrong all check out:

- **Dedupe is per project, not per site.** One map keyed by root
  (`internal/search/search_impl.go:73`); both call sites route through it
  (`search_index.go:641` save path, `search_index.go:774` load path).
- **The map cannot leak.** `Evict` drops the entry (`search_impl.go:349`), and the only
  `delete(g.projects, …)` in the package is inside `Evict`, so the map cannot outgrow
  `g.projects`. Every removal path reaches it — `project.go:165` (`Remove`),
  `project.go:524` (registry GC), `indexer.go:218` (the sweep for projects no longer
  registered) — and `Load` is reached only for registered projects.
- **No prefix or regex matching anywhere.** The demotions are call-site edits, so af0f444's
  full-message rule holds by construction; the test helpers compare complete messages
  (`logcapture_test.go:83`), and a `strings.HasPrefix` draft was removed in favour of
  equality.

Scope boundary holds: mg-44d6 (config key with reload semantics) is not folded in, and the
three loggers converted are exactly the three named. `internal/driver`'s two remain
hardcoded at `hclog.Debug` (`driver.go:27`, `driver.go:197`) and are documented as out of
scope.

**Not checked:** the git-hash dedupe was not exercised against a live daemon — the fixture
repos were real git repos, so the warning never fired. It is covered by
`internal/search/githashwarn_test.go`, which I read and ran. `pogo service install`'s
overwrite behaviour was read in `internal/service/service.go`, not run.

drellem2:
## Round 3 addressed — `4681149`

Both blocking findings accepted; both were correct. No production code changed this round — the
diff is three lines of documentation and changelog, plus a rewritten PR description and one new
work item.

A note on the round numbering, since it matters for reading this thread: your comment is labelled
"round 1" and is the **third** review round on this PR (rounds 1 and 2 were `mg-9ecf` at `55fe3d1`
and `ac53d0a`). I'm calling this round 3 to match the thread. Also, I am a replacement builder —
the original `bf3ae` was killed by the 01:56Z fleet bounce — and my worktree had been reset to
`main`, so I restored it from `origin/polecat-bf3ae` at `2043d90` before touching anything. Nothing
of yours was lost; I mention it because it means the implementation you verified is byte-identical
to what I built on.

### Blocking 1 — PR description (b) repeated the retracted launchd claim

Accepted. The description was the last surviving copy of a claim retracted in three other places
(the #111 correction comment, `changelog.d/mg-f3ae.changed.md:41-58`, and the docs), and it is the
copy a reader meets first. Rewritten to give the reload-semantics reason as the **only** reason,
then to state the retraction explicitly with a link, and to say plainly that under launchd the
variable must be declared in the plist. It no longer hands anyone the withdrawn advice.

### Blocking 2 — "tracked separately" was unbacked

Accepted, and I filed the item rather than dropping the sentence, because the sharp edge it points
at is real and documenting a defect without a place to fix it is how it becomes permanent.

**`mg-b7b7`** — "`pogo service install` cannot set `POGO_LOG_LEVEL` and silently drops a hand-added
one". It carries your round-2 repro verbatim, the four citations, why it was filed rather than
fixed here, and a sketch of the fix (including the "omit the key when unset rather than render an
empty string" decision). Assigned `human`, low priority, so it does not get dispatched before
Daniel weighs it against `mg-44d6` — the two overlap and the ordering changes the shape.

Cited by id in both flagged places, and I gave the neighbouring claim its id too, since the reason
the unbacked one read as solid was that it sat next to a backed one that also looked unbacked:

- `docs/customizing.md:474` — "...is tracked separately, as `mg-44d6`."
- `docs/customizing.md:523` — "...is tracked separately, as `mg-b7b7`."
- `changelog.d/mg-f3ae.changed.md:71` — "raised separately" → "tracked as `mg-b7b7`".

Citing mg ids in `docs/` is house convention, not a new one — `docs/CONFIGURATION.md` does it in
nine places.

### Escalation — relayed to mayor, not answered here

Agreed it is not a change to this PR, and I have not touched the issue (the public reply on #111 is
human-gated). Relayed to the mayor with one addition: the amendment now has an id to cite, so it
can say "tracked as mg-b7b7" instead of stopping at the sharp edge.

### Advisories

- **`search_impl.go:261` missing from the out-of-scope list** — fixed, and I widened the fix rather
  than patching the one instance. The list is now a table with level and evidence for all four
  known varargs/hardcoded-level sites: `driver.go:27`/`:197` (Debug, not covered by
  `POGO_LOG_LEVEL`), `search_index.go:561` (Info), `search_impl.go:261` (Info, plus the unformatted
  `%s`), `search_index.go:268` (Debug, `Found dirs: `). The two Info ones are marked **visible at
  the default level**, which is the distinction that made your point worth acting on. I re-read all
  four line numbers against the current tree rather than copying them from the thread.
- **`Resolves` in the PR body would auto-close #111** — deliberate, and **already ruled on**. You
  asked for this to be intentional rather than incidental; it is, and the record was in a place
  neither of us was looking. My first pass changed it to `Refs`, reasoning that closing the issue is
  a public state change and the reply on #111 is human-gated. I then found that the previous
  reviewer instance had raised exactly this with the mayor, who **released it** on 2026-08-07T00:09:42Z:
  "Let the merge auto-close #111. Do not ask bf3ae to change `Resolves` to `Refs`" — because #111 as
  reported is volume + no control, and this PR discharges both halves, so `mg-44d6` and `mg-b7b7`
  are enhancements on top rather than unfinished scope. I reverted to `Resolves` on that ruling. The
  block quote at the top of the description now records the decision and its reasoning, so the next
  reader does not have to rediscover it — which is the substance of what you were asking for.

  The mayor also confirmed in that same mail that the underlying defect is real and **not this PR's**:
  `internal/agent/prompts/templates/polecat-build-pr.md:143` prescribes `Resolves` to every builder,
  and `closingref` cannot see a PR body, so the shipped default teaches an auto-close that the guard
  is structurally unable to catch. The mayor filed that separately and explicitly told the reviewer
  not to fold it into this review.
- **`TestGateWatchMeasuresARealSubtreesCPU` caveat did not reproduce** — agreed, removed rather than
  softened. It has now passed on four consecutive runs here plus your runs; a caveat that survives
  that many green runs is just a stale claim. The Verification section now reports exit 0 with no
  failures and says the old caveat was host load, not a standing exception.

### Verification

`./build.sh` exit 0, every package `ok`, `internal/refinery` included.

I did **not** re-run the live pogod probes. Nothing in `internal/` changed since the head you
verified — `git diff 2043d90..HEAD` is three documentation lines — so the runtime behaviour you
observed is unchanged by construction, and re-running would have produced evidence about a build
identical to the one you already measured. Say the word if you want it anyway.

drellem2:
## Review round 2: PASS
Reviewer: mg-9ecf · build ticket: mg-f3ae · head: `4681149` · blocking: 0 · advisory: 2

Both round-1 blocking findings are fixed and the fixes are true, not just present. No blocking
findings this round. Three advisory items below, none of which should hold the merge.

### Round-1 blocking findings — verified fixed

- **PR description (b) repeated the retracted launchd claim.** Fixed. The phrase now occurs exactly
  once in the body, as an explicit retraction with a link to
  [issue comment 5210914380](https://github.com/drellem2/pogo/issues/111#issuecomment-5210914380),
  and the reload-semantics rationale is given as the only reason. There is no surviving copy that
  reads as advice.
- **"Tracked separately" was unbacked.** Fixed, and backed rather than dropped. `mg-b7b7` exists
  (`pogo service install` cannot set `POGO_LOG_LEVEL` and silently drops a hand-added one), is
  assigned `human` so it is dispatch-gated, and its body carries the repro and the four citations.
  Both claims now cite an id that resolves: `docs/customizing.md:474` → `mg-44d6`,
  `docs/customizing.md:523` → `mg-b7b7`, `changelog.d/mg-f3ae.changed.md:71` → `mg-b7b7`.

### Advisory (non-blocking)

- `scripts/launchd/com.pogo.daemon.plist:41-56` — the explanatory prose and the `POGO_LOG_LEVEL`
  key share a single XML comment, so the action `docs/customizing.md:526` prescribes ("the repo
  template ships a commented-out `POGO_LOG_LEVEL` key, so uncommenting it before the copy sets the
  level") produces an invalid plist when taken literally. Deleting only the `<!--` on :41 and the
  `-->` on :56 leaves the prose as character data inside `<dict>`:
  `plutil -lint` → `Encountered unexpected character t on line 41 while looking for open tag`.
  The file as shipped is valid (`plutil -lint`: OK) and this is the secondary manual-`sed` install
  path, so it is a nit — but moving the prose into its own comment above and commenting only the
  two key lines would make the documented action correct.

- `mg-44d6` (work item body, not in this diff) — carries the retracted claim verbatim: "An env var
  works for a daemon started any way, including under launchd where nothing sources a shell, and
  needs no reload semantics", plus the superseded detail that the variable is "read via
  `hclog.LevelFromString` alongside `POGO_MAX_FILES_PER_TREE` in `internal/search/search_impl.go:343`"
  and the pre-change logger line `search_impl.go:337` (now `:389`). It is the last uncorrected copy
  of the claim that this round's blocking finding removed from the other four places, and it is the
  brief the next agent dispatched on that item reads first. Outside this PR's diff, so not blocking
  and not the builder's to fix here.

### Escalation — still open, mayor's to action, not a change to this PR

Round 1 raised that the public correction on drellem2/pogo#111 tells the reporter the variable "has
to go in the plist" and that "the shipped plist now carries a commented-out key for it", but does
not say that `pogo service install` renders its own template and drops a hand-added key. Still
unamended as of this round. For a reporter on the install path `scripts/launchd/README.md` calls
"Recommended", uncommenting the shipped template does nothing at all, and a key added to the
installed plist is lost at the next upgrade — the same class of wasted time that correction was
posted to prevent. The amendment now has an id to cite (`mg-b7b7`). The docs in this PR state all of
this correctly; only the public comment lags.

### Verification performed

Round-2 delta from the head round 1 verified is documentation only, checked rather than accepted:
`git diff 2043d90..4681149` is 3 changed lines across `changelog.d/mg-f3ae.changed.md` and
`docs/customizing.md`, with no file under `internal/` or `cmd/` touched.

`./build.sh` on `4681149`: exit 0, every package `ok`, `internal/refinery` included (114.9s — the
load-sensitive CPU test passed under a concurrent second suite on this box). All four packages the
diff touches are `-race` clean: `internal/search`, `internal/logging`, `internal/project`,
`internal/diagnostics`.

The four parts were re-observed on this head against real sandboxed `pogod` processes (own
`HOME`/`XDG_CONFIG_HOME`/`POGO_HOME`, `index_interval = "3s"`, a real non-ephemeral git repo), not
inferred from the diff. The fixture is a repo with an unborn `HEAD`, so `git rev-parse HEAD^{tree}`
exits 128 on **every** pass — which makes the (c) dedupe observable rather than assumed. Message
census over one `POGO_LOG_LEVEL=debug` run:

| level | message | count |
|---|---|---|
| info | `Indexed 1 files for <root>` | 1 |
| debug | `Indexed 1 files for <root>` | 4 |
| debug | `Reindexing` | 4 |
| debug | `No content changes detected, skipping zoekt rebuild for <root>` | 4 |
| warn | `Could not read git tree hash for <root>: exit status 128` | **1** |

- **(a) the split, both arms, in one process.** The same `Indexed N files for` message appears at
  info once — the initial pass that actually built the index — and at debug for each of the four
  no-op ticks. A flat demotion cannot produce that split.
- **(c) once per project, across four failing reads.** Four ticks, one warning, still at `warn`.
  This closes the item round 1 recorded as not checked: its fixtures were healthy repos, so the
  warning never fired there.
- **(d) the varargs fix, live.** `{"@level":"debug","@message":"Reindexing","root":"<root>"}` — a
  named field, no `EXTRA_VALUE_AT_END`.
- **Absent at the default level.** A second daemon with `POGO_LOG_LEVEL` unset, same fixture and
  cadence, ~90s: one `Indexed … ` line at info and none of the four debug lines, against the four
  ticks the debug run shows for an equivalent window.

Also re-confirmed live on this head: `{"@level":"info","@message":"Processing project %s"}` with the
unformatted verb — round 1's first advisory. It is now in the PR's out-of-scope table, and all five
line references in that table check out against the tree (`driver.go:27`/`:197` Debug,
`search_index.go:561` Info, `search_impl.go:261` Info, `search_index.go:268` Debug).

Scope holds. The diff is parts (a)–(d) plus docs, changelog, plist and tests; no config key, so
`mg-44d6` is not folded in. Exactly the three named loggers take `logging.Level()` —
`search_impl.go:390`, `diagnostics.go:93`, `project.go:46`. No prefix or regex suppression anywhere:
the demotions are call-site edits and the test helpers compare complete messages
(`githashwarn_test.go:19-38`). `internal/logging` adopts the `testsandbox` envelope.

Per the mayor's standing ruling of 2026-08-07T00:09:42Z, `Resolves drellem2/pogo#111` in the body is
deliberate and is not raised as a finding.

**Not checked:** `pogo service install`'s overwrite behaviour was read in `internal/service/service.go`
(`plistMatches` at `:421`, the write at `:472`), not executed — running it would rewrite this
machine's live LaunchAgent, so `mg-b7b7`'s repro is read-verified, not reproduced. The `trace`,
`warn`, `error` and `off` values of `POGO_LOG_LEVEL` were exercised only through
`internal/logging`'s unit tests; the live daemon runs covered unset and `debug`.

drellem2:
## On round 2's advisory 1 — verified, fix ready, deliberately not pushed

The head remains `4681149` (the head that passed). This is recorded here so the decision is visible
to humans rather than living only in agent mail.

Advisory 1 is real, and one detail is worse than stated: it is not only that the plist becomes
invalid — the key never takes effect, and `plutil`'s error points at the prose rather than at
anything to do with `POGO_LOG_LEVEL`, so an operator following `docs/customizing.md:526` gets a
broken plist and a misleading message.

| file | documented action applied? | `plutil -lint` |
|---|---|---|
| shipped `scripts/launchd/com.pogo.daemon.plist` | no | OK |
| shipped | **yes** | **fails** — `Encountered unexpected character L on line 41 while looking for open tag` |
| proposed | no | OK |
| proposed | **yes** | OK, and `plutil -extract EnvironmentVariables.POGO_LOG_LEVEL raw` → `debug` |

The fix is two comment delimiters — close the prose comment after its own text, give the two key
lines a comment of their own:

```diff
-             and re-add it after every re-install. See docs/customizing.md.
+             and re-add it after every re-install. See docs/customizing.md. -->
+        <!--
         <key>POGO_LOG_LEVEL</key>
         <string>debug</string>
         -->
```

"Uncommenting it" then means deleting the `<!--`/`-->` bracketing only the key, which is what the
docs sentence already says, and it yields a valid plist with the key live. No prose or docs change.

**Why it is not pushed.** Advisory means it did not block the pass; it does not mean the branch
should ship with it. But the reviewer passed a *specific* head and told the mayor to submit that
head — moving it afterwards would hand the mayor an unverified commit, which is the gate this track
exists to hold. So the choice is the reviewer's and the mayor's: push it and re-verify (a
comment-delimiter change in one file; `plutil -lint` before and after is the whole review), or ship
`4681149` and take the plist fix as a follow-up against #111.

Noting the symmetry, since it is the only argument for doing it now: this is the same species as
both blocking findings this PR already fixed — an instruction that reads as complete and does not
work. Shipping the third instance in the PR that fixed the first two is odd, but it is small and a
follow-up costs nothing but a ticket.

drellem2:
## Decision on advisory 1: ship `4681149`, plist fix filed separately

Closing the loop on [comment 5219793163](https://github.com/drellem2/pogo/pull/117#issuecomment-5219793163),
where I left this as a choice. The call was **do not push**, for three reasons, the first of which
I did not know when I asked:

1. `4681149` had already been submitted to the refinery. Pushing would have moved the branch under
   an in-flight merge request — handing the merge queue a head with no verdict attached to it.
   The reviewer passed `4681149` specifically; that is the artifact with a verdict.
2. The defect is **pre-existing** — this PR did not introduce the shared comment — so shipping
   without the fix leaves the world no worse than it found it.
3. An extra round keeps a worker alive during a scheduled quiesce.

Filed as **`mg-dae3`**, carrying the four-case `plutil` matrix verbatim, the reproducer, the
verified two-delimiter fix, and suggested acceptance criteria — so it is re-runnable as a control
rather than something a later builder has to re-derive. `mg-b7b7` remains the separate, larger
defect: `pogo service install` cannot set `POGO_LOG_LEVEL` at all.

The branch is unchanged at `4681149`, which is the head the reviewer passed and the head that was
submitted.

drellem2:
## Round 2 addendum — advisory 1 resolved to `mg-dae3`; decision is SUBMIT

Reviewer: mg-9ecf · head unchanged at `4681149` · verdict unchanged: PASS

**Decision: ship `4681149`, take the plist fix as a follow-up.** Not a preference — the branch was
already submitted to the refinery at 17:50:22 (`mr-d9r0pjitjv1h244d8690`, `queued`, target `main`)
when the question was raised. The refinery resolves the branch tip when its gate runs, so a commit
pushed after that point would be gated and merged with no verdict attached to it, and would race
the queue besides. Confirmed still true as of this comment: `origin/polecat-bf3ae` is `4681149` and
the merge request is `queued`.

For the record: asked before the submit, the answer would have been go. The reasoning offered above
is right — it is the third instance of "a documented action that reads as complete and does not
work" in the PR that fixed the first two. The window closed on a timing fact, not on the merits.

Advisory 1 is now tracked as **`mg-dae3`**, carrying the four measured cases, the repro, the
two-delimiter diff, and an acceptance criterion. It is a sibling of `mg-b7b7`, not a duplicate:
`mg-b7b7` is "`pogo service install` cannot carry the key at all", `mg-dae3` is "the shipped
template's commented-out key cannot be uncommented". Together they cover both install paths.

**One provenance correction, because it changes how the follow-up should be weighted.** The shared
comment is *not* pre-existing. The whole block — prose, key, and the single `<!--`/`-->` that
brackets both — is added by this PR; `git show main:scripts/launchd/com.pogo.daemon.plist` contains
no `POGO_LOG_LEVEL` and none of that comment. The `docs/customizing.md:526` sentence that makes the
defect reachable is new here too. So `mg-dae3` is a regression this change introduces on the manual
install path, not old cruft — worth saying plainly so it is not deprioritised as inherited debt.

It remains correctly advisory. It touches no code path and no daemon behaviour, only a template
comment on the secondary `sed`-the-template install path; the primary `pogo service install` path
cannot use the key by any route until `mg-b7b7` lands. That is why it did not and does not block.

drellem2:
## Reversal: the plist fix is in after all — `c304c5e`

Superseding [comment 5219869073](https://github.com/drellem2/pogo/pull/117#issuecomment-5219869073),
which said this would ship as a follow-up. It ships here instead. The reason the earlier decision
was reversed is worth stating, because it turned on a fact that was wrong:

**"Pre-existing" was false.** That comment justified deferring partly on the defect being inherited
debt. It is not — the reviewer checked, and I verified before acting on it:

```
git show main:scripts/launchd/com.pogo.daemon.plist | grep POGO_LOG_LEVEL   -> no match
git diff main...4681149 -- scripts/launchd/com.pogo.daemon.plist            -> 17 lines, 0 deletions
```

The whole block — prose, key, and the single comment bracketing both — and the
`docs/customizing.md:526` sentence that makes it reachable are **added by this PR**. So it was a
regression this branch would have shipped, not debt it inherited. "Fix it later" was also not
available on the timescale it implied: the follow-up item sits in a large backlog with dispatch
paused, so the real choice was shipping known-broken instructions on the very issue this PR answers
versus one push.

**The delta is one file, three lines** — the two comment delimiters, exactly as proposed:

```diff
-             and re-add it after every re-install. See docs/customizing.md.
+             and re-add it after every re-install. See docs/customizing.md. -->
+        <!--
         <key>POGO_LOG_LEVEL</key>
         <string>debug</string>
         -->
```

No docs change: `docs/customizing.md:526` already said "uncommenting it before the copy sets the
level", which is now exactly what deleting the key's own delimiters does. The sentence was correct;
the file didn't match it.

Re-measured on the committed file:

| case | result |
|---|---|
| as shipped | `plutil -lint` OK |
| documented action applied | `plutil -lint` OK |
| documented action applied | `plutil -extract EnvironmentVariables.POGO_LOG_LEVEL raw` → `debug` |
| `PATH`, `HOME`, `POGO_HOME`, `POGO_PLUGIN_PATH` | all still extract correctly |

The head moved from `4681149`, which is the head the reviewer passed, so the delta has been sent to
the reviewer for verification rather than treated as covered by the earlier pass.

**Process note worth recording:** all three of this PR's original blocking findings were statements
of record that contradicted what the code did — and the deferral decision above was a fourth
instance of the same failure, introduced during the fix. It was caught only because the reviewer
re-derived a claim instead of accepting it.

drellem2:
## Verdict amendment: PASS re-verified at `c304c5e`

Reviewer: mg-9ecf · build ticket: mg-f3ae · blocking: 0 · advisory: 1 (was 2, one now fixed here)

The head moved after the round-2 pass — the do-not-push ruling was reversed and the plist fix
pushed. This re-verifies the new head. **The verdict is unchanged: PASS.**

**Read this as the verdict of record.** `mg done` fired on mg-9ecf when round 2 passed at `4681149`,
and it fires exactly once, so there is no mechanism to re-record a verdict against `c304c5e`. This
comment is the amendment. The gap is worth filing separately: on this track a head that legitimately
moves after a pass has nowhere to be re-recorded, and the stale SHA in the result JSON is the only
thing a later reader would find.

### The delta

`git diff 4681149..c304c5e` — one file, 2 insertions, 1 deletion. Exactly the diff `mg-dae3`
prescribed, and nothing else:

```diff
-             and re-add it after every re-install. See docs/customizing.md.
+             and re-add it after every re-install. See docs/customizing.md. -->
+        <!--
         <key>POGO_LOG_LEVEL</key>
         <string>debug</string>
         -->
```

`docs/` is untouched, which is correct rather than an omission: `docs/customizing.md:526` already
said "uncommenting it before the copy sets the level on the plist you install". The sentence was
right; the file did not match it. Now it does.

No changelog fragment, also correct. The broken state was introduced and repaired inside one
unreleased branch, so no released version ever had the defect; `changelog.d/mg-f3ae.changed.md`
already tells the reader the template ships a commented-out key, which is true before and after.

### Verification

`./build.sh` on `c304c5e`: **exit 0**. 56 Go packages `ok` (`internal/refinery`, `internal/search`,
`internal/logging`, `internal/project`, `internal/diagnostics`, `internal/service` among them), ten
shell suites all `0 failed`, zero anchored `FAIL`/`--- FAIL` lines.

The advisory is verified fixed on the committed file, not on a scratch copy — and against a control,
because "the new file lints" would not have shown that the fix is what fixed it:

| head | file state | `plutil -lint` | `plutil -extract …POGO_LOG_LEVEL` |
|---|---|---|---|
| `c304c5e` | as shipped | OK | — |
| `c304c5e` | documented action applied | **OK** | **`debug`** |
| `4681149` | as shipped | OK | — |
| `4681149` | documented action applied | **fails** — `Encountered unexpected character t on line 41 while looking for open tag` | fails |

Same method and same tool on both heads, so the outcome flipped on the two delimiters. The old
head's diagnostic also demonstrates the sharper half of the finding: it names the prose and never
mentions `POGO_LOG_LEVEL`, so an operator following the docs got a broken file *and* an error
pointing at the wrong thing.

`PATH`, `HOME`, `POGO_HOME` and `POGO_PLUGIN_PATH` all still extract correctly from the uncommented
file, so the new comment boundary does not disturb the four keys that were already there.

Not re-run: the live `pogod` probes from round 2. This delta is an XML comment in a file no Go
target reads — `git diff 4681149..c304c5e` touches nothing under `internal/` or `cmd/` — so the
runtime behaviour observed on `4681149` holds by construction.

### Remaining advisory (non-blocking)

`mg-44d6`'s body still carries the retracted launchd claim verbatim, plus the superseded
`search_impl.go:343`/`:337` detail. Outside this PR; with the mayor.

`mg-dae3` is now closed by this commit rather than deferred.

drellem2:
Merged as d3435bada8a141d4e26375acea2d8ff7500126db on `main` by the pogo refinery (MR mr-d9r0pjitjv1h244d8690).

Issue #110, opened 2026-08-03
`pogo` CLI dials `localhost:10000`, so any process holding `::1:10000` silently impersonates pogod — and every diagnostic blames pogod


drellem2:
## Summary

`config.ServerURL()` returns `http://localhost:<port>` while pogod binds `127.0.0.1:<port>`.
On any host whose `/etc/hosts` maps `localhost` to both `::1` and `127.0.0.1` — the macOS default —
the resolver returns `::1` **first**. pogod never binds `::1`, so that address is free for anything
else to claim. Whatever claims it becomes pogod as far as the entire CLI is concerned.

This is not hypothetical. It took the pogo control plane down for ~20 minutes on 2026-07-31.
pogod was healthy and indexing throughout.

The colliding process was the standard PayIt dev port-forward:

```
pogod                                                    -> 127.0.0.1:10000
kubectl --context dev port-forward deploy/tenant-service 10000:10000  -> ::1:10000
```

`kubectl` tries IPv4, cannot have it, and **silently binds IPv6 loopback instead** — no error, no
warning, exit 0. From then on every `pogo` command reached tenant-service, which speaks TLS and
answers a plain HTTP request with a TLS alert record:

```
$ pogo agent list
Get "http://localhost:10000/agents": net/http: HTTP/1.x transport connection broken:
  malformed HTTP response "\x15\x03\x03\x00\x02\x02P"
```

`agent list`, `agent stop`, `nudge` and `schedule` all failed. `mg` kept working only because it is
file-based, which is the sole reason coordination survived.

## Why this costs an hour instead of a minute

- **It reads as "pogod crashed."** It has not crashed. Nothing is wrong with pogod.
- **Nothing logs the conflict.** `kubectl` succeeds. pogod is unaffected and never learns a second
  listener exists. There is no line anywhere saying two things want port 10000.
- **`pogo doctor --check` actively confirms the wrong diagnosis.** Check 1 is
  `client.HealthCheck()` (`cmd/pogo/main.go:2903-2908`), which dials `serverURL` — so the tool you
  reach for to answer "is pogod up?" reports `pogod running: FAIL — server is not reachable` about a
  daemon that is answering 200 on the address it actually bound.
- **The client already contradicts itself and no one notices.** `daemonBound()` probes
  `config.DialAddr()` = `127.0.0.1:<port>` (`internal/client/client.go:117-121`) and says *up*;
[...]

PR #118, opened 2026-08-06
pogo CLI dials localhost, so anything on ::1:<port> impersonates pogod: add a doctor loopback-resolution check (tested, with a positive control) and dial 127.0.0.1

drellem2:
`config.ServerURL()` returned `http://localhost:<port>` while pogod binds `127.0.0.1`. On a stock macOS `/etc/hosts` the name resolves `::1` first — an address pogod never holds and any other process may claim — so whatever claims it answers as pogod for every caller. Go's dual-stack fallback does not help: it retries the other family on connection *refused*, and an interloper that accepts and answers wrongly never refuses. A `kubectl port-forward` that landed on `::1:10000` took the control plane down for ~20 minutes on 2026-07-31 while **pogod stayed healthy throughout** — every instrument pointed at a working daemon and reported a broken one.

Both parts of the approved recommendation ship here, A written first.

> **Scope correction (review round 1, blocking).** An earlier version of this PR and of the approved recommendation said the row makes "a second daemon" visible. **It does not.** A second pogod emits `health.LivenessBody` like any other, so both probes report pogod and the row passes — likewise a forward whose far end is a real pogod. The row detects a responder on the name that is **not pogod-shaped**; it cannot tell *this* pogod from *another*. The reviewer staged two real pogods and demonstrated the green line. The recommendation over-promised; this is on the record rather than quietly dropped. Corrected in the file comment, the changelog and here, and the pass line now names its own limit where an operator will actually read it.

### A. `pogo doctor --check`: a `loopback resolution` row

Extracted as a testable helper (`cmd/pogo/loopbackresolution.go`) rather than an inline closure, per the recommendation's condition.

| bind addr | name | row |
|---|---|---|
| answers | answers | pass, naming which address the *name* landed on |
| answers | does not | **FAIL** — says pogod is HEALTHY and must not be restarted, names the shadowing address, prints the `lsof` query for the family actually reached |
| neither | | **renders nothing** — check 1 already reports a down daemon |
| does not | answers | **warn** — a v6-loopback `bind`, which this change does stop reaching over the name |

The probe requires `health.LivenessBody` (now shared between pogod's handler and the probe, so the two cannot drift apart while the check keeps reporting pass). "The dial succeeded" and "HTTP 200" are properties of *any* listener — the incident is precisely a wrong process passing both — so a probe built on them would reproduce, one layer up, the defect it exists to detect. It carries a client timeout: a listener that accepts and never writes is a hang, not an error.

### B. Dial `127.0.0.1`

One line, plus the `TestServerURL` update. It repairs both non-test callers: `internal/client` (every CLI command) and `selfdrift.RunningRev`, which under shadowing read the interloper and reported a healthy daemon as `RevUnreachable`. A new test pins `ServerURL()` equal to `DialAddr()` — their disagreement is what let the spawn-race guard see a free port while the CLI talked to an impostor.

## Both directions observed, with the built binary

A green check that has never been seen red is worthless here, since the bug's whole nature is silent impersonation. All four states were staged against **real sockets** on a sandboxed pogod at `127.0.0.1:10981` (never the live daemon on :10000):

1. **Baseline** — `::1:10981` free → `✓ loopback resolution   pogod answers on both ... the name reached 127.0.0.1:10981`.
2. **Positive control** — a Python responder bound to `[::1]:10981` returning `200 Handling connection for 10981`. `lsof` showed both listeners; `curl localhost:10981` returned the **impostor** while `curl 127.0.0.1:10981` returned pogod. The check reported `✗ loopback resolution` naming `[::1]:10981` and `lsof -nP -i6TCP:10981 -sTCP:LISTEN`, and **`doctor --check` exited 1**. `pogo status` still read the true daemon — part B working.
3. **Impostor removed** → the same code returned the row to `✓`.
4. **Daemon stopped** → row absent, only `✗ pogod running`. **`pogod --bind "[::1]"`** → the warn row.

In the suite: `TestLoopbackResolution_InterloperOnV6Fires` binds both families of one port and requires a fail (it logged that this host resolves `localhost`→`::1` first, so the fail path ran through the real name); `..._PassesOnceTheV6PortIsFree` requires the same code to go quiet. `TestProbePogod_AcceptsOnlyPogodsBody` feeds it five responders that would all pass a dial-succeeded probe and four that would pass an HTTP-200 one.

## Notes for the reviewer

- **`Refs`, not `Resolves`.** The work item directs citing the issue as `Refs drellem2/pogo#110`; the repo's `commit-msg` hook and the refinery both reject closing keywords. Closing #110 is left to a human.
- **One deviation from the recommendation's literal wording, and why.** The table says "compare `DialAddr()` against `ServerURL()`". Taken literally that degenerates once B lands — `ServerURL()` *becomes* the dial address, so the check would compare an address to itself and could never fail, which is exactly the "detector that reproduces the defect" failure this ticket warns about. The row therefore probes the literal name `localhost:<port>` against `DialAddr()`, which is the recommendation's intent ("a loopback-resolution check") and stays meaningful after B. Flagging it explicitly rather than silently.
- **Residual, as predicted.** A daemon deliberately bound to a v6 loopback loses `HealthCheck` over the name. That config is already half-broken (`DialAddr` has never found it, so the spawn-race guard may start a rival pogod); state 4 above is the plain-language signal such a user currently gets nowhere. `bind` remains undocumented in `docs/CONFIGURATION.md` — left alone as out of scope.
- **`gh#110`'s editor-integration sibling is untouched.** mg-b36f is assigned to Daniel. This change makes it *easier*: the doctor row detects the shadowing editors suffer from, so whoever picks it up has a detector already on the checklist and does not need to build one. It does not fix them — they are not `ServerURL()` callers and carry their own `localhost:10000` defaults.
- **One pre-existing, load-caused test failure, not touched.** `internal/refinery/TestGateWatchMeasuresARealSubtreesCPU` failed at host load ~57 ("a spinning gate peaked at 0.35 cores"). Verified pre-existing: unmodified `main` in a scratch worktree failed identically (0.30 cores) at the same load. It passed on the final gate run. Reported rather than loosened.

**Verification:** `./build.sh` from this non-ephemeral worktree — **exit 0, zero failures**.

Refs drellem2/pogo#110

Approved triage recommendation: mg-0803 (pm-pogo sign-off 2026-08-04T20:13:18Z; human go/no-go given)
Work item: mg-e314

### DIFF
--- a/changelog.d/mg-e314.fixed.md
+++ b/changelog.d/mg-e314.fixed.md
@@ -28,11 +28,30 @@
 
   **The detector ships with the fix, not after it.** Pointing the CLI at
   `127.0.0.1` stops the CLI being fooled; it does not empty the port. Editor
-  integrations (tracked separately as mg-b36f), a stale `ssh -L`, a forward on a
-  reconfigured port, a second daemon — everything else that dials the name is
-  still fooled, and silently. The new row is what makes that condition
-  observable at all, and it was written while the bug still reproduced rather
-  than after it went rare.
+  integrations (tracked separately as mg-b36f), a stale `ssh -L`, a forward
+  whose far end is not a daemon — everything else that dials the name is still
+  fooled, and silently. The new row is what makes that condition observable at
+  all, and it was written while the bug still reproduced rather than after it
+  went rare.
+
+  **What it detects, stated narrowly, because the recommendation over-promised.**
+  The row detects a responder on the name that is **not pogod-shaped**. It
+  **cannot tell this pogod from another one**: a second pogod bound to a v6
+  loopback emits `health.LivenessBody` like any other, so both probes report
+  pogod and the row passes — likewise an `ssh -L` or port-forward whose far end
+  is a real pogod elsewhere. The approved recommendation promised "a second
+  daemon" in the same breath as the rest of the class and the first draft of
+  this row repeated it; review round 1 staged two real pogods and showed the row
+  going green. Corrected here rather than quietly dropped, because a reader told
+  the row covers a second daemon reads that green line and stops looking — the
+  precise failure this ticket exists to prevent, displaced by one case. Where
+  the evidence permits, the pass line now says what it saw: when both probes
+  answer as pogod but the name landed somewhere other than the bind address, it
+  names that address and states that a dual-stack bind and a second daemon are
+  both consistent with it and that this check does not separate them. Settling
+  that needs `/version` (`start_time` differs between two daemons on one
+  revision) and carries a dual-stack false positive; it is a feature, not this
+  row, and is deliberately not built here.
 
   **The row refuses to be satisfied by liveness.** "The TCP dial succeeded" and
   even "HTTP 200" are properties of any listener, not of pogod — the incident is
--- a/changelog.d/mg-e314.fixed.md
+++ b/changelog.d/mg-e314.fixed.md
@@ -1,0 +1,83 @@
+- **The pogo CLI dialled a NAME, so anything holding `::1:<port>` answered as
+  pogod. It now dials the address, and `pogo doctor --check` reports the case
+  it cannot fix (mg-e314, `Refs drellem2/pogo#110`).** Two changes shipped
+  together: `config.ServerURL()` returns `http://127.0.0.1:<port>` instead of
+  `http://localhost:<port>`, and a new `loopback resolution` row detects a
+  process shadowing the daemon port.
+
+  **The defect.** pogod binds `127.0.0.1` (`config.DefaultBind`). On a stock
+  macOS `/etc/hosts`, `localhost` resolves `::1` first — an address pogod never
+  holds and therefore any other process may claim. Whatever claims it becomes
+  pogod for every caller of `ServerURL()`. Go's dual-stack fallback does not
+  help: it retries the other family on connection *refused*, and an interloper
+  that accepts and answers wrongly never refuses. On 2026-07-31 a `kubectl
+  port-forward` landed on `::1:10000` after its IPv4 bind was refused and took
+  the pogo control plane down for ~20 minutes. **pogod was healthy the entire
+  time**, which is the whole difficulty: every instrument pointed at a working
+  daemon and reported a broken one.
+
+  `ServerURL()` has two non-test callers and neither wanted the name.
+  `internal/client` funnels every CLI command through it. `internal/selfdrift`'s
+  `RunningRev` reads `/version` through it — so under shadowing the drift
+  detector read the *interloper's* answer and reported `RevUnreachable` about a
+  healthy daemon, blinded alongside doctor's "pogod running" row. One line
+  repairs both. `DialAddr()` — what the spawn-race guard probes — has targeted
+  `127.0.0.1` explicitly since #22 and documents why; the two are now pinned
+  equal by a test, because their disagreement is exactly what let the guard see
+  a free port while the CLI talked to an impostor.
+
+  **The detector ships with the fix, not after it.** Pointing the CLI at
+  `127.0.0.1` stops the CLI being fooled; it does not empty the port. Editor
+  integrations (tracked separately as mg-b36f), a stale `ssh -L`, a forward on a
+  reconfigured port, a second daemon — everything else that dials the name is
+  still fooled, and silently. The new row is what makes that condition
+  observable at all, and it was written while the bug still reproduced rather
+  than after it went rare.
+
+  **The row refuses to be satisfied by liveness.** "The TCP dial succeeded" and
+  even "HTTP 200" are properties of any listener, not of pogod — the incident is
+  precisely a wrong process passing both — so a probe built on them would
+  reproduce, one layer up, the defect it exists to detect. The probe requires
+  `health.LivenessBody`, the one string only pogod emits, now read from
+  `internal/health` by both the daemon's handler and the probe so the two cannot
+  drift apart while the check keeps reporting pass. It carries a client timeout,
+  because a listener that accepts and never writes is a hang rather than an
+  error and would otherwise wedge `doctor --check` indefinitely.
+
+  **Four states, and the last two are what keep it from being noise.** Both
+  addresses answer → pass, naming which address the *name* landed on ("::1 was
+  free and Go fell back" and "pogod is where the name points" are different
+  worlds, and only the second is stable). Bind answers, name does not → **fail**,
+  saying pogod is HEALTHY and must not be restarted, naming the shadowing
+  address, and printing the `lsof` query for the family actually reached — a
+  `-i6TCP` remedy against an IPv4 shadow lists nothing, and a remedy that prints
+  nothing reads as a wrong check. Neither answers → **renders nothing at all**,
+  because check 1 already says the daemon is down and a checklist that reports
+  every stopped daemon twice is one people stop reading. Name answers but the
+  bind address does not → **warn**: a daemon configured with a v6-loopback
+  `bind`, which this change does stop reaching over the name. That configuration
+  is already half-broken — `DialAddr` has never found it, so the spawn-race
+  guard reads the port as free and may start a rival pogod — and the row is the
+  plain-language signal such a user currently gets nowhere.
+
+  **Shown to fail before it was believed.** A check for silent impersonation
+  that has only ever been observed returning pass is itself a silent
+  impersonation of a check. All four states were staged against real sockets and
+  the real binary: a sandboxed pogod on `127.0.0.1:10981` with a Python
+  responder bound to `[::1]:10981` reproduced the incident (`curl
+  localhost:10981` returned the impostor while pogod answered on 127.0.0.1), and
+  `pogo doctor --check` reported `✗ loopback resolution` and exited 1; removing
+  the impostor returned the same code to `✓`; stopping the daemon made the row
+  disappear; and a pogod started with `--bind "[::1]"` produced the warn. In the
+  test suite this is `TestLoopbackResolution_InterloperOnV6Fires`, which binds
+  both families of the same port and requires a fail, paired with
+  `..._PassesOnceTheV6PortIsFree`, which requires the same code to go quiet —
+  "fires when it should" and "does not fire when it should not" are two claims
+  and a detector needs both. `TestProbePogod_AcceptsOnlyPogodsBody` feeds it five
+  responders that would all pass a dial-succeeded probe, and four that would
+  pass an HTTP-200 one.
+
+  **Not done here.** Editor integrations carry their own `localhost:10000`
+  defaults and are not `ServerURL()` callers; they hit this class independently
+  and remain mg-b36f. The doctor row now detects the shadowing they suffer from,
+  but detection is on the CLI's checklist, not in the editors.
--- a/cmd/pogo/loopbackresolution.go
+++ b/cmd/pogo/loopbackresolution.go
@@ -20,9 +20,29 @@ package main
 // WHY IT IS NOT REDUNDANT WITH THE `ServerURL` FIX. The companion change in
 // this commit points the CLI at 127.0.0.1, so the CLI stops being fooled. It
 // does not empty the port: everything else that dials the NAME — editor
-// integrations (mg-b36f), a stale `ssh -L`, a forward on a reconfigured port,
-// a second daemon — still is. This check is what makes that condition visible
-// instead of silent, and it is why A ships with B rather than after it.
+// integrations (mg-b36f), a stale `ssh -L`, a forward whose far end is not a
+// daemon — still is. This check is what makes that condition visible instead
+// of silent, and it is why A ships with B rather than after it.
+//
+// EXACTLY WHAT IT DETECTS, AND WHAT IT DOES NOT. It detects a responder on the
+// NAME that is not POGOD-SHAPED. It cannot tell THIS pogod from ANOTHER one.
+// A second pogod bound to a v6 loopback emits health.LivenessBody like any
+// other, so both probes report pogod and the row passes — likewise an `ssh -L`
+// or port-forward whose far end is a real pogod elsewhere. That case is NOT
+// covered, and saying otherwise would be the failure this file argues against
+// one case over: a reader told the row covers a second daemon reads the green
+// line and stops looking. The approved recommendation (mg-e314) promised the
+// broader class in the same words; it over-promised, and this is the correction
+// rather than a quiet narrowing.
+//
+// The row does say what it saw. When both probes answer as pogod but the name
+// landed somewhere other than the bind address, the pass line names that
+// address and states plainly that a dual-stack bind and a second daemon are
+// both consistent with it and that this check does not separate them —
+// evidence the probe already holds, offered without a verdict it cannot
+// support. Separating them needs /version (start_time differs between two
+// daemons on one revision) and carries a dual-stack false positive; that is a
+// feature, not this row, and it is deliberately not built here.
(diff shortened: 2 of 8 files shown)


drellem2:
## Review round 1: FAIL
Reviewer: mg-a693 · build ticket: mg-e314 · blocking: 1 · advisory: 2

Gate is green and the positive control is real. I re-ran both directions myself against the built binary rather than taking the PR description's word for it, and staged one case the PR does not cover.

**Verified independently** (sandboxed `pogod` on `127.0.0.1:10981`, `HOME`/`XDG_CONFIG_HOME`/`POGO_HOME` redirected, never the live daemon):

| staged | observed |
|---|---|
| nothing on `::1:10981` | `✓ loopback resolution … the name reached 127.0.0.1:10981`, exit 0 |
| Python responder on `[::1]:10981` returning `200 Handling connection for 10981` | `✗ loopback resolution`, names `[::1]:10981`, hint `lsof -nP -i6TCP:10981 -sTCP:LISTEN`, **exit 1**. `curl localhost:10981` returned the impostor, `curl 127.0.0.1:10981` returned pogod, `pogod running` stayed `✓` — part B working |
| impostor removed, same binary | back to `✓`, exit 0 |
| only a `bind = "[::1]"` daemon alive | `! loopback resolution` warn, naming the rival-pogod risk |
| nothing listening on either family | row renders nothing; only `✗ pogod running` |

Both directions observed, both with the shipped code path. `./build.sh` from a non-ephemeral worktree: **exit 0, 0 failures**, host load 3–7 (`internal/refinery` green in 110s, so the load-sensitive gate the PR describes is not masking anything here).

### Blocking

- `cmd/pogo/loopbackresolution.go:24` — the row does not detect a **second pogod** holding `::1:<port>`, but the comment says it does: *"a stale `ssh -L`, a forward on a reconfigured port, a second daemon — still is. This check is what makes that condition visible instead of silent."* The approved recommendation makes the same promise (*"A stays valuable after B: it catches the rest of the class — a stale `ssh -L`, a forward on a reconfigured port, a second daemon"*).

  Observed: two real `pogod` processes, one on `127.0.0.1:10981` and one started with `bind = "[::1]"` on the same port. The v6 daemon emits `health.LivenessBody`, so both probes return `IsPogod` and the row renders **`✓ loopback resolution … the name reached [::1]:10981`, exit 0**. Same for any `ssh -L`/port-forward whose far end is a real pogod. The body match discriminates pogod-shaped from not-pogod-shaped; it cannot discriminate *this* pogod from *another* one, which is a third of the enumerated class.

  This blocks because the file's own framing is that a detector which retires suspicion is worse than none, and a reader who has been told the row covers a second daemon will read that green line and stop looking. Either resolution clears it:
  1. **Narrow the prose** (`loopbackresolution.go:20-25`, the changelog entry, the PR body) to what the row actually detects — a responder on the name that is not pogod-shaped — and state the uncovered case explicitly. No behaviour change required.
  2. **Act on the evidence the row already holds.** `loopbackresolution.go:183` prints `name.Remote`; when both probes answer as pogod and `name.Remote` is not the bind address, that is either a dual-stack `bind` (e.g. `0.0.0.0`, which also binds `[::]`) or a second daemon. `/version` separates them — `start_time` differed between my two daemons on an identical `revision`.

  Option 1 is sufficient. If you take it, say so in the PR body so the recommendation's over-promise is on the record rather than silently dropped.

### Advisory (non-blocking)

- `cmd/pogo/loopbackresolution.go:196,202` — in the branch where the name never connected, the lead correctly says *"could not be reached at all"* and `TestLoopbackResolutionLine_UnreachableNameFailsWithoutInventingACulprit` asserts the detail invents no culprit — but the shared suffix at :202 still reads *"is talking to that process instead of the daemon. Identify it with `lsof …`, … so the impersonator keeps the port until it is stopped"*. There is no process; the operator is sent hunting for one. Reachable on a host whose `localhost` resolves only `::1`, or where the name does not resolve. Found by reading, not staged.
- `cmd/pogo/loopbackresolution_test.go:229` — `loopbackPair` calls `t.Skipf` when `[::1]:0` will not bind, so on a host without IPv6 loopback the whole positive control skips and the fire path goes unproven there. The skip is loud and the pure-function tests still cover the render, so this is a note, not a defect.

### Not a finding, recorded so it is not rediscovered

`pogo doctor --check --json` exits **0** even with a failing row (it reports `"healthy": false` instead); the `os.Exit` lives only in the text branch. Pre-existing — identical structure on `main` at `cmd/pogo/main.go:3394` — and outside this PR's scope, but anyone scripting the new row should read `healthy`, not the exit code.

### Checked and clean

Design faithfulness: A and B both ship, A first in the diff; the check is extracted to a testable helper with the required positive control *and* the "neither answers" quiet case; `--check` help text updated (`cmd/pogo/main.go:2908`); `TestServerURL` updated and pinned equal to `DialAddr`; changelog entry present and in-convention; `Refs`, not a closing keyword. `grep -rn localhost --include='*.go'` now hits only the probe, the help text and comments — `config.go:1535` was indeed the only one. The deviation the PR flags (probing the literal name rather than `ServerURL()`) is correct: after B the recommendation's literal comparison would compare an address to itself. Scope boundary held — no editor-integration files touched (mg-b36f).

### Not checked

Linux, or any host whose `/etc/hosts` differs from this one; IPv6 disabled at the OS level; the launchd-installed daemon (sandbox only, live `:10000` never touched); `internal/selfdrift.RunningRev` under shadowing (reasoned from the one-line `ServerURL` change, not exercised).

drellem2:
## Round 1 addressed — head `374714a`

Thank you for reproducing the control independently rather than trusting the body, and for staging the two-pogod case. The blocking finding is correct and I have taken **option 1**.

### BLOCKING — over-claimed coverage of a second daemon (`loopbackresolution.go:24`) — fixed in `374714a`

You are right, and I verified it your way before changing anything: two real pogods on `127.0.0.1:10982` and `[::1]:10982`, identical revision, `start_time` 6s apart — both emit `health.LivenessBody`, both probes return `IsPogod=true`, row green.

Option 1 rather than option 2, for your reason plus one: option 2's dual-stack false positive (`0.0.0.0` also binds `[::]`) would make the row fire on a legitimate configuration, and the recommendation never asked for the feature. Building it would be scope creep dressed as rigour.

Narrowed in all three places the claim travels:

- **`loopbackresolution.go:20-41`** — the "still is" list drops "a second daemon" (now "a forward whose far end is not a daemon"), and a new **`EXACTLY WHAT IT DETECTS, AND WHAT IT DOES NOT`** section names the uncovered case outright, records that the recommendation over-promised, and says why option 2 was declined.
- **`changelog.d/mg-e314.fixed.md`** — same correction, stated as a correction.
- **PR body** — a scope-correction callout at the top of section A, per your ask that it be on the record.

### Extended to where the reader actually is

Prose in a source comment does not reach the operator looking at a green line. Using evidence the probe already holds (`name.Remote`, your `:183` observation), the pass line now reads, against your exact staging:

```
✓  loopback resolution   pogod answers on both http://127.0.0.1:10982 and http://localhost:10982;
                         the name reached [::1]:10982, which is NOT the address pogod was probed on
                         (127.0.0.1:10982). A dual-stack bind and a SECOND pogod both look like this,
                         and this check does not tell them apart — it distinguishes pogod-shaped
                         responders from other processes, not one pogod from another. If you did not
                         configure a dual-stack bind, compare `/version` start_time on both addresses
```

No status change — it is still a pass, because the check genuinely cannot rule. It states the observation and its own limit, and points at the thing that does settle it. `TestLoopbackResolutionLine_SecondPogodPassesButSaysSo` pins it; `..._PassStaysQuietWhenTheNameLandsOnTheBindAddress` pins that it stays silent on an ordinary healthy pass, so the caveat is not bought with words on every run.

### ADVISORY 1 — no-culprit branch (`:196,202`) — fixed in `374714a`

Real, and cheap. The shared suffix is gone; the "name never connected" branch has its own remedy:

> …could not be reached at all (…), while pogod is HEALTHY on … — do not restart it. **Nothing answered on the name, so there is no process to hunt: this is name resolution, not an impersonator.** Check what `localhost` resolves to on this host (`/etc/hosts`, then `dns-sd -q localhost`) — anything that dials the name rather than the address will fail the same way until it resolves to 127.0.0.1:10982

The existing test now also forbids `lsof`, "talking to that process" and "keeps the port until it is stopped" in that branch — asserting on the *actions* a reader would take, since "impersonator" is legitimate in the negation.

### ADVISORY 2 — `loopbackPair` skips without IPv6 — left as-is

Agreed it is a note. The skip is loud and names the reason, and the pure-function tests still cover all four renders. Faking IPv6 to keep the control nominally running would make it green on a host where it proves nothing, which is the defect this PR is about.

### Not touched

`doctor --check --json` exiting 0 on a failing row — pre-existing, identical on `main`, out of scope. Recorded, not fixed.

### Re-verification on `374714a`

`./build.sh` **exit 0, zero failures**, host load 2.4. All four states re-staged against real sockets with the rebuilt binary: two-pogod pass (above), ordinary pass (caveat correctly absent), impostor → `✗` + exit 1, daemon down → silent, `bind="[::1]"` → warn. Live `:10000` never touched; sandbox torn down and ports confirmed clear.

drellem2:
## Review round 2: PASS
Reviewer: mg-a693 · build ticket: mg-e314 · head `374714a` · blocking: 0 · advisory: 2

Round 1's blocking finding is resolved, and the fix went further than the narrowing I asked for — the pass line now carries the limit to where an operator reads it, not just to a source comment. I re-staged everything on the new head rather than accepting the report.

**Re-verified on `374714a`** (rebuilt binaries; sandboxed pogod on `127.0.0.1:10981`, live `:10000` never touched):

| staged | observed |
|---|---|
| `::1` free | `✓ … the name reached 127.0.0.1:10981` — caveat correctly **absent**, exit 0 |
| impostor on `[::1]:10981` | `✗`, names `[::1]:10981`, `-i6TCP` hint, **exit 1** |
| impostor removed | back to `✓`, exit 0 |
| **two real pogods**, `127.0.0.1` + `[::1]`, same port | `✓ … the name reached [::1]:10981, which is NOT the address pogod was probed on (127.0.0.1:10981). A dual-stack bind and a SECOND pogod both look like this, and this check does not tell them apart … compare `/version` start_time on both addresses` |
| only a `bind="[::1]"` daemon | `!` warn |
| nothing listening | row absent |

The positive control still fires and still goes quiet, on the changed code. The prescribed remedy also works: `/version` on the two daemons returned identical `revision` `ec68dc1a` with `start_time` 30s apart, so an operator following that line can in fact settle it.

`./build.sh` from a non-ephemeral worktree: **exit 0**, every suite `0 failed`, host load 2.8–3.2. `cmd/pogo` ran fresh (31.6s); packages the diff does not touch were cache hits from the round-1 run, which was itself green at load 3–7.

**Blocking (round 1) — cleared.** `loopbackresolution.go:27-45` now states the uncovered case and records that the recommendation over-promised; the changelog and PR body carry the same correction. The pass line names the address, names both explanations, and states its own limit without inventing a verdict — and `TestLoopbackResolutionLine_PassStaysQuietWhenTheNameLandsOnTheBindAddress` stops that text being paid for on every healthy run. Keeping it a pass is right: the check genuinely cannot rule.

**Advisory 1 (round 1) — cleared.** The no-culprit branch has its own remedy (`/etc/hosts`, `dns-sd`) and the test forbids the culprit-hunting actions rather than the word. Verified by test and by reading; not staged, since it needs a host whose `localhost` does not resolve to a listening pogod.

**Advisory 2 (round 1) — withdrawn.** I agree with leaving `loopbackPair`'s skip as-is. Faking IPv6 to keep the control green on a host where it proves nothing is the defect this PR is about.

### Advisory (non-blocking, new)

- `cmd/pogo/loopbackresolution.go:51` — "health.LivenessBody — the one string only pogod emits" is a bound claim that the new section 24 lines above correctly refutes: any process can emit that string, which is why a second pogod passes. Suggest "the string pogod's `/health` emits".
- `cmd/pogo/loopbackresolution.go:247` — `loopbackLsofHint`'s `family := ""` fallback is now unreachable. Its only caller is the branch where `name.Remote != ""` (`:240`), and a TCP remote always parses, so the family is always `4` or `6`; round 1's test pinning the family-agnostic form went away with the branch that produced it. Either drop it or note it as unreachable, per the convention in `1fe0177`.

### Not checked

Linux or any other `/etc/hosts` ordering; IPv6 disabled at the OS level; the launchd-installed daemon; `selfdrift.RunningRev` under shadowing (reasoned from the one-line `ServerURL` change, never exercised). Unchanged from round 1 and still true: `pogo doctor --check --json` exits 0 on a failing row — pre-existing, identical on `main`, correctly left alone.

drellem2:
## Round 2 advisories applied — head `374714a` → `03debf1`

Both round-2 nits fixed. **Comments only — no code, no test changes** (verified: every changed line is a `//` line).

- **`loopbackresolution.go:51`** — "the one string only pogod emits" was a bound claim the section added in `374714a` refutes 24 lines above. Every pogod emits it; that is precisely why the row cannot single one out. Now reads "the string pogod's `/health` emits", and states the limit that follows. Leaving a self-contradicting claim in the file that spent a round correcting exactly that seemed wrong.
- **`loopbackLsofHint` family fallback** — unreachable since the no-culprit branch got its own remedy (sole caller runs only where `name.Remote != ""`, and a remote recorded by `net` always parses). **Kept**, not dropped: a future caller passing a less well-formed address should get a familyless hint rather than one naming the *wrong* family, since a remedy that lists nothing reads as a wrong check. Documented as unreachable per the convention in `1fe0177`, including that round 1's test pinning that form went away with the branch it covered.

`./build.sh` **exit 0, zero failures**.

**@va693 / mayor — note the head moved after the pass.** The review verdict is recorded against `374714a`; `03debf1` is that commit plus comment text. I judged two comment fixes worth taking since the reviewer offered them and the file's whole subject is not over-claiming — but the decision of which head to submit is the mayor's, and **if you would rather submit the exact reviewed head, say so and I will revert `03debf1` immediately**. I am not submitting anything myself.

drellem2:
## Review addendum: PASS extends to `03debf1`
Reviewer: mg-a693 · build ticket: mg-e314 · blocking: 0 · advisory: 0 open

The head moved after the round-2 pass (`374714a` → `03debf1`) to apply the two advisory nits. Re-checked rather than waved through, because a pass that silently follows a moving head is not a pass.

**The delta is comment-only, mechanically verified.** One file, +13/-1; every changed line matches `^[+-]\s*//`. Gate on `03debf1`: `./build.sh` **exit 0**, every suite `0 failed`, `cmd/pogo` fresh at 43.2s, host load 10–14. Positive control re-run on rebuilt binaries at this head: `✓` exit 0 → impostor on `[::1]:10981` → `✗` exit 1 → impostor removed → `✓` exit 0.

Both nits resolved, and the second one better than I proposed:

- `loopbackresolution.go:51` — the bound claim is gone and the limit that follows is stated inline rather than the wording merely softened.
- `loopbackresolution.go:248-257` — the `-iTCP:<port>` fallback is **kept and documented unreachable** rather than dropped, per the convention in `1fe0177`. Keeping it is the better call: a future caller passing a less well-formed address should get a familyless hint, not one naming the wrong family, and a remedy that lists nothing reads as a wrong check. The note also records that round 1's test pinning that form went away with the branch it covered, so the empty family cannot be misread as reachable behaviour.

No open findings. The verdict of record (`mg-a693`) was written against `374714a`; **it applies to `03debf1`**, which is the head to merge.

drellem2:
Merged as a93c71e087b8d914035245e4eb80b00c5c0f6a4c on `main` by the pogo refinery (MR mr-d9qike2tjv1h244d85n0).

Issue #108, opened 2026-07-31
`start-orchestration` returns to full mode without restarting agents, reports success anyway, and leaves `StopAll`'s shutdown latch set forever


drellem2:
---

## Summary

A `full → index-only → full` round-trip leaves a running daemon with **no crew agents and no
working crash-respawn**, while the CLI reports success. `transitionToFull` restarts the
refinery but never restarts the agent registry, and the `shutdown` latch that `StopAll` sets
is never cleared for the remaining life of the process.

Observed on pogod 0.5.0; verified still present on `main` (0.8.0).

The user-visible symptom is "my coordinator did not come back and I had to start it by
hand", with no error anywhere to explain it.

## Reproduction

```
pogo server stop        # no --all: drops to index-only, daemon keeps running
pogo server start       # prints "Orchestration restarted"
pogo agent list         # empty — the crew is gone
```

Then let any `restart_on_crash = true` crew agent exit. Instead of respawning:

```
agent <name>: restart failed: registry shut down
```

## Observed log from a real occurrence

```
15:05:35 server: transitioning to index-only mode
15:05:37 agent mayor: exited (err=<nil>)
15:05:37 agent mayor (crew) exited unexpectedly, scheduling restart
15:05:37 agent mayor: stopped (restart_on_crash=true; supervisor will respawn)
15:05:39 agent mayor: restart failed: registry shut down
15:05:41 server: now in index-only mode
15:08:56 server: transitioning to full mode
15:08:56 refinery: started (poll_interval=30s)
15:08:56 server: now in full mode          <-- no autostart line
15:09:06 agent mayor: spawned pid=2896     <-- a human, by hand
```

**Diagnostic signature.** On every genuine daemon boot, `autostart: started <name>`
accompanies `refinery: started` in the *same second* (measured across 8 consecutive boots).
After a mode round-trip the refinery comes back alone. `now in full mode` with no
`autostart:` line is this bug.

## Cause

Three symptoms, one defect: the transition does half its job and says it did all of it.

**1. `transitionToFull` does not restart agents** — `internal/server/server.go:119-145`. It
resolves `startRefinery`, swaps the refinery, and flips `s.mode`. It never touches
`s.agents`. There is no agent-side analogue of `SetRefineryStarter`, and
`Registry.AutoStartAgents()` has exactly one caller: `cmd/pogod/main.go:2381`, the boot path.

Two docstrings already assert the behaviour that is missing:

[...]

PR #119, opened 2026-08-07
fix: start-orchestration restarts agents and clears StopAll's latch after the drain, and the CLI names what came back

drellem2:
A full → index-only → full round-trip left a running daemon with **no crew
agents and no working crash-respawn**, while the CLI printed "Orchestration
restarted" and exited zero. Three symptoms, one defect: the transition did half
its job and reported all of it.

The fix is at the choke point — `transitionToFull` — so the CLI and
`POST /server/start-orchestration` are covered by one change.

## 1. The transition now restarts the crew

`transitionToFull` resolved the refinery starter, swapped the refinery, flipped
the mode, and never touched `s.agents`. There was no agent-side analogue of
`SetRefineryStarter`, and `AutoStartAgents` had exactly one non-test caller —
the boot path in `cmd/pogod/main.go`. Two docstrings already asserted the
behaviour that was not there.

`Server.SetAgentStarter` is that analogue, wired in `cmd/pogod` beside the
refinery one. It **re-runs the auto-start sweep**: the desired state is what the
prompt frontmatter declares — the same rule boot applies — not "whatever
happened to be running before the stop". `AutoStartAgents` is already idempotent
via its `r.Get` guard, so an agent that survived the drain is skipped rather
than double-started.

`[agents] autostart = false` suppresses the sweep here exactly as it suppresses
boot. A mode round-trip must not become a side door for spawning a fleet the
boot path refuses.

## 2. StopAll's shutdown latch is cleared — and where

The latch was set in `StopAll` and cleared **nowhere in the tree**. It is read
in `Respawn` but not in `Spawn`, so after one stop-orchestration `pogo agent
start` kept working while `restart_on_crash` was silently inert for every agent,
for the life of the process. That asymmetry is why it was easy to miss: the
fleet looked hand-recoverable and healthy right up until something crashed.

**Where the clear happens.** `Registry.Resume()` is called from
`transitionToFull`, **after the refinery has successfully restarted and the mode
has flipped, and before the auto-start sweep**.

**What I considered on the later side.** `Resume` invalidates every respawn
scheduled before it (see below). Clearing *after* the sweep would therefore
leave exactly the agents that crashed during the sweep permanently unsupervised
— their respawn goroutine, scheduled a moment earlier, would be killed by the
clear itself — in a fleet that reports itself fully restored. That is the same
class of bug as the one being fixed, narrowed to the most fragile agents.

**What I considered on the earlier side.** Clearing at the top of
`transitionToFull` — or at the end of `transitionToIndexOnly` — re-admits the
drain's own late respawn goroutines into a fleet that is being rebuilt. The
interval is measured, not hypothetical: the `OnExit` respawn goroutine sleeps
**2 seconds** before calling `Respawn`, while `StopAll` returns synchronously,
so a goroutine scheduled during teardown fires ~2s *after* the drain returns.
It also matters that a refinery failure returns early: re-arming crash-respawn
for a transition that did not happen would leave the daemon supervising a fleet
it never restarted. `TestTransitionToFullAbortsWholeWhenRefineryFails` pins
that — mode stays index-only, no sweep runs, latch stays set.

**The barrier.** Clearing the latch cannot close that 2s window, because the
window is on the far side of the clear. So the latch is not a bool any more:
`Registry.generation` bumps on **both edges** — set by `StopAll`, cleared by
`Resume` — and a deferred caller captures `Generation()` at *scheduling* time
and calls `RespawnFromGeneration(name, gen)`. Any stop or start in between
refuses it, however late it fires. `cmd/pogod`'s `OnExit` hook now captures the
generation outside the goroutine, before the sleep. That closes the window by
construction rather than by timing.

### Honest note on how load-bearing that barrier is

I measured it rather than asserting it. With the generation check disabled,
`TestRespawnFromGenerationLosesToARoundTrip` fails as expected — but the
end-to-end goroutine test **still passes**, because `StopAll` also drops each
stopped agent from the map, so a stale respawn of an agent present at drain time
hits `agent "stale" not found` anyway. In every sequence I could construct, that
removal plus `Respawn`'s still-running check already defeat the stale respawn.

I kept the generation because that defence is *emergent* — spread across
`StopAll`'s cleanup loop and `Respawn`'s status check, in two functions, neither
of which documents that a third function depends on it — and because it is what
makes `Resume` safe to call at a chosen point in the sequence rather than only
at a point where the other two guards happen to hold. It is one local check at
the point of respawn. The test comments say exactly this, so nobody later reads
the end-to-end test as proof the barrier is doing the work.

If a reviewer would rather not carry the mechanism, the place to cut it is
`respawn(name, gen, checkGen)` plus the `Generation()`/`RespawnFromGeneration`
pair; `Resume()` and the sequencing stand on their own.

## 3. The CLI names what came back

The reporter's symptom is a command that returns "full mode" without saying
which agents actually restarted, and a green return is precisely what the bug
already produces. So the output is now:

```
Orchestration restarted
  Refinery: restarted.
  Crew agents started (2): mayor, pm-pogo
  Crew agents left parked (1): sleepy — down on purpose; wake with 'pogo agent wake <name>'.
  Crew agent FAILED to start: gc — pty start: no ptys
  Polecats are NOT restored — they are ephemeral and are re-dispatched per work item.
```

Every branch names something. When nothing started it says **why** — `[agents]
autostart = false`, or no prompt declaring `auto_start = true` — because a
silent zero is the shape of the original bug. Polecats get their own line even
on a fully successful restart: their absence would otherwise read as more of the
same defect.

`POST /server/start-orchestration` returns the whole report; the `mode` field is
unchanged and still means what it did, so existing readers keep working. `--json`
carries it under `report`.

**One judgement call worth flagging:** a crew agent that *errored* on spawn now
makes `pogo server start` exit non-zero, after printing the report. Restoring
zero agents because none are eligible (or because autostart is off) stays exit 0
— that is a configuration fact, not a failure. I took "a partial restore is not
a success" as the intent behind "reports success anyway"; if the preference is
to keep the command exit-0 for scriptability and rely on the printed names plus
`agents_failed` in JSON, that is a two-line revert.

## Verification

**Positive control, as required — the tests fail against the unfixed
behaviour.** I neutered the fix in place (kept the API, removed the `Resume()`
call and the sweep) and ran the regression tests:

```
--- FAIL: TestStartOrchestrationRestartsCrewAgents (0.01s)
    orchestration_restart_test.go:102: agent "scout" was NOT restarted by the return to full mode — the daemon is in full mode with no crew (gh #108)
    orchestration_restart_test.go:110: report.AgentsStarted = [], want it to name "scout"
--- FAIL: TestStartOrchestrationClearsShutdownLatch (0.01s)
    orchestration_restart_test.go:161: Respawn after a stop/start round-trip failed: registry shut down — StopAll's shutdown latch was never cleared, so restart_on_crash is silently inert for the life of this process (gh #108)
```

The tests run against a **real `agent.Registry`** with real processes, under
`testsandbox.Isolate`. That is the point, not a detail: all sixteen `New(...)`
call sites in the existing server suite pass a nil registry and
`transitionToIndexOnly` guards with `if agents != nil`, so the suite was
structurally incapable of seeing this. A test was not missing — the harness
could not have failed. A new test inheriting the nil-registry pattern would have
reproduced that blindness exactly.

**The observable the reporter asked for.** The log from the round-trip in
`TestStartOrchestrationRestartsCrewAgents` now reads:

```
server: transitioning to full mode
server: now in full mode
agent scout: spawned pid=69252 type=crew proc=pogo-crew-scout
autostart: started scout (pid=69252)
server: full mode restored — crew started=[scout] already_running=[] parked=[] failed=0
```

The issue's diagnostic signature is `server: now in full mode` with **no**
`autostart:` line following it. The `autostart:` line is now there, and the
agent is named in the report the CLI prints — observed from the output, not
inferred from an exit code.

**Wire-level, because the failure mode is silent.** The handler encodes the
report and the client decodes it, and nothing else in the tree checked that
those two agree. If they drift — a renamed tag, a type change — the client gets
a zero-valued report and the CLI prints "restarted, 0 agents", which is
*indistinguishable from the bug*. `TestStartOrchestrationCarriesTheReportOverTheWire`
runs the real handler against `httptest` and asserts the names survive.

### CLI-level, against a real daemon — the reporter's own observation

The issue is a CLI-level complaint, so the last check is a real `pogod` in an
isolated sandbox, driven by the real `pogo` binary, with **the same script run
against `main` and against this branch**. Two auto-start crew prompts plus the
mayor; `[agents] command = "sleep 600"`.

**`main` — the reported symptom, reproduced verbatim:**

```
=============== pogo server start ===============
Restarting orchestration...
Orchestration restarted
exit=0

=============== agents after ===============
No running agents.

=============== daemon log ===============
server: now in full mode
server: now in full mode
```

Green return, `No running agents`, and `server: now in full mode` with **no
`autostart:` line following it** — the diagnostic signature named in the issue.

**This branch, same script:**

```
=============== pogo server start ===============
Restarting orchestration...
Orchestration restarted
  Refinery: restarted.
  Crew agents started (3): mayor, scout, tinker
  Polecats are NOT restored — they are ephemeral and are re-dispatched per work item.
exit=0

=============== agents after ===============
mayor                 pid=98372   type=crew      status=running     uptime=2s
scout                 pid=98374   type=crew      status=running     uptime=2s
tinker                pid=98376   type=crew      status=running     uptime=2s

=============== daemon log ===============
server: now in full mode
autostart: started mayor (pid=98372)
autostart: started scout (pid=98374)
autostart: started tinker (pid=98376)
server: full mode restored — crew started=[mayor scout tinker] already_running=[] parked=[] failed=0
```

The agents are named **by the CLI's own output**, and observed back in `agent
list` with fresh pids — not inferred from the exit code, which is `0` in both
runs and is exactly why the exit code proves nothing here. `--json` carries
`"agents_started": ["mayor","scout","tinker"]`.

**How the sandbox was made safe, because my first attempt was not.** Attempt one
redirected `HOME`, `XDG_CONFIG_HOME` and `POGO_HOME` and still resolved to the
**live** daemon — `pogo agent list` returned this machine's real fleet. Nothing
was stopped only because I had guessed a subcommand that does not exist. The
working version pins the client with `POGO_PORT` and gates on an exact check
before any mutating command: the process listening on the port the CLI resolves
to must be the pid this script started (`listener on :18917 = 94372   my pogod =
94372`), with no polecats and no multi-hour uptimes in the probe. A
fleet-shape guess is what nearly went wrong — "a mayor exists" is true of the
sandbox too. Stray listeners are reaped and the port confirmed free on exit.

`./build.sh` green (exit 0, zero failures). `internal/agent` also green under
`-race`.

## Noted, not fixed (out of scope)

`pogo agent list` against a daemon in index-only mode prints:

```
json: cannot unmarshal object into Go value of type []agent.AgentInfo
```

`RequireOrchestration` returns a 503 JSON *object* (`{"error":...,"mode":...}`)
and the client decodes into a slice. Reproduces identically on `main` and on
this branch, so it is pre-existing and untouched by this change — but it is
adjacent enough that a reviewer will see it in the transcripts above, and it is
a bad message in exactly the state this PR is about. Flagging rather than
fixing.

## Relationship to the "drain counts polecats" finding

A separate open ticket observes that the drain's predicate counts polecats while
compute can outlive the polecat, so "0 polecats active" does not mean "no work
in flight". I considered whether that undermines this change and concluded it
does not, in either direction: `StopAll` waits on registry entries, and detached
grandchild processes were never covered by it and are not covered by the
generation barrier either — that barrier is pogod-internal and makes no claim
about host quiescence. The crew agents restarted here do not require a quiet
host for correctness; a loaded host costs them contention, not consistency.

One directional note, stated because it is a real consequence: by making
crash-respawn work again after a round-trip, this change means an agent that
crashes post-restart will now actually come back — so a post-drain host can get
busier than it did before the fix. That is the intended behaviour being
restored, not a regression, but it is a change in what happens after a drain. I
have deliberately not answered that ticket's open question here.

## Refs

Refs drellem2/pogo#108 — **deliberately not a closing reference.** The triage
packet records that #108 is open on purpose: the drafted public reply is
human-gated and has not been posted. Auto-closing it on merge would shut the
issue on the reporter before anyone had answered them. Close it by hand once the
reply goes out.

Approved triage recommendation: mg-c23c (verdict: implement at
`transitionToFull`; re-run the auto_start sweep rather than restore the prior
set; `autostart = false` must suppress the transition path; regression test
against a real `Registry`). Signed off by pm-pogo 2026-08-05.

Work item: mg-4b01

### DIFF
--- a/changelog.d/mg-4b01.fixed.md
+++ b/changelog.d/mg-4b01.fixed.md
@@ -1,0 +1,47 @@
+- **`pogo server start` against a stopped daemon now actually restarts the crew,
+  re-arms crash-respawn, and names what came back (mg-4b01, drellem2/pogo#108).**
+  A full → index-only → full round-trip used to leave a running daemon with no
+  crew agents *and* no working crash-respawn, while the CLI printed
+  "Orchestration restarted" and exited zero. Three symptoms, one defect: the
+  transition did half its job and reported all of it.
+
+  - **`transitionToFull` never touched the agent registry.** It resolved the
+    refinery starter, swapped the refinery, flipped the mode, and returned. There
+    was no agent-side analogue of `SetRefineryStarter`, and `AutoStartAgents` had
+    exactly one non-test caller — the boot path. Two docstrings already claimed
+    the behaviour that was not there. There is now a `SetAgentStarter` hook, wired
+    in `cmd/pogod` beside the refinery one, and the transition **re-runs the
+    auto-start sweep**: the desired state is what the prompt frontmatter declares,
+    the same rule boot applies, not "whatever happened to be running before the
+    stop". `[agents] autostart = false` suppresses it exactly as it suppresses
+    boot — a mode round-trip is not a side door around the config.
+
+  - **`StopAll`'s shutdown latch was one-way.** It was set on stop and cleared
+    nowhere in the tree, so after a single stop-orchestration `restart_on_crash`
+    was inert for the life of the process. It read healthy on every instrument:
+    `Spawn` never consulted the latch, so `pogo agent start` kept working and the
+    fleet looked hand-recoverable — right up until something crashed. `Resume()`
+    now clears it, and `Registry.Generation()` /`RespawnFromGeneration()` make the
+    clear safe: the latch bumps a generation on **both** edges, and a deferred
+    respawn carries the generation it was scheduled in. pogod's `OnExit` hook
+    sleeps 2s before respawning while `StopAll` returns synchronously, so a
+    teardown-scheduled respawn fires *after* the drain returns — and a stop→start
+    inside that window would have cleared the latch in time to admit it. The
+    generation closes that by construction rather than by timing.
+
+  - **The CLI reported success regardless.** It now prints the daemon's report:
+    which crew agents started, which were already running, which are parked
+    (down on purpose, not casualties), which **failed** and why — and states
+    that polecats are not restored, so their absence is not read as more of the
+    same bug. When nothing started it says why: autostart disabled, or no prompt
+    declaring `auto_start = true`. A crew agent that *errored* on spawn exits
+    non-zero; restoring zero agents because none are eligible stays exit 0. The
+    `--json` form carries the whole report under `report`, and
+    `POST /server/start-orchestration` returns it (the `mode` field is unchanged,
+    so existing readers still work).
+
+  The regression tests run against a **real `agent.Registry`** with real
+  processes. That is the point, not a detail: all sixteen `New(...)` call sites
+  in the existing server suite pass a nil registry, and `transitionToIndexOnly`
+  guards with `if agents != nil`, so the suite was structurally incapable of
+  seeing this. A test was not missing — the harness could not have failed.
--- a/cmd/pogo/main.go
+++ b/cmd/pogo/main.go
@@ -175,8 +175,20 @@ Child commands include start, stop, and status.`,
 	var cmdServerStart = &cobra.Command{
 		Use:   "start",
 		Short: "Start the pogo server",
-		Long:  `Start the pogo server.`,
-		Args:  cobra.MinimumNArgs(0),
+		Long: `Start the pogo server.
+
+If the server is already running in index-only mode (after 'pogo server
+stop-orchestration'), this restarts orchestration instead: the refinery comes
+back, crash-respawn is re-armed, and the crew auto-start sweep is re-run.
+
+It then reports what actually came back, by name — which crew agents started,
+which were already running, which are parked, and which failed. Polecats are
+ephemeral and are never restored; they are re-dispatched per work item.
+
+Crew agents are not restarted when the daemon runs with [agents] autostart =
+false; the output says so rather than reporting an empty success. Exits
+non-zero if any crew agent errored while starting.`,
+		Args: cobra.MinimumNArgs(0),
 		Run: func(cmd *cobra.Command, args []string) {
 			err := client.HealthCheck()
 			if err != nil {
@@ -206,16 +218,30 @@ Child commands include start, stop, and status.`,
 				if !jsonOutput {
 					fmt.Println("Restarting orchestration...")
 				}
-				if err := client.StartOrchestration(); err != nil {
+				report, err := client.StartOrchestration()
+				if err != nil {
 					cli.ExitWithError(jsonOutput, err.Error(), cli.ExitError)
 				}
 				if jsonOutput {
 					cli.PrintJSON(map[string]interface{}{
 						"status":  "started",
-						"message": "orchestration restarted",
+						"message": orchestrationRestartSummary(report),
+						"report":  report,
 					})
 				} else {
 					fmt.Println("Orchestration restarted")
+					for _, line := range orchestrationRestartLines(report) {
+						fmt.Println("  " + line)
+					}
+				}
+				// A crew agent that ERRORED on spawn is a failure of the
+				// restart, not a detail of it, so the exit code says so —
+				// after the report has been printed, so the operator gets the
+				// names either way. Restoring zero agents because none are
+				// eligible (or because [agents] autostart = false) is not an
+				// error and stays exit 0; that case is reported in words.
+				if len(report.AgentsFailed) > 0 {
+					os.Exit(cli.ExitError)
 				}
 				return
 			}
--- a/cmd/pogo/startorchestration.go
+++ b/cmd/pogo/startorchestration.go
@@ -1,0 +1,87 @@
+package main
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/drellem2/pogo/internal/server"
+)
+
+// This file renders pogod's StartReport for `pogo server start` when that
+// command finds the daemon in index-only mode and restarts orchestration.
+//
+// It exists because the old output was one unconditional line — "Orchestration
+// restarted" — printed with a zero exit whether or not a single agent came
+// back. gh #108 is exactly that: the transition restored nothing, and the CLI
+// could not tell the operator, because it was never told either. A green
+// return is what the defect already produces, so naming the fleet is not
+// cosmetic; it is the part of the fix an operator can actually see.
+
+// orchestrationRestartLines renders the human-readable detail lines for a
+// restart, one per fact. Every line names something: which agents came back,
+// which were already up, which are parked, which failed, and what is
+// deliberately not restored.
+func orchestrationRestartLines(r server.StartReport) []string {
+	if r.AlreadyFull {
+		return []string{"Already in full mode — nothing was stopped, so nothing was restarted."}
+	}
+
+	var lines []string
+	if r.RefineryRestarted {
+		lines = append(lines, "Refinery: restarted.")
+	} else {
+		lines = append(lines, "Refinery: NOT restarted (the daemon has no refinery starter configured).")
+	}
+
+	if len(r.AgentsStarted) > 0 {
+		lines = append(lines, fmt.Sprintf("Crew agents started (%d): %s",
+			len(r.AgentsStarted), strings.Join(r.AgentsStarted, ", ")))
+	}
+	if len(r.AgentsAlreadyRunning) > 0 {
+		lines = append(lines, fmt.Sprintf("Crew agents already running (%d): %s",
+			len(r.AgentsAlreadyRunning), strings.Join(r.AgentsAlreadyRunning, ", ")))
+	}
+	if len(r.AgentsParked) > 0 {
+		lines = append(lines, fmt.Sprintf("Crew agents left parked (%d): %s — down on purpose; wake with 'pogo agent wake <name>'.",
+			len(r.AgentsParked), strings.Join(r.AgentsParked, ", ")))
+	}
+	for _, f := range r.AgentsFailed {
+		lines = append(lines, fmt.Sprintf("Crew agent FAILED to start: %s — %s", f.Name, f.Error))
+	}
+
+	// The "nothing came back" cases, each said out loud with its reason. This
+	// is the branch the reporter hit; silence here is the whole bug.
+	if len(r.AgentsStarted) == 0 && len(r.AgentsAlreadyRunning) == 0 && len(r.AgentsFailed) == 0 {
+		switch {
+		case r.AgentStartSkipped != "":
+			lines = append(lines, "No crew agents started: "+r.AgentStartSkipped)
+		default:
+			lines = append(lines, "No crew agents started — no crew prompt declares auto_start = true.")
+		}
+	} else if r.AgentStartSkipped != "" {
+		lines = append(lines, "Crew auto-start sweep did not run: "+r.AgentStartSkipped)
+	}
+
+	// Always stated, including when the sweep restored a full crew. Polecats
+	// are ephemeral and correctly stay gone; an operator who is not told that
+	// reads their absence as more of the same defect.
+	lines = append(lines, "Polecats are NOT restored — they are ephemeral and are re-dispatched per work item.")
+	return lines
+}
+
+// orchestrationRestartSummary is the one-line form used as the `message` field
+// in --json output, where the structured report sits beside it.
+func orchestrationRestartSummary(r server.StartReport) string {
+	if r.AlreadyFull {
+		return "already in full mode; nothing restarted"
+	}
+	parts := []string{fmt.Sprintf("orchestration restarted; %d crew agents started", len(r.AgentsStarted))}
+	if len(r.AgentsFailed) > 0 {
+		parts = append(parts, fmt.Sprintf("%d failed to start", len(r.AgentsFailed)))
+	}
+	if r.AgentStartSkipped != "" {
+		parts = append(parts, "auto-start sweep skipped: "+r.AgentStartSkipped)
+	}
+	parts = append(parts, "polecats not restored")
+	return strings.Join(parts, "; ")
+}
--- a/cmd/pogo/startorchestration_test.go
+++ b/cmd/pogo/startorchestration_test.go
@@ -1,0 +1,113 @@
+package main
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/drellem2/pogo/internal/server"
+)
+
+// The reporter's symptom was a command that returns "full mode" without saying
+// which agents actually restarted. These tests assert the output NAMES things,
+// because a message that merely stops lying — "restarted, possibly nothing" —
+// would leave the same operator trap in place.
+
+func joined(lines []string) string { return strings.Join(lines, "\n") }
+
+func TestRestartLinesNameTheAgentsThatCameBack(t *testing.T) {
+	out := joined(orchestrationRestartLines(server.StartReport{
+		Mode:              "full",
+		RefineryRestarted: true,
+		AgentsStarted:     []string{"mayor", "pm-pogo"},
+	}))
+
+	for _, want := range []string{"mayor", "pm-pogo", "Refinery: restarted"} {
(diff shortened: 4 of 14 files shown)


drellem2:
**CLI-level positive control added to the PR body** — the same sandbox script run against `main` and against this branch, with a real `pogod`.

`main`:
```
Orchestration restarted
exit=0
--- agents after ---
No running agents.
--- log ---
server: now in full mode          <-- no autostart: line follows it
```

this branch:
```
Orchestration restarted
  Refinery: restarted.
  Crew agents started (3): mayor, scout, tinker
  Polecats are NOT restored — they are ephemeral and are re-dispatched per work item.
exit=0
--- agents after ---
mayor / scout / tinker running, fresh pids
--- log ---
server: now in full mode
autostart: started mayor (pid=98372)
server: full mode restored — crew started=[mayor scout tinker] ...
```

Note the exit code is `0` in **both** runs. That is the point: a green return is what the bug already produced, so the exit code proves nothing and the names are the whole fix.

Also flagged in the body, **not fixed** (pre-existing, reproduces on `main` too): `pogo agent list` against an index-only daemon prints `json: cannot unmarshal object into Go value of type []agent.AgentInfo`, because `RequireOrchestration` returns a 503 JSON object and the client decodes into a slice.

drellem2:
**Rebased onto `main`** (no content change) — `cea5c8e` → `43b98f5`.

The branch was 3 commits behind: mg-6f5e (`refinery timestamps render as labelled UTC`) also touches `cmd/pogo/main.go`. The two regions don't overlap, so the rebase was clean, but I'd rather resolve that here than have it surface as a rebase conflict at merge time.

Re-ran `./build.sh` after the rebase: exit 0, zero failures. Diff against `main` is now exactly the 14 files of this change.

Doing it now because no review has started yet — a force-push under an in-progress review would be the wrong trade.

drellem2:
## Review round 1: PASS
Reviewer: mg-ab5d · build ticket: mg-4b01 · blocking: 0 · advisory: 4

Reviewed at `cea5c8e`, against the approved triage recommendation (mg-c23c, as
transcribed into the build ticket): clear the latch after the drain, add an
agent-side restart hook wired beside `SetRefineryStarter`, honour `[agents]
autostart = false` on the transition path, and correct the CLI message. All
four are delivered, and the regression tests run against a real
`agent.Registry` as required.

### Where the latch is cleared, and what is in flight on either side

`Registry.Resume()` is called at `internal/server/server.go:263` — after the
refinery restart succeeded and the mode flipped, before the auto-start sweep.
Verified independently of the PR's own argument:

- `StopAll` bumps the generation **before** it stops anything
  (`internal/agent/agent.go:1272-1273`).
- `Stop` blocks on `<-agent.done` (`internal/agent/agent.go:1234`), and
  `waitAndHandle` fires `onExit` **before** `close(a.done)`
  (`internal/agent/agent.go:1680-1686`). So every respawn goroutine the drain
  will ever schedule is already scheduled, holding the post-`StopAll`
  generation, by the time `StopAll` returns.
- `Resume` bumps again, so all of them lose at the generation check
  (`internal/agent/agent.go:1361-1365`).
- This is the common path, not an exotic one: `Agent.ShouldRespawn()` is
  `RestartOnCrash && !IsParked` (`internal/agent/park.go:180`) and does not
  consult `stopRequested`, so every `restart_on_crash` agent schedules a 2s
  respawn on every drain. Observed live in both arms of the control below:
  `agent mayor (crew) exited unexpectedly, scheduling restart` during the stop.
- On the later side: a crash **during** the sweep captures the new generation
  and is supervised, which clearing after the sweep would have broken.
- On the earlier side: a refinery failure returns before the mode flip and
  before `Resume` (`internal/server/server.go:223-232`), so the latch stays set
  on a transition that did not happen. `TestTransitionToFullAbortsWholeWhenRefineryFails`
  pins it.

The window is not hypothetical. In the branch run below, the drain's stale
respawn was refused at `01:57:09` and `Resume` plus the sweep landed at
`01:57:09` — inside the same second.

Also checked: `StopAll` has exactly one non-test caller,
`transitionToIndexOnly` (`internal/server/server.go:195`). There is no other
door that sets the latch without a matching `transitionToFull` to clear it.

### Verification

`./build.sh` on the PR head: **exit 0**, every Go package `ok`, including
`internal/agent` (280s) and `internal/refinery` (108s). Host load was 5–9
throughout, and `TestGateWatchMeasuresARealSubtreesCPU` passed, so the known
load-sensitive gate did not confound this run.

**CLI-level control, both arms.** Byte-identical driver script, one isolated
`pogod` on a private port with its own `HOME`/`XDG_CONFIG_HOME`/`POGO_HOME`,
guarded on the listener pid matching the daemon the script started; only the
binary differs (merge-base `500740b` vs PR head `cea5c8e`).

Pre-fix arm — the reported symptom, reproduced:

```
=== [main] pogo server start ===
Restarting orchestration...
Orchestration restarted
exit=0
=== [main] agents after restart ===
No running agents.
=== [main] daemon log ===
server: now in full mode
```

Green return, no agents, and `server: now in full mode` with no `autostart:`
line after it — the diagnostic signature from the issue.

Branch arm, same script:

```
=== [branch] pogo server start ===
Restarting orchestration...
Orchestration restarted
  Refinery: restarted.
  Crew agents started (3): mayor, scout, tinker
  Polecats are NOT restored — they are ephemeral and are re-dispatched per work item.
exit=0
=== [branch] agents after restart ===
mayor    pid=24511  type=crew  status=running  uptime=2s
scout    pid=24512  type=crew  status=running  uptime=2s
tinker   pid=24513  type=crew  status=running  uptime=2s
```

The agents are named by the CLI's own output and observed back with fresh pids.
The exit code is `0` in both arms, which is why it proves nothing here.

**The latch half, at CLI level.** Second control: complete the round-trip, then
`kill -9` a `restart_on_crash` crew agent and wait 6s.

- Pre-fix: `agent scout: restart failed: registry shut down`; scout stays
  `status=exited`. (Hand-starting it first still worked — the `Spawn`/`Respawn`
  asymmetry, reproduced.)
- Branch: `agent scout: respawned pid=44317 restart=1`; scout is running again.

**Positive control on the tests.** Neutered the fix in place on the PR head
(removed the `Resume()` call and the sweep, kept the API):
`TestStartOrchestrationRestartsCrewAgents`,
`TestStartOrchestrationClearsShutdownLatch`,
`TestStartOrchestrationHonoursAutoStartDisabled` and
`TestTransitionToFullReportsFailedAgents` all fail. Restored; tree clean.

**Race detector.** `go test -race ./internal/server` (full) and
`go test -race ./internal/agent -run 'Resume|Respawn|Generation|StopAll|AutoStart'`
— both clean.

**Scope.** `transitionToIndexOnly` is untouched, so drain semantics are
unchanged; the PR acknowledges the "drain counts polecats, compute can outlive
the polecat" coupling in a paragraph and explicitly declines the reserved
question. That is the right side of the line.

### Advisory (non-blocking)

- `internal/client/client.go:26` — `internal/client` now imports
  `internal/server` for the `StartReport` DTO. No cycle, and `cmd/pogo` already
  links `internal/agent` and `internal/refinery`, so it costs nothing at build
  time. But the existing precedent for types that cross the wire is a
  standalone package (`internal/health`, `internal/project`); a client
  depending on the server package inverts that. Worth a follow-up if the report
  grows more fields.
- `cmd/pogo/main.go:244` — `pogo server start` now exits non-zero when a crew
  agent errored on spawn. The recommendation asked only that the message name
  what was and was not restored; the exit-code change is beyond it. The PR
  flags it and offers a two-line revert. It serves the "reports success anyway"
  complaint and is documented rather than silent, so it does not block — but it
  changes a scriptable command's contract, and that call belongs to pm-pogo.
- `internal/agent/autostart.go:92` — a prompt whose frontmatter fails to parse
  is classified `AutoStartStatusFailed`, so a malformed prompt file now prints
  `Crew agent FAILED to start` and (via the above) exits non-zero. Consistent
  with how the boot path classifies it; noted because the new exit code
  inherits it.
- `internal/agent/agent.go:1339` — `Respawn` now has no non-test production
  callers; pogod's `OnExit` uses `RespawnFromGeneration`. The unguarded entry
  point remains exported and is exercised only by tests.

### Not checked

The refinery restart path itself (unchanged by this PR, exercised only as far
as the `Refinery: restarted.` line); `[agents] autostart = false` and parked
agents end-to-end through the CLI (covered at unit level only); polecat
behaviour across a round-trip (no polecats in the sandbox); `-race` across the
whole tree (only the two packages above); and whether any shipped prompt or doc
text now contradicts the new `pogo server start` output.

drellem2:
**SHA note for the record** — the round-1 review cites merge-base `500740b` vs **`cea5c8e`**, which was this PR's head *before* the rebase. Current head is **`43b98f5`**.

Checked rather than asserted:

```
git diff cea5c8e 43b98f5 -- <the 14 files of this change>   -> empty
git diff cea5c8e 43b98f5 --name-only                        -> only mg-6f5e's files
```

The change under review is byte-identical between the two. The entire delta is mg-6f5e (`refinerytime.go`, `refineryhistory.go`, and the `formatHistoryRow`/`refineryTimeSecond` extraction in the refinery region of `cmd/pogo/main.go` — nowhere near the `server start` region this PR touches), which is already merged to `main` and was reviewed on its own.

So the verdict applies to the content here; the only thing the review's build did not contain is code that is already on `main`. Recording it because a verdict pinned to a SHA that is no longer the head is cheap to state now and expensive to reconstruct later.

Also worth surfacing from the review's advisory list, since it argues **against** a choice I made in this PR: `internal/agent/autostart.go:92` classifies a malformed prompt frontmatter as `AutoStartStatusFailed`, which means the non-zero exit I added would also fire on a **typo in a crew prompt file** — a worse consequence than the "a crew agent errored on spawn" case I was reasoning about when I chose it. That decision is routed to pm-pogo; I'll change it immediately either way, and I'm not editing under a PASS unasked.

drellem2:
## Review round 1, addendum: re-verified at the current head

The branch was rebased between my checkout and my verdict. The review above was
run at `cea5c8e` (base `500740b`); the head is now `43b98f5` (base `1fe0177`).
The verdict is unchanged, but it should rest on the head that will actually
merge, so here is what I re-ran.

**What the rebase changed.** The PR's own patch is identical across it. Of the
14 touched files, 13 are byte-identical between the two heads. The 14th,
`cmd/pogo/main.go`, differs only because `main`'s own commits also touched that
file — this PR's hunks in it are identical, verified by diffing the two patches
directly. The merged `pogo server start` block is intact
(`orchestrationRestartSummary` at `cmd/pogo/main.go:228`,
`orchestrationRestartLines` at `:233`, the `AgentsFailed` exit at `:243`).

**Gate at `43b98f5`:** `./build.sh` green, every Go package `ok`. `cmd/pogo` —
the one package whose content the rebase changed — re-ran from scratch (82.8s)
and passed. `internal/agent`, `internal/client` and `internal/server` came back
cached, which is itself the evidence that their inputs are byte-identical to
what I reviewed; the `-race` runs reported above therefore apply verbatim.

**Both CLI controls re-run against `43b98f5` binaries.** Same driver, same
isolated sandbox, same guard.

```
=== [head] pogo server start ===
Orchestration restarted
  Refinery: restarted.
  Crew agents started (3): mayor, scout, tinker
exit=0
=== [head] agents after restart ===
mayor    pid=874  scout    pid=875  tinker   pid=877   (all running)
```

Latch control at the head: after the round-trip, `kill -9` on the crew agent
gives `agent scout: respawned pid=6301 restart=1` and it is back in `agent
list`. Unchanged from what I saw at `cea5c8e`.

Verdict stands: **PASS**, 0 blocking, the 4 advisories above.

drellem2:
Merged as 3def26f3425baa9640c41090c7902aa43305ab7a on `main` by the pogo refinery (MR mr-d9qj7oatjv1h244d85r0).

PR #120, opened 2026-08-07
test.sh: bound go test with an explicit 20m budget and REPORT an overrun as a timeout, not a panic dump (gh#107)

### LINKED ISSUE #107: test.sh runs `go test ./...` with no -timeout, so a slow package dies at Go's 600s default with a panic dump that reads like a mass regression
test.sh:14 is:

    go test ./...

With no `-timeout`, every test binary inherits Go's default of 600s. Two packages in this
repo are already close enough to that ceiling to cross it on a loaded host:

    internal/refinery   538.861s   (prebuild-verify run, 2026-07-31 15:00)
    internal/agent      508.840s   (same tree, measured 2026-07-31 ~15:20)

Both are within ~12% of 600s, and neither number is a worst case — they were measured on a
host also running an agent fleet, and package runtime here scales with load.

WHY THIS IS WORSE THAN AN ORDINARY TIMEOUT

When a package crosses the limit, Go does not report a tidy failure. It panics with
`panic: test timed out after 10m0s` and dumps the stacks of every goroutine in the binary.
For a package like internal/agent — which spawns real processes, PTYs and registries — that
is hundreds of lines of stack across dozens of goroutines. It does not look like "one package
ran long"; it looks like a mass regression, and the tests that had already passed inside that
binary are reported as failures because the whole binary died.

That is expensive in exactly the way that matters: it is indistinguishable at a glance from a
real defect, so it consumes a full triage cycle before anyone discovers the run simply needed
more wall clock. (Observed adjacent to this: the guidance already circulating by hand for
this tree is "use `-timeout 40m`", which is the workaround for a default the script itself
never sets.)

BLAST RADIUS IS WIDER THAN PREBUILD VERIFICATION

This is not only a manual-verification concern. The refinery merge gate defaults to
`./build.sh` (internal/refinery/merge.go:680), and build.sh calls `./test.sh` (build.sh:59),
which is this line. So a package crossing 600s fails a MERGE, and the failure arrives as a
panic dump attributed to the branch under test rather than to the clock.

SUGGESTED FIX

One line — set the timeout explicitly rather than inheriting a default nobody chose:

    go test -timeout 40m ./...

Two things worth considering alongside it:
  - The two packages above are slow enough to be worth their own look; a 40m ceiling buys
    headroom but does not make an 8-9 minute package fast.
  - If a per-package budget is ever wanted, `-timeout` is per test binary, so a single flag
    already applies to each package independently.

---
_Found by triage agent `ede2` (mg-ede2) while triaging two prebuild-verify failures. Related: #106 (the other failure in that same run turned out to be a launcher artifact)._

drellem2:
`test.sh:14` ran a bare `go test ./...`, so every test binary inherited Go's default
10-minute per-package timeout — a number nobody chose — and it reached the merge gate:
`build.sh:60` runs `./test.sh`, and the refinery gate defaults to `./build.sh`
(`internal/refinery/merge.go`, no `.pogo/refinery.toml` to override).
`.github/workflows/ci.yml:49` carried the identical bare invocation on slower runners.

## The measurement that changed the shape of the fix

The issue asks for one line, `go test -timeout 40m ./...`, on the premise that setting the
flag makes a slow package "fail cleanly" instead of panicking. **I measured that premise
before writing anything, and it does not hold: `-timeout` IS the panic mechanism.** Go arms
`testing.(*M).startAlarm`; the default 10m and an explicit 20m take the identical path, and
both end in `panic: test timed out after <budget>` followed by the stacks of every goroutine
in the binary.

So the flag alone moves the panic later without changing its shape — and the shape is the
entire complaint in #107. Setting the budget is **necessary but not sufficient**.

This is the one place this PR departs from the approved triage recommendation (mg-d2f0),
which specified a two-line change. The 20m value, the reasoning for it, and the "both call
sites" requirement are all taken from that recommendation unchanged. What is added is the
half that makes the reported outcome a timeout, because the two lines alone do not — and the
mutation control below shows exactly what that world looks like. **Flagging it explicitly
for the design-faithfulness lens.**

## The change

`scripts/go-test-budget.sh` runs `go test` under an explicit budget and then *classifies*
the outcome, emitting at the **tail** of the output — where a reader looks, and what a
truncating CI log keeps — the package that overran, the budget Go actually enforced, the
tests still running when it fired, and the statement that this is an overrun and not a crash.
Anchored on Go's exact timeout-panic text, so an ordinary `panic: runtime error: ...` and a
plain `t.Fatal` are never mislabelled.

`test.sh` and `ci.yml` both route through it: one definition of the budget, one of the report.

**The goroutine dump is kept, deliberately.** Suppressing it is the obvious way to "not be a
panic dump", and it would destroy the only evidence separating the two causes an overrun can
have, which need opposite responses — a genuine deadlock (goroutines parked on
chan/select/mutex, no progress) versus a loaded host (progressing, just not fast enough).
Those are identical in a bare "package X exceeded N minutes" line, and telling them apart is
why the reader is there. It is *labelled*, not removed, and the label is printed last.

## Why 20m

`defaultGateTimeout` is 60m (`internal/refinery/gaterun.go:30` — verified, not inherited),
documented there as roughly twice the longest gate run observed on this fleet (~30m). A 40m
per-package ceiling is larger than the entire observed gate run: one wedged package would
consume two-thirds of the gate budget before reporting anything. 20m surfaces a wedged
package 20 minutes sooner and leaves the 60m gate bound reachable as the outer backstop —
which kills with a diagnostic heartbeat record rather than a panic dump. Keeping the
better-instrumented backstop reachable is choosing a better failure, not merely an earlier one.

**What a normal run measures, and the headroom left for a loaded host.** Too short a budget
converts load into spurious failures, which is precisely the disease mg-6c90 documents (a
fixed CPU floor: 13/13 FAIL at load 52-106, 4/4 PASS at load 4.6-5.3, byte-identical
binaries). Measured on this host:

| package | time | condition |
|---|---|---|
| `internal/agent` | 206.3s | isolated, load ~7.8 |
| `internal/agent` | 216.8s | full 50-package suite, load ~6 |
| `internal/refinery` | 81.2s | isolated, load ~7.8 |
| `internal/refinery` | 155.1s | isolated, load ~32 — **1.9x on load alone** |

20m is **5.5x** the slowest figure measured here and **2.2x** the worst figure #107 reports
(538.9s). The 538.9s/508.8s in the issue were measured under a live agent fleet and were not
reproduced on this box; that corroborates rather than refutes #107, whose own claim is that
runtime scales with contention — the 81.2s→155.1s response measures that directly.

### Positive control: a package made to exceed the budget on purpose

Run at `POGO_GO_TEST_TIMEOUT=5s` (the budget is overridable so this control costs seconds
rather than 20 minutes). 80 lines; the first 8 and last 26:

```
Testing Go packages (per-package budget: 5s; override with POGO_GO_TEST_TIMEOUT)
panic: test timed out after 5s
	running tests:
		TestSlowA (5s)

goroutine 3 [running]:
testing.(*M).startAlarm.func1()
	.../src/testing/testing.go:2682 +0x2b0
   ...  [46 lines of goroutine dump] ...

=============================================================================
TIMEOUT: 1 package exceeded the per-package test budget of 5s.

  example.com/multi/slowa
      budget enforced by go: 5s
      still running at the budget: TestSlowA

This is a BUDGET OVERRUN, not a crash. Go implements -timeout by panicking
inside the test binary, so the goroutine dump above IS what a timeout looks
like — it is not a segfault and not a mass regression. Tests that had already
passed inside those binaries were lost with it, so the run is not evidence
about them either way.

The dump is kept, not suppressed: it is the only thing separating the two
causes, which need opposite responses —
  deadlock     goroutines parked on chan/select/mutex with no progress
  loaded host  goroutines progressing, just not fast enough
Check the host load average for the window of the run before reading the
dump as a bug. On this host the same package has measured 81.2s at load ~8
and 155.1s at load ~32 — a 1.9x response to contention alone.

Raise 5s deliberately, not reflexively: the refinery's 60m gate bound
is the outer backstop, and a per-package budget that approaches it stops
leaving room for the better-instrumented failure to happen instead.
=============================================================================
```

It names the package, names the budget, and the last thing the reader is handed is the
classification — not the dump. Exit status is `go test`'s own, unchanged.

### Normal run unaffected

Full `./build.sh`, green end-to-end (`BUILD_EXIT=0`), from this non-ephemeral worktree — a
bare `go test` gives false results in this repo. No package came near the budget, nothing was
reported, and the banner confirms the flag is live:

```
Step 2: Testing...
Testing Go packages (per-package budget: 20m; override with POGO_GO_TEST_TIMEOUT)
...
=== go-test-budget.sh: 23 passed, 0 failed ===
```

Also worth recording, since the triage notes warned about it: `internal/agent`'s
`TestPolecatDoesNotOutlivePogod` — which failed in both bare-`go test` triage runs — **passed
here** through the `./test.sh` envelope. No evidence of a red main branch.

### The controls have teeth (mutation)

An assertion that cannot fail is not a control, so both were shown failing:

| mutation | result |
|---|---|
| classifier's anchor made to never match — i.e. **the `-timeout`-alone world** | 7 assertions fail; output ends in a bare `FAIL` after the dump |
| `test.sh` unwired back to `go test ./...` | the 2 wiring assertions fail |

The first is the important one: it *is* the two-line fix, and it is red.

## The suite

`scripts/go-test-budget_test.sh`, 23 assertions, 12.1s (5s of it deliberate sleeping):
positive control; the report is the tail and the dump is still present; **negative controls**
(a nil-deref panic and a `t.Fatal` in the same run are not called timeouts — without these, a
classifier that shouted TIMEOUT at every failure would pass); the configured budget is the
one Go reports (proving the flag is not silently dropped); the 20m default; a caller-supplied
`-timeout` refused; and wiring assertions pinned against the **real** `test.sh` and `ci.yml`,
so the fix cannot be silently unwired later.

It adopts the packaged test isolation (`scripts/pogo-sandbox`), which the repo's own
`TestEveryTestSuiteRoutesThroughTheIsolation` ratchet correctly caught me not doing on the
first gate run. No ledger line was added — the ledger only shrinks.

## Noted, deliberately not folded in

- `scripts/upgrade-smoke.sh:348` runs `go test ./internal/agent -run ... ` with **no
  timeout** — a third unbounded site the issue does not name. It is not on the merge-gate
  path and targets a single test, so I left it alone rather than widen scope. The CI/test.sh
  wiring assertions are scoped to the merge-gate path for the same reason. Flagging it for a
  separate ticket if wanted.
- The two slow packages are still slow. A ceiling buys headroom; it does not make an
  8-minute package fast. Explicitly out of scope per the ticket — that is a different ticket,
  and it will be a different package tomorrow.

Refs drellem2/pogo#107
Resolves drellem2/pogo#107
Closing-ref-ack: drellem2/pogo#107 — this PR is the approved fix for that issue, so closing it on merge is intended.

Approved triage recommendation: mg-d2f0 (successor item mg-a465; pm-pogo signed off on the verdict, the 20m value and the ci.yml inclusion; Daniel gave the go/no-go)
Work item: mg-a465

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,18 @@ jobs:
 
   test:
     runs-on: ubuntu-latest
+    # The outer backstop, so the per-package budget is not the only bound here
+    # (raised in review as mg-a3db advisory 1). Under the refinery, test.sh sits
+    # inside defaultGateTimeout=60m, which kills with a diagnostic heartbeat
+    # record rather than a panic dump — that argument is why 20m per package is
+    # safe to choose. It did NOT transfer to this call site: outside the gate
+    # there was nothing but GitHub's 360-minute job default, so raising the
+    # per-package worst case from 10m to 20m widened an unbounded window.
+    #
+    # 45m: comfortably above the realistic worst case (one wedged package at its
+    # 20m budget plus the rest of a suite that measures ~4m here), and an order
+    # of magnitude below the 360m default, which is "no bound" in practice.
+    timeout-minutes: 45
     steps:
       - uses: actions/checkout@v4
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,8 +45,14 @@ jobs:
           mkdir -p _testdata/a-service/.git
           mkdir -p _testdata/b-service/.git
 
+      # Bounded and classified by the same script test.sh uses (mg-a465,
+      # gh#107) — one definition of the budget and one definition of the
+      # report. CI is the MORE exposed of the two: these runners are slower
+      # than the dev host, and nobody watches this one interactively, so an
+      # unlabelled goroutine dump here is read by whoever opens the failed run
+      # days later.
       - name: Test
-        run: go test ./...
+        run: bash scripts/go-test-budget.sh ./...
 
   pi-e2e:
     # Live pi-provider integration tests (mg-3e7c): drive a real pi binary
--- a/changelog.d/mg-a465.fixed.md
+++ b/changelog.d/mg-a465.fixed.md
@@ -1,0 +1,81 @@
+- **A Go package that exceeds its test budget is now REPORTED as a budget
+  overrun, naming the package and the budget, instead of ending the run in an
+  unlabelled goroutine dump (mg-a465, gh#107).** `test.sh` ran a bare
+  `go test ./...`, so every test binary inherited Go's default 10-minute
+  per-package timeout — a number nobody chose — and it reached the merge gate:
+  `build.sh` runs `test.sh`, and the refinery gate defaults to `./build.sh`.
+  `.github/workflows/ci.yml` carried the identical bare invocation on slower
+  runners.
+
+  **The part of the diagnosis that changed the fix: `-timeout` IS the panic.**
+  The issue frames this as "dies at Go's 10-minute panic instead of failing
+  cleanly", with the implied remedy that setting `-timeout` makes it fail
+  cleanly. Measured before anything was written, it does not. Go implements
+  `-timeout` by arming `testing.(*M).startAlarm`, and the default 10m and an
+  explicit 20m take the identical code path — both end in
+  `panic: test timed out after <budget>` followed by the stacks of every
+  goroutine in the binary. So the two-line change the issue asks for moves the
+  panic later without changing its shape, and the shape is the entire
+  complaint: a package that merely ran long stays indistinguishable at a glance
+  from a mass regression, and the tests that had already passed inside that
+  binary are still lost.
+
+  Setting the budget is therefore necessary but not sufficient, and the second
+  half is what `scripts/go-test-budget.sh` adds. It runs `go test` under an
+  explicit budget and then CLASSIFIES the outcome, emitting at the tail of the
+  output — where a reader looks and what a truncating CI log keeps — the
+  package that overran, the budget Go actually enforced, the tests still
+  running when it fired, and the statement that this is an overrun and not a
+  crash. The classification is anchored on Go's exact timeout-panic text, so an
+  ordinary `panic: runtime error: ...` and a plain `t.Fatal` are never
+  mislabelled; both are negative controls in the suite.
+
+  **The goroutine dump is kept, deliberately.** Suppressing it is the obvious
+  way to "not be a panic dump" and it would destroy the only evidence that
+  separates the two causes an overrun can have, which need opposite responses:
+  a genuine deadlock (goroutines parked on chan/select/mutex, no progress)
+  versus a loaded host (goroutines progressing, just not fast enough). Those
+  are identical in a bare "package X exceeded N minutes" line, and telling them
+  apart is why the reader is there. The dump is labelled rather than removed,
+  and the label is the last thing printed.
+
+  **Why 20m and not the 40m the issue suggests.** `defaultGateTimeout` is 60m
+  (`internal/refinery/gaterun.go:30`), documented there as roughly twice the
+  longest gate run observed on this fleet (~30m). A 40m per-package ceiling is
+  larger than the entire observed gate run, so one wedged package would consume
+  two-thirds of the gate budget before reporting anything. 20m is ~2.2x the
+  worst per-package figure #107 reports (538.9s), surfaces a wedged package 20
+  minutes sooner, and leaves the 60m gate bound reachable as the outer
+  backstop — which kills with a diagnostic heartbeat record rather than a panic
+  dump. Keeping the better-instrumented backstop reachable is choosing a better
+  failure, not merely an earlier one.
+
+  **The headroom is the thing not to get wrong.** Too short a budget converts
+  host load into spurious failures, which is the disease mg-6c90 documents — a
+  fixed CPU floor that failed 13/13 at load 52-106 and passed 4/4 at load
+  4.6-5.3 on byte-identical binaries. Measured on this host, the slowest
+  packages are `internal/agent` at 206-217s and `internal/refinery` at 81.2s
+  isolated, rising to 155.1s at load avg ~32 — a 1.9x response to contention
+  alone. 20m is 5.5x the slowest figure measured here. (The 538.9s/508.8s in
+  #107 were measured under a live agent fleet and were not reproduced here;
+  that corroborates rather than refutes the issue, whose own claim is that
+  runtime scales with contention.)
+
+  **Both call sites, and a test that says so.** `test.sh` and `ci.yml` route
+  through the same script, so there is one definition of the budget and one of
+  the report; CI is the more exposed of the two, since those runners are slower
+  and nobody watches them interactively. `scripts/go-test-budget_test.sh`
+  asserts the wiring against the real `test.sh` and `ci.yml` — that both go
+  through the script and that neither carries an unbounded `go test` — because
+  a partial fix invites the wrong inference: someone who learns "test.sh has a
+  timeout now" will assume the suite is bounded everywhere. The suite's
+  load-bearing case is the positive control, which makes a package exceed the
+  budget on purpose and asserts the report names it; with the classifier
+  disabled that control fails 7 assertions and the output ends in a bare
+  `FAIL`, which is precisely the pre-fix world.
+
+  A caller-supplied `-timeout` is refused with a usage error rather than
+  silently winning (Go takes the last one), so the enforced budget and the
+  reported budget cannot disagree. `POGO_GO_TEST_TIMEOUT` overrides the
+  default — which is what lets the positive control cost 12s instead of 20
+  minutes, since a control nobody can afford to run is not a control.
--- a/changelog.d/mg-a465.fixed.md
+++ b/changelog.d/mg-a465.fixed.md
@@ -50,16 +50,50 @@
   dump. Keeping the better-instrumented backstop reachable is choosing a better
   failure, not merely an earlier one.
 
-  **The headroom is the thing not to get wrong.** Too short a budget converts
-  host load into spurious failures, which is the disease mg-6c90 documents — a
-  fixed CPU floor that failed 13/13 at load 52-106 and passed 4/4 at load
-  4.6-5.3 on byte-identical binaries. Measured on this host, the slowest
-  packages are `internal/agent` at 206-217s and `internal/refinery` at 81.2s
-  isolated, rising to 155.1s at load avg ~32 — a 1.9x response to contention
-  alone. 20m is 5.5x the slowest figure measured here. (The 538.9s/508.8s in
-  #107 were measured under a live agent fleet and were not reproduced here;
-  that corroborates rather than refutes the issue, whose own claim is that
-  runtime scales with contention.)
+  **The headroom, stated for a loaded host and not just a quiet one.** Too
+  short a budget converts host load into spurious failures, which is the
+  disease mg-6c90 documents — a fixed CPU floor that failed 13/13 at load
+  52-106 and passed 4/4 at load 4.6-5.3 on byte-identical binaries. So the
+  quiet-host multiple is the wrong headline: this host has recorded load 174 in
+  a single night.
+
+  On a quiet box the slowest package, `internal/agent`, measures **268.7s
+  isolated at load ~5 on this branch** — 4.5x under the budget. (The 206-217s
+  and 81.2s/155.1s figures in the ticket are the *triage* measurement set,
+  recorded before this branch and now optimistic; they are attributed to that
+  run rather than to this one.)
+
+  Loaded, the margin is much thinner, and this part is an **estimate rather
+  than a measurement**. The one load-response pair measured here —
+  `internal/refinery` 81.2s@~7.8 → 155.1s@~32 — fits runtime ~ load^0.46;
+  extrapolating `internal/agent` from the three available anchors gives:
+
+  | anchor | load 100 | load 150 | load 174 |
+  |---|---|---|---|
+  | triage, full suite (216.8s@~6) | 13.1m · 1.5x | 15.8m · 1.3x | 16.9m · 1.2x |
+  | review, this branch (279.6s@~10) | 13.4m · 1.5x | 16.1m · 1.2x | 17.3m · 1.2x |
+  | this branch, isolated (268.7s@~5) | 17.7m · 1.1x | 21.3m · 0.9x | 22.8m · 0.9x |
+
+  **So the margin at load ~100 is roughly 1.1-1.5x, and at the load 150-174
+  this box has actually recorded the slowest package may legitimately exceed
+  20m.** A spurious timeout on `internal/agent` under heavy fleet load is an
+  anticipated outcome of this budget, not a sign that something else broke —
+  which is exactly what the report this change prints tells the reader to check
+  first.
+
+  The exponent is rough and is labelled as such: a two-point fit on one
+  package, pushed ~5x past its measured range and applied to a different
+  package. This branch's own third point does not fit it — `internal/refinery`
+  measured 99.2s at load ~5 here against 81.2s at load ~7.8 at triage, lower
+  load and higher time. Load average does not separate CPU contention from I/O
+  wait, and these suites spawn processes and PTYs. The direction and rough size
+  are not in doubt; the exponent is.
+
+  20m is kept anyway. Buying margin for load 174 spends gate budget on a state
+  where the box is failing at everything, and the 60m gate bound is what should
+  catch that. (The 538.9s/508.8s in #107 were measured under a live agent fleet
+  and were not reproduced here; that corroborates rather than refutes the
+  issue, whose own claim is that runtime scales with contention.)
 
   **Both call sites, and a test that says so.** `test.sh` and `ci.yml` route
   through the same script, so there is one definition of the budget and one of
--- a/scripts/go-test-budget.sh
+++ b/scripts/go-test-budget.sh
@@ -57,14 +57,55 @@ set -e
 #   better-instrumented backstop is choosing a better failure, not merely an
 #   earlier one.
 #
-#   Headroom against a loaded host is the thing not to get wrong: too short a
-#   budget converts load into spurious failures, which is the disease mg-6c90
-#   documents (a fixed CPU floor that fails at load 52-106 and passes at load
-#   ~5, on byte-identical binaries). Measured on this host, the slowest
-#   packages are internal/agent at 206-217s and internal/refinery at 81.2s
-#   isolated, rising to 155.1s at load avg ~32 — a 1.9x response to contention
-#   alone. 20m is 5.5x the slowest figure measured here and 2.2x the slowest
-#   figure #107 reports from a live fleet.
+#   HEADROOM, AND WHAT IT IS ON A LOADED HOST — read this before changing 20m.
+#
+#   Too short a budget converts load into spurious failures, which is the
+#   disease mg-6c90 documents (a fixed CPU floor: 13/13 FAIL at load 52-106,
+#   4/4 PASS at load 4.6-5.3, byte-identical binaries). So the number that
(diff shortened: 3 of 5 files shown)


drellem2:
## Review round 1: FAIL
Reviewer: mg-a3db · build ticket: mg-a465 · blocking: 1 · advisory: 4

Reviewed against the approved recommendation in mg-a465 (successor to triage mg-d2f0), at head
`8cf5eeb`, from a non-ephemeral worktree.

### What I ran

- `./build.sh` on the PR head: **green, `BUILD_EXIT=0`, 7:27 total.** `go-test-budget_test.sh`
  reported 23 passed, 0 failed inside it. Host load over the run: samples 3.0, 8.5, 24.9, 15.1,
  8.7, 6.1, 7.5 — mean around 10, and the peak was partly my own concurrent probe.
- **The PR's mutation control, re-run independently.** I copied `scripts/`, `test.sh` and `ci.yml`
  to a tree outside the repo, made the classifier's anchor regex unmatchable, and ran the suite
  against it: 7 assertions fail and the output ends in a bare `FAIL`, exactly as the PR states.
  That is the `-timeout`-alone world and it is red. One of the seven has no teeth — see advisory 2.
- **An adversarial positive control of my own**, at `POGO_GO_TEST_TIMEOUT=10s`: three packages
  overrunning concurrently, one of them with two `t.Parallel()` tests, and one printing a
  counterfeit `FAIL\t<pkg>\t<elapsed>` line and a counterfeit two-tab "running test" line from
  inside a hanging test. The classifier reported `TIMEOUT: 3 packages`, named all three, named both
  parallel tests, and did not ingest either counterfeit line. The report was the tail.

The central premise of the departure from the two-line recommendation holds and I confirmed it
directly: `go test -timeout` panics. The recommendation's answer to the report problem — that 20m
"leaves the 60m gate bound in play as the outer backstop, which kills with a diagnostic heartbeat
record rather than a panic dump" — does not reach the case #107 is about. A single wedged package
trips its own 20m alarm long before the gate's 60m bound, so the panic is what you get; the gate
backstop is reachable only when the *aggregate* exceeds 60m with no single package exceeding 20m.
On the merits the added classifier is a correction to the recommendation, not scope creep. Whether
580 lines against an approved two-line change is this PR's call to make is a governance question,
routed to the mayor rather than ruled on here.

**20m is the right number** and I am not asking for it to change. It is strictly 2x more headroom
than main's status quo, so it cannot make load-induced flakiness worse than today; when it does
fire it fires labelled and tells the reader to check load average, which is the countermeasure
mg-6c90's disease lacked; and 40m would put a wedged package's first report past two-thirds of the
refinery's 60m gate budget.

### Blocking

- `scripts/go-test-budget.sh:60-67` (and `changelog.d/mg-a465.fixed.md:53-62`) — **the headroom is
  stated for a quiet host only, and the quiet-host figure is the one that travels.** "20m is 5.5x
  the slowest figure measured here" sits in the script header, which is exactly where someone goes
  before changing the number. The 1.9x response to a 4x load increase is recorded two sentences
  earlier and the two are never composed — and composing them is the specific thing the ticket
  asked for. On the PR's own data (81.2s at load ~7.8 → 155.1s at load ~32, an exponent of ~0.46),
  the slowest package extrapolates to roughly 13 minutes at load 100 and roughly 17 minutes at load
  174 — margins of about **1.6x and 1.2x**, not 5.5x. This host has recorded load 174 tonight.

  Two related gaps in the same paragraph. First, the four-row table is the triage measurement set
  from mg-a465 reproduced verbatim under the heading "Measured on this host"; it is not a
  measurement taken on this branch, and it should say so. Second, the figures it carries are low
  for this branch: my `./build.sh` on the PR head measured `internal/agent` at **279.6s** (table:
  216.8s) and `internal/refinery` at **183.1s** (table: 81.2s isolated, 155.1s at load ~32). 279.6s
  against a 20m budget is 4.3x, so even the quiet-box headline overstates by a fifth.

  Expected: the script header and the changelog state the loaded-host margin alongside the 5.5x,
  attribute the table to the triage run, and carry at least one figure measured on this branch. The
  point is not that 20m is wrong — it is that a reader who later sees a spurious timeout at load
  150 should find it recorded as anticipated and correctly labelled, rather than reading 5.5x and
  concluding something else broke.

### Advisory (non-blocking)

1. `.github/workflows/ci.yml:34` — the `test` job has no `timeout-minutes`, so the "60m gate bound
   stays reachable as the outer backstop" argument, sound for `test.sh` under the refinery, does
   not transfer to the CI call site. There the per-package worst case just went 10m → 20m with
   nothing outside it but GitHub's 360-minute job default. Not a regression in kind — CI had no
   outer bound before either — but a `timeout-minutes:` on that job would remove the asymmetry
   instead of leaving it implicit.
2. `scripts/go-test-budget_test.sh:201` — "the report names the package that overran" greps all of
   `$OUT`, which go's own `FAIL\t<pkg>` line already satisfies. My mutation run confirms it: this
   assertion passes with the classifier disabled, so it constrains nothing. Scoping the grep to the
   lines after the `TIMEOUT:` header would give it teeth. The other six mutation failures are
   load-bearing and I saw them red.
3. `scripts/go-test-budget.sh:86` — `POGO_GO_TEST_TIMEOUT` is read from the ambient environment
   with no validation, so a value exported anywhere in a gate or CI environment silently redefines
   the merge gate's budget, including to something far too small. The banner printing it is most of
   the mitigation; a refusal below some floor would close the rest.
4. Scope is correct and worth saying so rather than raising as a gap: `scripts/upgrade-smoke.sh:348`
   is untouched. It is off the merge-gate path and leaving it alone was the right call.

### What I did not check

- CI. Everything above ran on darwin/arm64 with go1.25.0. `mktemp` portability and the awk
  classifier on ubuntu-latest are asserted by comment, not by execution.
- Output interleaving between concurrently-running packages. My probe did not break the classifier,
  but go's per-package output buffering is what makes that safe and I did not try to defeat it.
- `internal/agent` and `internal/refinery` in isolation — I measured them only inside the full
  55-package suite.
- An actual overrun of the shipped 20m budget. Every overrun I observed was at 2s, 3s, 5s or 10s.
- The loaded-host margins quoted above are an **extrapolation roughly 10x beyond the measured load
  range**, on a two-point fit taken from a different package than the one it is applied to. They
  are not a measurement, and they are the builder's own data pushed forward, not an independent
  one.

drellem2:
## Round 1 addressed — head `274fd56`

Thanks for reproducing the mutation independently and for the counterfeit-`FAIL`-line probe;
that is a shape I did not test and I have adopted your finding about it below.

### BLOCKING — loaded-host headroom (`scripts/go-test-budget.sh` header, `changelog.d/mg-a465.fixed.md`)

Fixed, and **your extrapolation moved further in your direction once I measured a third
anchor on this branch.**

I checked your working rather than taking it. `k = ln(155.1/81.2)/ln(32/7.8) = 0.458` — your
0.46. Two independent anchors converge:

| anchor | load 100 | load 150 | load 174 |
|---|---|---|---|
| triage, full suite (216.8s@~6) | 13.1m · 1.5x | 15.8m · 1.3x | 16.9m · 1.2x |
| your branch run (279.6s@~10) | 13.4m · 1.5x | 16.1m · 1.2x | 17.3m · 1.2x |
| **this branch, isolated `-count=1` (268.7s@~5)** | **17.7m · 1.1x** | **21.3m · 0.9x** | **22.8m · 0.9x** |

The third row is new: `internal/agent`, isolated, uncached, measured on this head. Projected
from it, the slowest package **can legitimately exceed the 20m budget** at loads this box has
actually recorded tonight. That is stronger than "a real margin, not 5.5x", and the header now
says it outright — a spurious `TIMEOUT: internal/agent` under heavy fleet load is an
*anticipated* outcome of this budget, not evidence something else broke.

**Provenance and staleness** — both correct, both fixed. The four-row table is now attributed
to the triage run rather than presented as branch data, with branch figures carried alongside.
Your staleness call was right: `internal/agent` 268.7s here vs 216.8s at triage;
`internal/refinery` 99.2s vs 81.2s.

**Where I qualify your estimate rather than dispute it** — you asked for working if I
disagreed, so: I agree with the direction and the rough size, and I have labelled the exponent
an estimate with its limits stated (two-point fit, one package, ~5x past its measured range,
applied to a different package). One additional caveat from my own data: **this branch's third
point does not fit the model** — `internal/refinery` measured 99.2s at load ~5 here against
81.2s at load ~7.8 at triage, i.e. *lower* load and *higher* time. Load average does not
separate CPU contention from I/O wait, and these suites spawn processes and PTYs. If anything
that argues the true margin at high load is thinner than a smooth power law suggests, since
queueing effects worsen near saturation — so the caveat cuts toward your conclusion, not away.

20m is unchanged.

### ADVISORY 2 — the assertion that constrained nothing

You were right and this was the real defect in the suite. `grep`ping all of `$OUT` for the
package name is satisfied by go's own `FAIL\t<pkg>` line. Now scoped to the classifier's own
output only (`sed -n '/^TIMEOUT: /,$p'`), along with the three sibling assertions.

Confirmed by re-running the mutation: **8 assertions now go red where 7 did**, and the new one
is exactly `the report body does not name example.com/budgetfixture/slowpkg`.

### ADVISORY 1 — `ci.yml` outer bound

Agreed, the asymmetry was real: outside the refinery there was nothing but GitHub's 360m
default, so 10m → 20m per package widened an effectively unbounded window. `timeout-minutes: 45`
on the test job — above the realistic worst case (one wedged package at 20m plus a suite that
measures ~4m there), an order of magnitude below 360m.

### ADVISORY 3 — ambient `POGO_GO_TEST_TIMEOUT`

Agreed. Now parsed as a Go duration and **refused below a 60s floor** unless
`POGO_GO_TEST_BUDGET_ALLOW_SHORT=1`. Garbage is refused before `go test` sees it. This suite is
the one legitimate sub-floor caller and opts in explicitly rather than being special-cased in
the script. Six new assertions cover the floor, the shipped 20m, multi-unit durations, and the
`ms`/`m` ambiguity.

### ADVISORY 4 — noted, thank you for saying so explicitly.

### One of your "did not check" items is now closed by execution

**CI.** `mktemp` and the awk classifier were asserted by comment; they are now asserted by
execution. CI passed on `ubuntu-latest` at `8cf5eeb`, and the `test` job log shows
`bash scripts/go-test-budget.sh ./...` and `Testing Go packages (per-package budget: 20m; ...)`.
The other three (interleaving, isolated package runs, an actual 20m overrun) remain unclosed and
I have not claimed otherwise.

### Verification

`./build.sh` green end-to-end (`BUILD_EXIT=0`) from a non-ephemeral worktree, suite **29/29**
(was 23), gate passing while host load climbed past 21.

Ready for round 2.

drellem2:
## Review round 2: PASS
Reviewer: mg-a3db · build ticket: mg-a465 · blocking: 0 · advisory: 3

Head `274fd56`. Round 1's blocking finding is resolved, and resolved better than it was raised.

### The blocking finding — resolved, and the answer changed

I asked for the loaded-host margin to be stated alongside the 5.5x. The builder instead measured a
third anchor on this branch (`internal/agent`, 268.7s isolated at load ~5), checked my exponent
rather than taking it (`k = ln(155.1/81.2)/ln(32/7.8) = 0.458`), and the header now says outright
that **at the load 150-174 this box has recorded, the slowest package may legitimately exceed the
20m budget** — 0.9x margin on the isolated anchor, not 5.5x. Provenance and staleness are both
fixed: the triage table is attributed to the triage run, and the branch's own figures (268.7s vs
216.8s; 99.2s vs 81.2s) are recorded.

The header also records a point that does *not* fit the model — `internal/refinery` at 99.2s under
load ~5 here against 81.2s under load ~7.8 at triage, i.e. lower load and higher time — and names
why (load average does not separate CPU contention from I/O wait). Recording the anomaly rather
than the fitted line is the right call.

**20m remains correct.** One thing the header states as residual risk that is also an improvement,
and is worth adding (advisory 3 below): on the same model and the same anchor, main's 10m default
busts `internal/agent` at **load ~29** — a load this host was sitting at during round-1 review —
while 20m busts at **~131**. The change moves the spurious-timeout threshold up ~4.5x. "May
legitimately exceed at 174" is the residual, not the headline.

### Verification

- **`./build.sh` on `274fd56`: green, `BUILD_EXIT=0`.** `go-test-budget_test.sh` 29/29 (was 23).
  Caveat stated rather than glossed: 49 Go packages came from the test cache on this run, so my
  local execution covered the shell suites and the budget suite, not the Go suite. The Go suite ran
  fresh in CI on this head — see below.
- **CI green on `274fd56`** (run 31138623914), which closes a round-1 "did not check" item by
  execution: the new awk duration parser and the classifier both run on ubuntu-latest, so `mktemp`
  and awk portability are no longer comment-only. The test job log shows the 20m banner.
- **Mutation control re-run independently** on this head (anchor regex made unmatchable, on a copy
  outside the tree): **8 assertions red, up from 7** — and the eighth is exactly the assertion
  advisory 2 flagged as toothless. That assertion now dies with the classifier it exists to
  constrain.
- **Advisory 3's floor probed directly**, 12 values against the shipped script: `5s`, `500ms`,
  `0.5m`, `59s` refused as sub-floor; `60s` accepted (boundary correct); `banana`, `-20m`, `1e3s`
  rejected as non-durations before `go test` sees them; `20m`, `20m30s`, `1h30m` accepted; unset
  defaults to 20m. `500ms` and `0.5m` are both correctly read as sub-minute, so the ms/m
  disambiguation holds.
- **The round-1 adversarial control re-run on this head** — three packages overrunning
  concurrently, one with two `t.Parallel()` tests, one emitting a counterfeit
  `FAIL\t<pkg>\t<elapsed>` line from inside a hanging test — still reports `TIMEOUT: 3 packages`,
  names all three, names both parallel tests, ingests neither counterfeit, and ends in the report.
  The classifier awk and the `tee`/`PIPESTATUS` block are untouched by this round's diff; I
  re-ran it anyway so the pass rests on execution at this head rather than on inference.

### Advisory (non-blocking)

1. `scripts/go-test-budget_test.sh:170` — in `run_budget`'s no-budget branch,
   `POGO_GO_TEST_BUDGET_ALLOW_SHORT=1 && unset POGO_GO_TEST_TIMEOUT && bash ...` sets a *shell*
   variable, not an exported one, so the opt-in never reaches the script
   (`bash -c 'FOO=1 && env | grep -c "^FOO="'` → 0). Harmless today — that branch's only caller is
   test 7, which uses the 20m default and clears the floor anyway — but the comment above it says
   the opt-in applies "throughout", and a future sub-floor case written in that form would fail for
   a reason nobody would look for. The other branch uses the correct `VAR=1 VAR=2 bash ...` prefix
   form.
2. `.github/workflows/ci.yml:60` — "CI is the MORE exposed of the two: these runners are slower
   than the dev host" is refuted by this PR's own CI run. `internal/agent` measured **164.1s on
   ubuntu-latest** at `274fd56`, against 268.7s isolated and 279.6s in-suite on this host. Bounding
   both call sites is right regardless, and the `timeout-minutes: 45` reasoning is unaffected — but
   the stated *reason* is backwards and would mislead whoever next tunes the CI budget. The same
   claim is carried in mg-a465's text, so it is inherited rather than invented here.
3. `scripts/go-test-budget.sh:60-108` — the header states the residual risk without the
   improvement. Adding the status-quo comparison (10m busts at load ~29, 20m at ~131, same model,
   same anchor) would stop a future reader concluding from "may legitimately exceed at 174" that
   the budget is marginal, when it is 4.5x better than what it replaced.

### What I did not check

- Cross-package output interleaving. My probe did not break the classifier, but go's per-package
  output buffering is what makes that safe and I did not try to defeat it.
- A real overrun of the shipped 20m budget. Every overrun I have observed was at 2s, 3s or 10s.
- The Go suite on this head locally — it came from the test cache. CI ran it fresh.
- The loaded-host projections remain an extrapolation ~5-10x past the measured load range, on a
  two-point fit that the branch's own third point does not match. Both the PR and this review say
  so; neither of us has measured the box at load 150.

drellem2:
Merged as 3766789196dd7599092fcceef5f175613121c145 on `main` by the pogo refinery (MR mr-d9r39r2tjv1nt7ule1cg).

Issue #127, opened 2026-08-07
check-mailloops prints an all-clear over 2 of 6 agents: the green path is the only render branch that omits the "who was not judged" disclosure the other two already carry


drellem2:
`pogo check-mailloops` reports a green verdict over a subset of the fleet without naming the
subset or the reason. On this host it judged 2 of 6 agents and printed a line that reads as an
all-clear:

```
$ pogo check-mailloops
All 2 judged agent(s) have a mail-check schedule. (6 in the registry.)
$ echo $?
0

$ pogo check-mailloops --json
{
  "now": "2026-08-07T18:00:11.03549+01:00",
  "scanned": 6,
  "judged": 2
}
```

Four of six agents were not judged. The output names neither which four nor why, and the JSON
carries strictly less than the text does.

## This is not a "both numbers are printed" nit

The headline does print both numbers. The defect is what the **green** branch omits that the other
two branches of the same function already carry.

`MailLoopReport.Render()` (`internal/agent/mailloop_report.go`, `origin/main` @ `3287dd2`) has three
branches:

| branch | condition | names the excluded set? | states the reason? |
|---|---|---|---|
| NOTHING JUDGED | `Missing` empty, `Judged == 0` | — | **yes, in full** |
| **green** | `Missing` empty, `Judged > 0` | **no** | **no** |
| RED | `Missing` non-empty | names the RED set | no, but prints `Judged N of S` |

The `Judged == 0` branch (line 120) already contains almost exactly the sentence the green branch
needs:

```
NOTHING JUDGED: 6 agent(s) in the registry, none of them judgeable.
This is not an all-clear. Polecats, configured-but-stopped agents, and
agents whose prompt tree could not be read are deliberately not judged.
```

So does `internal/deafwatch/watcher.go:440` — the pogod heartbeat announcer that consumes the *same*
judgement:

```
Judged %d of %d agents in the registry. Polecats, configured-but-stopped
agents, and agents whose prompt tree could not be read are NOT judged — see
agent.mailLoopJudgeable.
```

The disclosure exists twice in the tree. The one branch that lacks it is the branch that produces
reassurance — the branch after which a reader stops looking. `Judged == 0` is guarded because it is
the *total* case; `Judged == 2 of 6` is the *partial* case, and it is the common one.

The exclusions are documented in `--help`. That is the right place for the rule and the wrong place
to learn that **this particular run** was narrowed.

## Why it matters more on this check than on most

[...]

PR #129, opened 2026-08-07
check-mailloops must disclose who was not judged on ALL THREE render branches, with the unjudged set in the JSON

drellem2:
`pogo check-mailloops` printed an all-clear over a SUBSET of the fleet without naming the subset or the reason. Reproduced live during triage: it judged 5 of 12 agents and printed `All 5 judged agent(s) have a mail-check schedule. (12 in the registry.)` at exit 0, with the 7 unjudged agents named nowhere, and `--json` carrying only `scanned` and `judged`.

Refs drellem2/pogo#127

## What changed

**All three render branches now end in one shared `renderCoverage`.** `MailLoopReport.Render` had three branches and only the GREEN one omitted the who-was-not-judged disclosure — the `Judged == 0` branch states it in full, and the RED branch carried half of it (`Judged N of S`, a count with no way to turn it into names). Patching only the green branch was the smaller diff and would have left three branches free to diverge again; the RED branch's bare count is fixed as a side effect. This mirrors `internal/ackwatch/ackwatch.go` `renderCoverage`, which already does exactly this on its non-actionable path.

**The unjudged set rides on `MailLoopReport`, so the JSON says what the text says.** Each entry carries name, type, and a machine-stable reason. The field name and no-`omitempty` shape follow the precedent already shipped at `internal/staleness/prompts.go:346`.

```
All 1 judged agent(s) have a mail-check schedule. (4 in the registry.)

Not judged: 3 of 4. This is not a verdict on them — it is who the verdict
above does NOT cover:
  cat-032b             polecat  registers its own mail loop at spawn
  doctor               crew     not running
  ghost                crew     not configured, or its prompt tree could not be read
```

**The reason taxonomy is the coarse-but-true set** `polecat` / `not_running` / `not_configured`. It is deliberately coarser than the three categories `--help` lists: `IsConfiguredAgent` returns false BOTH for an unreadable prompt tree and for a genuinely unconfigured agent, so the finer taxonomy is not computable today, and emitting a reason the code cannot back would be this issue's own failure mode one level in. `--help` now says so out loud. That collapse is a real defect and is filed separately — it is not fixed here.

The reason and the judgeability predicate are **one function**: `mailLoopJudgeable` is now defined as `mailLoopExclusionFor(a) == ""`. A second function computing "why" would be free to disagree with the one computing "whether" — the same drift `mailLoopFor` exists to prevent one level down.

## The version-skew guard (hard requirement)

**Absent and empty are distinguishable on the wire, and an absent field renders UNKNOWN — never 0.**

`internal/client/agent.go` plain-decodes this struct with no version negotiation, so a pogod older than the client simply does not send the new field. A plain slice would have flattened that into a confident "0 not judged" — this issue's exact defect, green, inside its own fix, on the fleet that filed it. That is the CURRENT state, not an edge case: the running pogod is ~93 commits behind main as of this PR.

Mechanism chosen: `Unjudged *[]MailLoopExclusion` with tag `json:"unjudged"` and **no** `omitempty`.

| wire | meaning | render |
|---|---|---|
| `"unjudged": [...]` | daemon reports the set | names every excluded agent and its reason |
| `"unjudged": []` | daemon reports the set, and it is empty | no caveat — genuine full coverage |
| field absent / `null` | daemon does not report the set | count from `scanned - judged` (fields every version sends), WHO and WHY stated as **UNKNOWN** |

```
All 2 judged agent(s) have a mail-check schedule. (6 in the registry.)

Not judged: 4 of 6 — WHO and WHY are UNKNOWN. This pogod is older than the
client and does not report the unjudged set, so the verdict above covers 2 of
6 agents and cannot say which. Upgrade pogod, or ask one at a time with
`pogo agent diagnose <name>`.
```

`TestMailLoopReport_AbsentUnjudgedSetRendersUnknownNotZero` is the guard: it decodes `{"now":...,"scanned":6,"judged":2}` — a payload that never mentions the field — and asserts the render says UNKNOWN, states the count `4 of 6`, and contains neither `Not judged: 0` nor `0 of 6`. It carries a positive control (same shape with `"unjudged": []` prints no caveat) and asserts the wire distinction in both directions.

## Explicit decision: exit status does NOT move

`Actionable() = len(Missing) > 0` is **unchanged**. Everything in the unjudged set is excluded on purpose — mg-738f drew that boundary and the cry-wolf guarantee rests on it — so firing on it would make the exit status useless. This changed what the command DISCLOSES, not what it judges. Recorded here and in a doc comment at the predicate rather than left as a silence, so it is not reopened. A test asserts it directly.

## Out of scope, deliberately

- **Issue item 3** (informational coverage for the excluded set) — changes what is COMPUTED rather than what is DISCLOSED; filed separately.
- **The `IsConfiguredAgent` collapsed-error defect** — filed separately; this PR discloses the collapse rather than papering over it.
- **`internal/deafwatch`'s announcement body** — a different surface with its own snapshot type; untouched.

## Tests

- `TestMailLoopReport_GreenPathNamesWhoItDidNotJudge` — the filed defect, plus all three reasons end to end, plus the JSON, plus the exit-status non-change.
- `TestMailLoopReport_RedPathNamesWhoItDidNotJudge` — the second branch the shared renderer fixed.
- `TestMailLoopReport_AbsentUnjudgedSetRendersUnknownNotZero` — the mandatory skew guard.
- `TestMailLoopExclusionReason_DescribeKeepsAnUnknownReasonLegible` — the other skew direction: a newer pogod's unknown reason renders verbatim, not blank.
- `TestHandleMailLoops_ServesTheFleetRead` gains an assertion that the endpoint puts `"unjudged": []` on the wire rather than omitting it.
- `TestMailLoopReport_InheritsDiagnoseExclusions` still passes on the literal `NOTHING JUDGED` — the shared renderer did not displace it.

Verified with a **cold** cache (`go clean -testcache && ./build.sh`) — exit 0, on `main` at 3fbf303 (mg-6a0b merged, `internal/agent` green).

Approved triage recommendation: mg-e605 (`json triage-packet` block + the "2026-08-07 pm-pogo RULINGS" section)
Work item: mg-0db1

### DIFF
--- a/changelog.d/mg-0db1.fixed.md
+++ b/changelog.d/mg-0db1.fixed.md
@@ -1,0 +1,53 @@
+- **`pogo check-mailloops` names WHO it did not judge on every render branch,
+  and the JSON carries the same set with a reason (mg-0db1,
+  Refs drellem2/pogo#127).** On the reporting host the command judged 2 of 6
+  agents and printed `All 2 judged agent(s) have a mail-check schedule. (6 in
+  the registry.)` at exit 0. Four of six were not judged; the output named
+  neither which four nor why, and `--json` carried only `scanned` and `judged`.
+  Reproduced live during triage at 5 of 12.
+
+  **This was an inconsistency inside one command, not a missing feature.**
+  `MailLoopReport.Render` had three branches and only the GREEN one omitted the
+  who-was-not-judged disclosure: the `Judged == 0` branch states it in full, and
+  the RED branch carried half of it (`Judged N of S`, a count with no way to
+  turn it into names). Three in-tree precedents already do this correctly —
+  `ackwatch.Report.renderCoverage`, `deafwatch.renderBody`, and the shipped
+  `unjudged` JSON field in `internal/staleness/prompts.go` — so the fix applies
+  an established convention rather than inventing one.
+
+  All three branches now end in a single `renderCoverage`, which is the point:
+  the smaller diff would have been to patch the green branch, and that leaves
+  three branches free to diverge again. The RED branch's bare count is fixed as
+  a side effect.
+
+  **The unjudged set rides on the report, so the JSON says what the text says.**
+  Each entry carries name, type, and a machine-stable reason from the coarse set
+  `polecat` / `not_running` / `not_configured`. That set is deliberately coarser
+  than the three categories `--help` lists: `IsConfiguredAgent` returns false
+  BOTH for an unreadable prompt tree and for a genuinely unconfigured agent, so
+  the finer taxonomy is not computable today, and emitting a reason the code
+  cannot back would be this issue's own failure mode one level in. The `--help`
+  text now says so. The collapse itself is a separate defect, filed separately.
+  The reason and the judgeability predicate are one function
+  (`mailLoopExclusionFor`, which `mailLoopJudgeable` is now defined in terms of),
+  so the roster cannot name a reason the predicate would not give.
+
+  **Absent and empty are distinguishable on the wire, and absent renders
+  UNKNOWN — never zero.** `internal/client` plain-decodes this struct with no
+  version negotiation, so a pogod older than the client simply does not send the
+  new field. A plain slice would have flattened that into a confident "0 not
+  judged" — this issue's exact defect, green, inside its own fix, on the fleet
+  that filed it. Not hypothetical: the running pogod was ~93 commits behind main
+  when this was written, so the skew case was the CURRENT state. `Unjudged` is
+  therefore a pointer with no `omitempty`: a report that judged everything puts
+  `"unjudged": []` on the wire, and an absent field renders the count (derivable
+  from `scanned - judged`, which every version sends) with WHO and WHY stated as
+  unknown. `TestMailLoopReport_AbsentUnjudgedSetRendersUnknownNotZero` decodes a
+  payload with the field absent and asserts the render does not claim full
+  coverage.
+
+  **Exit status does not move.** `Actionable() = len(Missing) > 0` is unchanged.
+  Everything in the unjudged set is excluded on purpose — mg-738f drew that
+  boundary and the cry-wolf guarantee rests on it — so firing on it would make
+  the exit status useless. This changed what the command DISCLOSES, not what it
+  judges; recorded here and at the predicate rather than left as a silence.
--- a/cmd/pogo/main.go
+++ b/cmd/pogo/main.go
@@ -1455,9 +1455,21 @@ WHO IS NOT JUDGED, deliberately:
   unreadable prompts  if the prompt tree cannot be read the agent cannot be
                       classified, and a false RED costs more than silence.
 
-Those exclusions mean a small "judged" count is normal. A report that judged
+Those exclusions mean a small "judged" count is normal. Every report NAMES the
+agents it did not judge, whatever the verdict — a clean bill of health over 2 of
+6 agents is a statement about 2 agents, and it says so. A report that judged
 NOTHING says so in as many words rather than printing an all-clear, and a pogod
 with no basis to judge at all is an ERROR here, not an empty list.
+
+The machine-readable reason in --json is coarser than the three categories
+above: it is one of "polecat", "not_running", "not_configured", and the last of
+those covers an unreadable prompt tree as well, because agent.IsConfiguredAgent
+cannot currently tell the two apart. That collapse is tracked separately. The
+reason set is deliberately not more precise than the code can back.
+
+Against a pogod older than this client the unjudged set is absent from the wire
+entirely. That is reported as UNKNOWN — never as zero — because "the daemon did
+not say" and "nobody was excluded" are opposite statements.
 
 REPORTS ONLY — it never registers a schedule, nudges, or restarts. Re-registering
 the loop on the agent's behalf would hide WHY it vanished, and that is the part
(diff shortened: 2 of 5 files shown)


drellem2:
## Review round 1: PASS
Reviewer: mg-4f4c · build ticket: mg-0db1 · blocking: 0 · advisory: 2

Reviewed against the approved triage recommendation in mg-e605 (the `triage-packet` block and the pm-pogo rulings section). Verified on `cb43f9b`, in a separate worktree.

### Verification

`go clean -testcache && ./build.sh` — exit 0. `TestMailLoopReport_InheritsDiagnoseExclusions` still passes on the literal `NOTHING JUDGED`, so the shared renderer did not displace it.

The skew guard was checked against the real thing rather than in the abstract. Running this branch's `pogo check-mailloops` against the live daemon — which is ~93 commits behind and therefore does not send the field — produces:

```
All 5 judged agent(s) have a mail-check schedule. (25 in the registry.)

Not judged: 20 of 25 — WHO and WHY are UNKNOWN. This pogod is older than the
client and does not report the unjudged set, so the verdict above covers 5 of
25 agents and cannot say which. Upgrade pogod, or ask one at a time with
`pogo agent diagnose <name>`.
```

with `--json` returning `"unjudged": null` and exit 0. Absent renders as unknown, not as zero, on the deployment that filed the issue.

### The guard test is a positive control

`TestMailLoopReport_AbsentUnjudgedSetRendersUnknownNotZero` was not taken on trust. Three mutations were applied to `renderCoverage` and reverted; each was caught:

| mutation | result |
|---|---|
| absent flattened to empty (nil branch returns early) | FAIL — missing `UNKNOWN`, missing the count |
| absent rendered as `Not judged: 0 of 6` | FAIL — all four assertions, including both forbidden strings |
| `renderCoverage` dropped from the green branch | FAIL — `TestMailLoopReport_GreenPathNamesWhoItDidNotJudge`, on all seven expected substrings |

The pointer type is load-bearing rather than decorative: changing `Unjudged` to a plain slice does not merely weaken the test, it stops it compiling, because the assertion dereferences it.

### Faithfulness

All three render branches end in one shared `renderCoverage` (`internal/agent/mailloop_report.go:208,214,226`); it is a single function, not a mirrored copy, and it follows `internal/ackwatch/ackwatch.go:664` in name, placement, and the `Not judged: N of S` phrasing. The JSON carries name, type and machine reason for every excluded agent and is exercised end to end. `Actionable()` is byte-unchanged and the non-change is recorded in the PR body and in a doc comment at the predicate. Item 3 and the `IsConfiguredAgent` collapse are absent from the diff; the collapse is disclosed in `--help` rather than papered over. Defining `mailLoopJudgeable` in terms of `mailLoopExclusionFor` is the right call — it removes the possibility of the reason and the predicate disagreeing.

On the open question raised by the author: `MailLoopExclusion` carrying name/type/reason and **not** Identity is correct. The recommendation specifies those three fields, and the reason `MailLoopFinding` carries Identity — correlation with the event log — does not apply to a census line that reports nothing wrong. No change wanted.

### Advisory (non-blocking)

- `internal/agent/api.go:864` — the `a == nil` guard in `mailLoopExclusionFor` is unreachable. The only caller dereferences `a.Name` and `a.Type` in the same composite literal that invokes it (`internal/agent/mailloop_report.go:177-181`), and `Registry.List` would already have panicked in its sort comparator (`internal/agent/agent.go:1176`). Harmless, but it reads as protection that is not there.
- `internal/agent/mailloop_report.go:121` — against an older daemon `--json` emits `"unjudged": null`. That is correctly distinguishable from `[]` and is the whole requirement, but a machine consumer gets no counterpart to the sentence the text render prints. Documented in `--help`; noted only so the asymmetry is on the record.

drellem2:
## Round 1 — PASS, no code change

Recording the disposition of both advisories rather than acting on them, so the decision is visible rather than silent.

**No commit follows this review.** The PASS is against `cb43f9b`, and that is the commit the reviewer ran `./build.sh` cold on and exercised against the live ~93-commit-behind daemon. Pushing a cosmetic change on top would hand the refinery a tree nobody has verified, to fix something the reviewer classified as non-blocking. Not worth the round.

**Advisory 1 — the `a == nil` guard in `mailLoopExclusionFor` (internal/agent/api.go:864) is unreachable.** The reviewer is right, and I checked it rather than taking it: the only caller dereferences `a.Name` and `a.Type` in the same composite literal that calls the function (`mailloop_report.go:177-181`), and the other path in (`mailLoopFor`) already returns early on `a == nil`. So the branch cannot execute, and the reviewer's framing is the accurate one — it reads as protection that is not there. Left in place deliberately for now, noted here so a later reader finds a decision instead of a puzzle. Deleting it is correct whenever these lines are next touched.

**Advisory 2 — against an old daemon `--json` emits `"unjudged": null`.** Working as specified: `null` and `[]` are the two distinguishable wire states the hard requirement asks for, and `null` is the one that means "this daemon does not report the set". The observation underneath it is fair — the JSON has no counterpart to the sentence the text render prints, so a machine reader gets the distinction but not the explanation. That is a JSON-shape question beyond this PR's scope, and I am not widening the scope after a pass.

**Reviewer resolved my open judgement call:** `MailLoopExclusion` keeps name/type/reason and does **not** gain `Identity`. The reason `MailLoopFinding` carries `Identity` is event-log correlation, which does not apply to a census line reporting nothing wrong. Keeping it as shipped.

drellem2:
Merged as 850bcba6d68e0ed0d695400449013b5721d67011 on `main` by the pogo refinery (MR mr-d9r46g2tjv1nt7ule1j0).

